PR #12, opened 2020-03-25
update otel dependency to v0.3.0

rghetia:


exporter/trace/trace_proto.go
```diff
@@ -54,7 +54,7 @@ const (
 	labelHTTPPath       = `/http/path`
 	labelHTTPUserAgent  = `/http/user_agent`
 
-	version = "0.2.3"
+	version = "0.1.0"
```
mayurkale22: Is this intentional? Other places you changed from 2.3 to 3.0.

rghetia: yes. I am going to publish version 0.1.0 for stackdriver exporter.

rghetia: This version reflects stackdriver version and not the otel version.

Issue #25, opened 2020-05-15
Handling timestamp appropriately based on MetricKind


ymotongpoo:
In current spec and implementation of opentelemetry-go, it doesn't return timestamps in aggregated data. (ref. https://pkg.go.dev/go.opentelemetry.io/otel@v0.5.0/sdk/export/metric/aggregator)

But, for Cloud Monitoring, timestamps (`TimeInterval`) are required for [`Point`](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#Point)

In the current implementation (#24), I leave this with tentative implementation but this should be handled in the upstream or workarounds in metric exporter.

PR #28, opened 2020-05-17
Fix aggregation handling and example

ymotongpoo:
Fixed error on submitting Observer value to the backend and fixed the example to eliminate the unrequired variable.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -171,6 +171,7 @@ func (me *metricExporter) exportMetricDescriptor(ctx context.Context, cps export
 	// have many diffrent descriptors. In the cps.ForEach above, it should spawn
 	// goroutines to send CreateMetricDescriptorRequest asynchronously in the case
 	// the descriptor does not exist in global cache (me.mdCache).
+	// See details in #26.
 	for key, md := range mds {
 		req := &monitoringpb.CreateMetricDescriptorRequest{
 			Name:             fmt.Sprintf("projects/%s", me.o.ProjectID),
@@ -227,13 +228,8 @@ func (me *metricExporter) recordToTspb(r *export.Record, res *resource.Resource)
 	}
 
 	p := &monitoringpb.Point{
-		Value: tv,
-	}
-	// MetricKind in CUMULATIVE or DELTA requires Interval field
-	key := keyOf(r.Descriptor())
-	md := me.mdCache[key]
-	if md.MetricKind == googlemetricpb.MetricDescriptor_CUMULATIVE {
-		p.Interval = t
+		Value:    tv,
+		Interval: t,
 	}
 
 	return &monitoringpb.TimeSeries{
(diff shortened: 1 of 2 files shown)


exporter/metric/metric.go
```diff
@@ -374,12 +370,30 @@ func recordToTypedValueAndTimestamp(r *export.Record) (*monitoringpb.TypedValue,
...
+				Seconds: now,
 			},
-		}, t, nil
+		}
+		switch kind {
+		case apimetric.Int64NumberKind:
+			return &monitoringpb.TypedValue{
+				Value: &monitoringpb.TypedValue_Int64Value{
+					Int64Value: value,
+				},
```
rghetia, line 382: This code is repeated so may be it could be converted to a func.

rghetia:
cc/ @nilebox

exporter/metric/metric.go
```diff
@@ -374,12 +370,30 @@ func recordToTypedValueAndTimestamp(r *export.Record) (*monitoringpb.TypedValue,
 			StartTime: &googlepb.Timestamp{
 				Seconds: now,
 			},
-		}
-		return &monitoringpb.TypedValue{
-			Value: &monitoringpb.TypedValue_Int64Value{
-				Int64Value: value,
+			EndTime: &googlepb.Timestamp{
+				Seconds: now,
```
nilebox, line 374: This is to be later fixed in #25 as far as I understand.

exporter/metric/metric.go
```diff
@@ -374,12 +370,30 @@ func recordToTypedValueAndTimestamp(r *export.Record) (*monitoringpb.TypedValue,
...
 			},
-		}, t, nil
+		}
+		switch kind {
```
nilebox, line 377: You have a repetitive code `return ..., t, nil` below.
If you extract this switch-case block into a separate function `countToTypedValue(kind, value)`, the code will be much cleaner:
```go
typedValue := countToTypedValue(kind, value)
return typedValue, t, nil
```

nilebox: You can also do the same for LastValue and Sum below.
Actually would also be nice to extract every if condition there:
```
if count, ok := agg.(aggregator.Count); ok {
  return countToTypedValueAndTimestamp(count, kind)
} else if lv, ok := agg.(aggregator.LastValue); ok {
  return lastValueToTypedValueAndTimestamp(lv, kind)
} else ...
```

exporter/metric/metric.go
ymotongpoo: Worrysome part here is that only Counter returns int64 where as the other kinds return apimetric.Number.

The combination of this and the comment by @nilebox looks better, i.e. extracting each cases to functions and in addition to it make TypeValue part a function.

WDYT?

exporter/metric/metric.go
ymotongpoo: Thank you, Nail. I extracted all cases to functions as suggested. Actually it looks much cleaner.

PR #32, opened 2020-05-21
Add comments need to be noted for future works

ymotongpoo:
This PR just adds small amount of comments that explains current issues on the implementation and how it should be fixed in future works.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -368,7 +368,7 @@ func recordToTypedValueAndTimestamp(r *export.Record) (*monitoringpb.TypedValue,
 	// TODO: Ignoring the case for Min, Max and Distribution to simply
 	// the first implementation.
 	//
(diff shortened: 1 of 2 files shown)


exporter/metric/metric.go
```diff
@@ -367,6 +367,23 @@ func recordToTypedValueAndTimestamp(r *export.Record) (*monitoringpb.TypedValue,
 
 	// TODO: Ignoring the case for Min, Max and Distribution to simply
 	// the first implementation.
+	//
+	// Currently the selector used in the integrator is `simple.NewWithExactMeasure`
```
nilebox, line 371: Looks like the function has been renamed to `NewWithExactDistribution` in v0.6.0 (based on your changes in #34) so the comment will need to be updated accordingly.

ymotongpoo: Thank you for catching this. Updated!

ymotongpoo:
This should be merged before #33

Issue #19, opened 2020-05-07
Add workers to export spans in parallel.


rghetia:
Problem: Currently it exports all spans using a single go subroutine. In a high QPS environment you can increase the batch size but eventually it will result into rpc timeout. 

Solution:
- Create multiple workers (go subroutines) and let each worker take spans from the queue to send it to the backend. See https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/246 and https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/213
- Make the number of workers configurable.
- Add tests to verify with N workers (N > 1).

PR #39, opened 2020-06-08
Add bundler to exporter

stevencl1013:
Bundles spans together and uploads them in separate worker threads

### DIFF
--- a/exporter/trace/cloudtrace_test.go
+++ b/exporter/trace/cloudtrace_test.go
@@ -213,3 +213,79 @@ func TestBundling(t *testing.T) {
 	case <-time.After(time.Second / 5):
 	}
 }
+
+
+func TestBundling_ConcurrentExports(t *testing.T) {
+	var exportMap sync.Map // maintain a collection of the spans exported
+	wg := sync.WaitGroup{}
+	uploadFn := func(ctx context.Context, spans []*tracepb.Span) {
+		for _, s := range spans {
+			exportMap.Store(s.SpanId, true)
+		}
+		wg.Done()
+
+		// Don't complete the function until the WaitGroup is done.
+		// This ensures the semaphore limiting the concurrent uploads is not
+		// released by one goroutine completing before the other.
+		wg.Wait()
+	}
+
+	workers := 3
+	spansPerWorker := 50
+	delay := 2 * time.Second
+	exporter, err := texporter.NewExporter(
+		texporter.WithProjectID("PROJECT_ID_NOT_REAL"),
+		texporter.WithTraceClientOptions(clientOpt),
+		texporter.WithBundleDelayThreshold(delay),
+		texporter.WithBundleCountThreshold(spansPerWorker),
+		texporter.WithMaxNumberOfWorkers(workers),
+		texporter.WithUploadFunction(uploadFn),
+	)
+	assert.NoError(t, err) 
+
+	tp, err := sdktrace.NewProvider(
+		sdktrace.WithConfig(sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
+		sdktrace.WithSyncer(exporter))
+	assert.NoError(t, err)
+
+	global.SetTraceProvider(tp)
+
+	waitCh := make(chan struct{})
+	wg.Add(workers)
+
+	totalSpans := workers * spansPerWorker
+	var expectedSpanIDs []string
+	go func() {
+		// Release enough spans to form two bundles
+		for i := 0; i < totalSpans; i++ {
+			_, span := global.TraceProvider().Tracer("test-tracer").Start(context.Background(), "test-span")
+			expectedSpanIDs = append(expectedSpanIDs, span.SpanContext().SpanID.String())
+			span.End()
+		}
+
+		// Wait for the desired concurrency before completing
+		wg.Wait()
+		close(waitCh)
+	}()
+
+	select {
+	case <-waitCh:
+	case <-time.After(delay / 2): // fail before a time-based flush is triggered
+		t.Fatal("timed out waiting for concurrent uploads")
+	}
+
+	// all the spans are accounted for
+	var exportedSpans []string
+	exportMap.Range(func(key, value interface{}) bool {
+		exportedSpans = append(exportedSpans, key.(string))
+		return true
+	})
+	if len(exportedSpans) != totalSpans {
+		t.Errorf("got %d spans, want %d", len(exportedSpans), totalSpans)
+	}
+	for _, id := range expectedSpanIDs {
+		if _, ok := exportMap.Load(id); !ok {
+			t.Errorf("want %s; missing from exported spans", id)
+		}
+	}
+}
--- a/exporter/trace/cloudtrace_test.go
+++ b/exporter/trace/cloudtrace_test.go
@@ -166,3 +166,49 @@ func TestExporter_Timeout(t *testing.T) {
 		t.Fatalf("err.Error() = %q; want %q", got, want)
 	}
 }
+
+func TestBundling(t *testing.T) {
+	ch := make(chan []*tracepb.Span)
+	uploadFn := func(ctx context.Context, spans []*tracepb.Span) {
+		ch <- spans
+	}
+	exporter, err := texporter.NewExporter(
+		texporter.WithProjectID("PROJECT_ID_NOT_REAL"),
+		texporter.WithBundleDelayThreshold(time.Second / 10),
+		texporter.WithBundleCountThreshold(10),
+		texporter.WithUploadFunction(uploadFn),
+	)
+	assert.NoError(t, err) 
+
+	tp, err := sdktrace.NewProvider(
+		sdktrace.WithConfig(sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
+		sdktrace.WithSyncer(exporter))
+	assert.NoError(t, err)
+
+	global.SetTraceProvider(tp)
+
+	for i := 0; i < 35; i++ {
+		_, span := global.TraceProvider().Tracer("test-tracer").Start(context.Background(), "test-span")
+		span.End()
+	}
+
+	// Read the first three bundles.
+	<-ch
+	<-ch
+	<-ch
+
+	// Test that the fourth bundle isn't sent early.
+	select {
+	case <-ch:
+		t.Errorf("bundle sent too early")
+	case <-time.After(time.Second / 20):
+		<-ch
+	}
+
+	// Test that there aren't extra bundles.
+	select {
+	case <-ch:
+		t.Errorf("too many bundles sent")
+	case <-time.After(time.Second / 5):
+	}
+}
--- a/exporter/trace/cloudtrace_test.go
+++ b/exporter/trace/cloudtrace_test.go
@@ -101,6 +101,8 @@ func TestExporter_ExportSpans(t *testing.T) {
 	exp, err := texporter.NewExporter(
 		texporter.WithProjectID("PROJECT_ID_NOT_REAL"),
 		texporter.WithTraceClientOptions(clientOpt),
+		// handle bundle as soon as span is received
+		texporter.WithBundleCountThreshold(1),
 	)
 	assert.NoError(t, err)
 
@@ -127,14 +129,19 @@ func TestExporter_Timeout(t *testing.T) {
 	mockTrace.spansUploaded = nil
 	mockTrace.delay = 20 * time.Millisecond
 	var exportErrors []error
+	var wg sync.WaitGroup
+	wg.Add(1)
 
 	// Create Google Cloud Trace Exporter
 	exp, err := texporter.NewExporter(
 		texporter.WithProjectID("PROJECT_ID_NOT_REAL"),
 		texporter.WithTraceClientOptions(clientOpt),
 		texporter.WithTimeout(1*time.Millisecond),
+		// handle bundle as soon as span is received
+		texporter.WithBundleCountThreshold(1),
 		texporter.WithOnError(func(err error) {
 			exportErrors = append(exportErrors, err)
+			wg.Done()
 		}),
 	)
 	assert.NoError(t, err)
@@ -149,6 +156,8 @@ func TestExporter_Timeout(t *testing.T) {
 	span.End()
 	assert.True(t, span.SpanContext().IsValid())
 
+	// wait for error to be handled
+	wg.Wait()
 	assert.EqualValues(t, 0, mockTrace.len())
 	if got, want := len(exportErrors), 1; got != want {
 		t.Fatalf("len(exportErrors) = %q; want %q", got, want)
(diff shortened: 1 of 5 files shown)


ymotongpoo:
@stevencl1013 Please run `make precommit`

exporter/trace/trace.go
```diff
@@ -44,23 +53,72 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
+			ctxToSpansMap[cs.ctx] = append(ctxToSpansMap[cs.ctx], cs.spans...)
+		}
+		for ctx, spans := range ctxToSpansMap {
+			e.uploadFn(ctx, spans)
+		}
+	})
+	if o.BundleDelayThreshold > 0 {
+		b.DelayThreshold = o.BundleDelayThreshold
+	} else {
+		b.DelayThreshold = 2 * time.Second
```
nilebox, line 70: Could you extract constants to the top of the file, and give them descriptive names, please?

exporter/trace/trace.go
```diff
@@ -44,23 +53,72 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
+	}
+	if o.BundleCountThreshold > 0 {
+		b.BundleCountThreshold = o.BundleCountThreshold
+	} else {
+		b.BundleCountThreshold = 50
+	}
+	if o.NumberOfWorkers > 0 {
+		b.HandlerLimit = o.NumberOfWorkers
```
nilebox, line 78: Based on the description https://github.com/googleapis/google-api-go-client/blob/cfc62336a21b9af45e2830a9898569d8aac1c5cc/support/bundler/bundler.go#L80-L82, the config parameter should be called `MaxNumberOfWorkers` rather than just `NumberOfWorkers`, or Just `BundleHandlerLimit` to directly refer to the bundler parameter.

exporter/trace/trace.go
```diff
@@ -44,23 +53,72 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
+	}
+	// The measured "bytes" are not really bytes, see exportReceiver.
+	b.BundleByteThreshold = b.BundleCountThreshold * 200
```
nilebox, line 81: This should be done right after setting `BundleCountThreshold`, to make sure they are set together.
Also, there is no `exportReceiver` in this repo?

exporter/trace/trace.go
```diff
@@ -44,23 +53,72 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
+	b.BundleByteLimit = b.BundleCountThreshold * 1000
```
nilebox, line 82: Move next to setting `BundleCountThreshold`

exporter/trace/trace.go
```diff
@@ -44,23 +53,72 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
-	e.uploadFn(ctx, []*tracepb.Span{protoSpan})
+	protoSize := proto.Size(protoSpan)
+	err := e.bundler.Add(&contextAndSpans{
+		ctx: ctx, 
+		spans: []*tracepb.Span{protoSpan},
+	}, protoSize)
+	switch err {
+	case nil:
+		return
+	case bundler.ErrOversizedItem:
```
nilebox, line 105: You need to use the `fallthrough` keyword here if you want to process both cases the same way.
Currently you are just ignoring this error.
See https://github.com/golang/go/wiki/Switch#fall-through

nilebox:
Looks like some changes aren't committed, please run `make precommit` and push missing changes.

exporter/trace/trace.go
```diff
@@ -29,11 +34,19 @@ import (
 type traceExporter struct {
 	o         *options
 	projectID string
+	bundler   *bundler.Bundler
```
ymotongpoo, line 37: @rghetia As far as I remember, there's some issues on using bundler and that's the reason why I avoided using bundler and used BatchWriteSpans in the first implementation. Could you share the background for the record?

rghetia: @ymotongpoo I kind of remember some issue with bundler but I don't remember the details.

ymotongpoo: For record: at least the report in OC pointed the possibility of memory leak.
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/issues/71

nilebox: From that issue, looks like the memory leak issue only appears when the backend is not reachable. Should be relatively easy to reproduce the issue?

exporter/trace/trace.go
```diff
@@ -92,3 +164,48 @@ func (e *traceExporter) uploadSpans(ctx context.Context, spans []*tracepb.Span)
...
+func (o *overflowLogger) delay() {
+	o.pause = true
+	time.AfterFunc(5*time.Second, func() {
+		o.mu.Lock()
+		defer o.mu.Unlock()
+		switch {
+		case o.accum == 0:
+			o.pause = false
+		case o.accum == 1:
+			log.Println("OpenCensus Stackdriver exporter: failed to upload span: buffer full")
```
rghetia, line 191: s/OpenCensus Stackdriver/OpenTelemetry Cloud Trace/

exporter/trace/trace.go
```diff
@@ -92,3 +164,48 @@ func (e *traceExporter) uploadSpans(ctx context.Context, spans []*tracepb.Span)
...
+			o.accum = 0
+			o.delay()
+		default:
+			log.Printf("OpenCensus Stackdriver exporter: failed to upload %d spans: buffer full", o.accum)
```
rghetia, line 195: same as above.

exporter/trace/trace.go
```diff
@@ -92,3 +164,48 @@ func (e *traceExporter) uploadSpans(ctx context.Context, spans []*tracepb.Span)
...
+			o.delay()
+		}
+	})
+}
+
+func (o *overflowLogger) log() {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if !o.pause {
+		log.Println("OpenCensus Stackdriver exporter: failed to upload span: buffer full")
```
rghetia, line 206: same.

ymotongpoo:
Thank you @nilebox for asking me comments. As far as I have heard OC exporter users' experiences, I seldom heard the issues that originate to memory leak. However, still the issue is known, and haven't solved yet. So my request here is for this PR to have at least either (preferably both) of the following:

1. Add todo or note comment in bundler field in traceExporter to touch the potential issue and existing similar issue linked in my previous comment.
2. Add test to reproduce the issue.

I think we need at least 1.

exporter/trace/trace.go
stevencl1013: @ymotongpoo I don't really understand that issue in the ocagent repo. It says 'as long as there is a valid connection to the collector, the “handler” function, i.e. uploadTraces, will be called on the bundle to offload the bundled traces to the collector.' Isn't the handler function called regardless of whether there is a connection? That connection check in the issue is already inside the `uploadTraces` function. And in this PR, there is no such check for a connection, the upload function just calls `BatchWriteSpans`.
Also, since the bundler's [BufferedByteLimit](https://github.com/googleapis/google-api-go-client/blob/master/support/bundler/bundler.go#L76-L78) is set [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/39/files#diff-548b706e759db54d6bb8ae08370364caR86), the bundler will not keep more than that many bytes in memory.

stevencl1013:
@ymotongpoo I don't understand how that issue relates to this exporter. I've looked into it, and I mentioned what I found in this comment: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/39#discussion_r442363302

james-bebbington:
Sorry to come to this one very late. Yesterday I noticed that there are two different versions of the `TraceExporter` in the OpenTelemetry SDK: https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/export/trace/trace.go

* `SpanSyncer` which implements `ExportSpan`
* `SpanBatcher` which implements `ExportSpans`

When setting up your Tracing pipeline, you can configure either of these, by choosing using `WithSyncer` or `WithBatcher` (see https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/provider.go#L186). The Batching work is done in https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/batch_span_processor.go. In the operations exporter, we could just have a helper function that sets up the SpanBatcher pipeline.

So, I wanted to make sure you'd taken a look at that & considered what do we gain by doing our own bundling in the exporter as opposed to just making use of the work that's already done when using the Batch Span Processor?

nilebox:
Follow-up to @james-bebbington 's comment: currently we basically have 2 identical implementations of `ExportSpan` and `ExportSpans`: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/8c1bbc5eb74bd29b602d739f88e98b185a13fc76/exporter/trace/trace.go#L51-L64, implementing both interfaces he mentioned above.

What if we remove the implementation of `SpanSyncer`, and only support `SpanBatcher` with `ExportSpans`?
That way we will force users to always register our exporter via `WithBatcher`, and the bundler won't be necessary.

FWIW there is already a precedent of this approach:
- Zipkin exporter only implements `SpanBatcher`: https://github.com/open-telemetry/opentelemetry-go/blob/e5267a3aa898f14cce75cb640ee4153379965057/exporters/trace/zipkin/zipkin.go#L92
- Jaeger only implements `SpanSyncer`: https://github.com/open-telemetry/opentelemetry-go/blob/e5267a3aa898f14cce75cb640ee4153379965057/exporters/trace/jaeger/jaeger.go#L193

stevencl1013:
Thanks for the info @james-bebbington and @nilebox, I wasn't aware of the difference between `WithSyncer` and `WithBatcher` before. I've been doing my experimentation on a simple application similar to the example in this repo, which uses the `WithSyncer` option. 

I read in internal documentation that OpenTelemetry exporters should use both batching and async writes, so I still believe this PR implements those functionalities for this exporter when using the `WithSyncer` option. When I created this PR, I had two main goals:

1. Add the ability to export spans in parallel, and let users configure the number of worker threads (#19).
2. Reduce the high latency that I had experienced in my testing, and that was also reported later by a user in #47.

If we were to only support `SpanBatcher` and remove the bundler, then I think the 2nd goal would be achieved, but the 1st would not. Would it make sense to keep supporting `SpanSyncer`, either by keeping implementations of both interfaces, or by taking a similar approach to the Jaeger exporter (only implementing `SpanSyncer` and using a bundler)?

james-bebbington:
> Thanks for the info @james-bebbington and @nilebox, I wasn't aware of the difference between `WithSyncer` and `WithBatcher` before. I've been doing my experimentation on a simple application similar to the example in this repo, which uses the `WithSyncer` option.

No worries, we also weren't aware of it until recently. There is probably some general OpenTelemetry documentation missing with recommendations around how to go about implementing exporters.

> If we were to only support `SpanBatcher` and remove the bundler, then I think the 2nd goal would be achieved, but the 1st would not. Would it make sense to keep supporting `SpanSyncer`, either by keeping implementations of both interfaces, or by taking a similar approach to the Jaeger exporter (only implementing `SpanSyncer` and using a bundler)?

Yea we definitely want to support batched & parallel exporting. At the same time, we want to make sure we don't make this confusing for users:

- If they use `WithSyncer` but then we batch the data in the exporter anyway, that's somewhat confusing, although maybe that's not a big deal
- If they use `WithBatcher`, then it will be confusing that the batch size can be configured in both the [batch span processor](https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/batch_span_processor.go#L143) and in the exporter via `BundleCountThreshold`

We can essentially choose which setup the user has to use by only implementing one of those interfaces (and optionally also creating a wrapper initialization function to set up the pipeline for them).

I'm not really sure what the right thing is to do here though. Any thoughts @rghetia @ymotongpoo @nilebox?

exporter/trace/cloudtrace_test.go
```diff
@@ -127,14 +134,19 @@ func TestExporter_Timeout(t *testing.T) {
 	mockTrace.spansUploaded = nil
 	mockTrace.delay = 20 * time.Millisecond
 	var exportErrors []error
+	var wg sync.WaitGroup
+	wg.Add(1)
```
james-bebbington, line 138: Would be more idiomatic to use a channel for this I think

nilebox:
> If we were to only support SpanBatcher and remove the bundler, then I think the 2nd goal would be achieved, but the 1st would not.

@stevencl1013 Can we try addressing this issue in the SDK batcher then instead? That would benefit all users of Go SDK rather than just Stackdriver.

---
I'm not sure what to suggest that would allow using bundler without making it confusing tbh:
- It seems that `Syncer` is designed for use when user wants to send ("sync") spans straight to the backend without buffering, so bundler would go against that design.
- Similarly, cascading batchers (SDK batcher + bundler) might be confusing as well.

ymotongpoo:
@stevencl1013 Sorry for delayed response.

Here's the reason why I mentioned that issue in OC exporter:

* Having connectivity check should be TODO, thus the implementation should be similar to OC exporter.
* Though BufferedByteLimit exists, that code is copied from OC exporter, where the issue exists.

Anyway, as @nilebox and @james-bebbington suggested, now we have `SpanBatcher` in the SDK and it look better to use it for batching to avoid additional complexity.

ymotongpoo:
@stevencl1013 I had a offline discussion with @rghetia about the use of bundler. In short, as you confirmed first, we don't check the connectivity and won't in this exporter, so it's safe to use bundler here.

Background:
Originally I thought we need to have the connectivity check later on like OC exporter, which may pile up the unsent spans in the memory. However, the spec for error handling in the upstream doesn't mention such behavior, and how to handle these unsent spans are up to exporter implementations. We should just throw the unbuffered spans in this exporter and let the collector to handle such complex cases.

reazahmed:
Apologies for jumping in in the middle of the discussion.

As per OTel specification (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#built-in-span-processors):

The OTel standard SDK MUST implement the simple processor (i.e., sync case). With current simple process, a user will experience very high delay for each RPC. This change handles the simple processor case, by brining down latency from ~10sec to ~200 milliseconds.

james-bebbington:
For reference, given the way the SDK is set up, we decided there isn't any ideal solution, but for now decided to:

- Merge this PR as is
- Create a follow up PR, that includes a breaking change, to:
  - Remove the `ExportTraces` function, and only keep `ExportTrace` function
  - Add a `NewExportPipeline` function that we expect consumers to use rather than setting up a `Syncer` themselves

i.e. we will implement this in the same way as the Jaeger exporter (see the Otel Go SDK).

FYI @nilebox I know you had some reservations around the explicitness of the word `Syncer` but we decided this is the least bad of two not great solutions for now: duplicate batching options is likely to add more confusion and create more difficulties with regards to making backwards compatible changes in the future.

@stevencl1013 - in the next PR it'd also be good to add/update the `example` code in this repo

Issue #45, opened 2020-06-12
Be less prescriptive about span names


EronWright:
I observe that the exporter has an opinion about span names, and I feel that it shouldn't. 

Note that `exporter/trace/trace_proto.go` computes the span display name, which makes it impossible to use short, application-specific span names as enjoyed by other Cloud Tracing applications (e.g. a method name).

Suggestion for improvement:
1. introduce a formatter in the exporter options to generate the span display name.
2. Use the existing logic as the default formatter.

stevencl1013:
This should be resolved now, you can see this [test](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/commit/a1393affee9a13ed21974ae1d791da1b3189f4a0#diff-30183b1f2afef55ac0a5fff5bf0b340dR133) for an example of how to write and use your own formatter.

EronWright:
Thanks @stevencl1013 it looks great!

PR #48, opened 2020-06-17
Allow custom display name format for spans (#45)

stevencl1013:
This allows users to, when creating the exporter, use a custom DisplayNameFormatter function to specify how to create the display name from span data. Default format remains the same. Addressing issue #45


ymotongpoo:
@stevencl1013 Can you run `make precommit`? It's failing as the following:

```
=== RUN   TestExporter_Timeout
    TestExporter_Timeout: cloudtrace_test.go:157: err.Error() = "rpc error: code = Unknown desc = context deadline exceeded"; want "rpc error: code = DeadlineExceeded desc = context deadline exceeded"
--- FAIL: TestExporter_Timeout (0.00s)
FAIL
FAIL	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace	0.029s
FAIL
make: *** [Makefile:85: test-386] Error 1
```

stevencl1013:
@ymotongpoo When I ran `make precommit` on my machine, all the tests passed. I think there might be an issue with the timeout test, because it did get a `context deadline exceeded` error, but failed because it got `rpc error: code = Unknown` instead of `rpc error: code = DeadlineExceeded`. I've experienced this issue occasionally on my machine as well in the past.

exporter/trace/trace_proto.go
```diff
@@ -69,10 +69,14 @@ func protoFromSpanData(s *export.SpanData, projectID string) *tracepb.Span {
 	spanIDString := s.SpanContext.SpanID.String()
 
 	name := s.Name
```
nilebox, line 71: Please rename this variable to `displayName` - Span has the `Name` field which will have completely different value (see below).
Also, the initialization with `s.Name` value seems redundant since it will always be overwritten below.

nilebox: I would actually suggest to extract this entire code block into `generateDisplayName(s, format)` to make the code cleaner / easier to read.

exporter/trace/trace_proto.go
```diff
@@ -60,25 +60,33 @@ const (
...
+func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) (displayName string) {
+	if format == nil {
+		switch s.SpanKind {
+		// TODO(ymotongpoo): add cases for "Send" and "Recv".
+		default:
+			displayName = fmt.Sprintf("Span.%s-%s", s.SpanKind, s.Name)
+		}
+	} else {
+		displayName = format(s)
+	}
```
james-bebbington, line 72: It's always a little easier to read code if you minimize nesting imo:

```suggestion
    if format != nil {
        return format(s)
    }

    switch s.SpanKind {
        // TODO(ymotongpoo): add cases for "Send" and "Recv".
        default:
            return fmt.Sprintf("Span.%s-%s", s.SpanKind, s.Name)
    }
```

exporter/trace/cloudtrace_test.go
```diff
@@ -122,6 +123,45 @@ func TestExporter_ExportSpans(t *testing.T) {
...
+		))
+	assert.NoError(t, err)
+
+	global.SetTraceProvider(tp)
+	_, span := global.TraceProvider().Tracer("test-tracer").Start(context.Background(), spanName)
+	span.End()
+	assert.True(t, span.SpanContext().IsValid())
+
+	// wait exporter to flush
+	time.Sleep(20 * time.Millisecond)
```
james-bebbington, line 160: Sleeping for a block amount of time in tests in never ideal. To make this signficiantly more robust, and faster, a better alternative to this is to do something like
```go
assert.Eventually(t, function() {
    ...
}, 500 * time.Millisecond, 2 * time.Millisecond)
```

I wouldn't bother changing this now though. After you add the `NewPipeline` method that returns a flush function (see the Jaeger exporter), you can avoid the sleep here entirely.

stevencl1013: ok, sounds good. I'll leave this for now, and when I create the `NewPipeline` method, I plan to update this test and others to use the pipeline and flush function as appropriate.

PR #49, opened 2020-06-17
New exporter

YANYZP:
a new exporter that will implement both trace and metrics exporters and collect Monitored resources to inject into the spans


exporter/trace/trace_proto.go
```diff
@@ -84,7 +84,7 @@ func protoFromSpanData(s *export.SpanData, projectID string) *tracepb.Span {
 		SameProcessAsParentSpan: &wrapperspb.BoolValue{Value: !s.HasRemoteParent},
 	}
 	if s.ParentSpanID != s.SpanContext.SpanID && s.ParentSpanID.IsValid() {
-		sp.ParentSpanId = fmt.Sprintf("%.16x", s.ParentSpanID)
```
nilebox, line 87: This change seems unrelated?

exporter/exporter.go
```diff
@@ -0,0 +1,104 @@
...
+	NumberOfWorkers int
+
+	// MetricDescriptorTypeFormatter is the custom formtter for the MetricDescriptor.Type.
+	// By default, the format string is "custom.googleapis.com/opentelemetry/[metric name]".
+	MetricDescriptorTypeFormatter func(*apimetric.Descriptor) string
+}
+
+
+// Exporter is a trace and metric exporter
+type Exporter struct {
```
nilebox, line 100: We probably want to have a working Exporter with tests before merging it.

@james-bebbington do you think having this exporter is useful, or some "auto-detection" should be built into trace and metric exporters themselves instead?

james-bebbington: I'm not exactly sure how this wrapper is intended to be used.

I think we want the Trace & Metric exporters to share the same code for translating from Otel Resource -> Cloud Ops Monitored Resource, but I would have thought just using a shared function.

What would the API look like for configuring this wrapper exporter, as opposed to the existing APIs? e.g. https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/master/example/metric/example.go#L71

exporter/trace/trace_proto.go
YANYZP: yes

PR #50, opened 2020-06-19
New exporter

YANYZP:
The trace part of the wrapper exporter with tests(covering cloudtrace_test.go)


exporter/exporter.go
```diff
@@ -0,0 +1,160 @@
...
+	// for now. The minimum number of workers is 1.
+	NumberOfWorkers int
+
+	// MetricDescriptorTypeFormatter is the custom formtter for the MetricDescriptor.Type.
+	// By default, the format string is "custom.googleapis.com/opentelemetry/[metric name]".
+	MetricDescriptorTypeFormatter func(*apimetric.Descriptor) string
+}
+
+
+// Exporter is a trace and metric exporter
```
nilebox, line 98: This description doesn't explain why do we need a wrapper. Could you please create / link a GitHub issue describing the end goal in more detail?

exporter/exporter.go
```diff
@@ -0,0 +1,160 @@
...
+func (e *Exporter) ExportSpan(ctx context.Context, sd *texport.SpanData) {
+	_, monitoredResMap := readMonitoredResourcesFromMetricsExporter()
+	sd = injectLabelsIntoSpan(sd, monitoredResMap)
+	e.traceExporter.ExportSpan(ctx, sd)
+}
+
+// ExportSpans exports a slice of SpanData to Stackdriver Trace in batch
+func (e *Exporter) ExportSpans(ctx context.Context, sds []*texport.SpanData) {
+	_, monitoredResMap := readMonitoredResourcesFromMetricsExporter()
+	sds = injectLabelsIntoSpans(sds, monitoredResMap)
```
nilebox, line 153: Seems like this is the main purpose of the wrapper - injecting auto-detected resource labels into spans.
Is that correct?

exporter/exporter.go
```diff
@@ -0,0 +1,160 @@
...
+	
+	return &Exporter{
+		traceExporter: te,
+	}, nil
+}
+
+// readMonitoredResourcesFromMetricsExporter obtains Monitored Resources labels from metrics exporter
+func readMonitoredResourcesFromMetricsExporter() (resType string, labels map[string]string) {
+	//TODO: read the resources from metric. Now the autodetect is for mockup
+	return monitoredresource.Autodetect().MonitoredResource()
```
nilebox, line 124: AFAIK the long term plan was to move auto-detection to https://github.com/open-telemetry/opentelemetry-go, so that it can be reused by any exporter. 
The only problem I am aware of is that OpenTelemetry resources don't have the dedicated "type" field, so we have to define it as a "well known" resource label (convention).

For the short term implementing auto-detection in the exporter might be fine, but we shouldn't use OpenCensus code here, I think. So maybe we should copy the code over to this repo, and eventually extract it to `opentelemetry-go`?

@james-bebbington may have other opinion / know more details.

go.mod
```diff
@@ -1,3 +1,15 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go
 
 go 1.14
+
+require (
+	cloud.google.com/go v0.57.0 // indirect
+	contrib.go.opencensus.io/exporter/stackdriver v0.13.1
```
nilebox, line 7: We shouldn't have this dependency, as we want to eventually get rid of dependency on the OpenCensus code.

exporter/exporter.go
nilebox: Also, why do we need a single wrapper for both trace and metric exporter? Wouldn't it make sense to keep them separate, since Go SDK declares them as separate interfaces?

exporter/exporter.go
YANYZP: Yes, it is. We want a combined exporter to upload spans with monitored resource labels. Besides, we may modify metric exporter so that it can access and preserve the monitored resource labels and export them to the combined exporter.

go.mod
YANYZP: Yes. The Opencensus autodetect would be removed from exporter dependency when we are able to get these monitored resources from metrics exporter. We can finally completely get rid of it when the autodetect feature in OpenTelemetry is implemented.

james-bebbington:
@YANYZP I don't think this PR is needed anymore based on our discussion today?

PR #52, opened 2020-06-23
Res mapper

YANYZP:
Now ready for final review
======================================================
Implemented a resource mapper from OT resources to Monitored Resources.
For now, it only covers the K8S. More types would be added in the future.
OT resources:
https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/StandardResources.md#kubernetes
Monitored resources:
https://cloud.google.com/monitoring/api/resources#tag_k8s_container
The OpenCensus had a similar function.
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/resource.go#L185


exporter/metric/metric.go
```diff
@@ -276,21 +319,80 @@ func (me *metricExporter) recordToMdpb(record *export.Record) *googlemetricpb.Me
...
+	resLabelMap := make(map[string]string)
+	for _, label := range resLabelList {
+		resLabelMap[string(label.Key)] = label.Value.AsString()
+	}
+
+	resTypeStr := "global"
+	monitoredReslabelsMap := make(map[string]string)
+	match := genericResourceMap
+	
+	if resType, found := resLabelMap["type"]; found {
```
james-bebbington, line 349: I assume this PR is still work in progress, but the overall approach looks like its in the right direction.

Note though that OpenTelemetry resources no longer have a "type" property, but instead just a list of attributes where the prefix of the attribute can be used to determine the type (e.g. the k8s prefix in the key "**k8s**.pod.name"). Check the implementation in the JS operations exporter for reference as to how they've done this mapping.

exporter/metric/metric_test.go
```diff
@@ -118,3 +118,87 @@ func TestRecordToMpb(t *testing.T) {
...
+			"location": "us-central1-a",
+			"cluster_name": "opentelemetry-cluster",
+			"namespace_name": "default",
+			"pod_name": "opentelemetry-pod-autoconf",
+			"container_name": "opentelemetry",
+		},
+		map[string]string{
+			"project_id":     "",
+		},
+		map[string]string{
```
dukeng, line 166: Should we always ignore the entire k8s type if a field is missing? Maybe parsing partial data is fine. What do you think?

exporter/metric/metric.go
```diff
@@ -74,6 +74,48 @@ type metricExporter struct {
...
+	K8SKeyPodName        = "k8s.pod.name"
+	K8SKeyDeploymentName = "k8s.deployment.name"
+
+	stackdriverGenericTaskNamespace = "contrib.opencensus.io/exporter/stackdriver/generic_task/namespace"
+	stackdriverGenericTaskJob       = "contrib.opencensus.io/exporter/stackdriver/generic_task/job"
+	stackdriverGenericTaskID        = "contrib.opencensus.io/exporter/stackdriver/generic_task/task_id"
+)
+
+var k8sContainerMap = map[string]string{
+//	"project_id":     "",
```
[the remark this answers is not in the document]
YANYZP: I think we do not need to care about the field project_id in the OT resources because we would set the project_id  to the value of the project_id of the metrics exporter option in the resourceToMonitoredResourcepb function.

exporter/metric/metric_test.go
YANYZP: I see. In fact, many k8s Monitored resources like k8s_node, k8s_pod, and k8s_cluster, share some common fields. (https://cloud.google.com/monitoring/api/resources#tag_k8s_cluster) I think we need to follow the definition strictly otherwise we may misjudge a type of Monitored resources.

exporter/metric/metric.go
YANYZP: Thanks for pointing out. The JS version used CloudProvider to differentiate the type. (https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#cloud) However, there are only three types: "aws", "gcp", and "azure". Since we need "k8s", I guess we need to talk with backend team to add a new constant "k8s"?

exporter/metric/metric_test.go
dukeng: Historically in Opencensus we only use k8s_container (https://cloud.google.com/monitoring/api/resources#tag_k8s_container) resource type as it already consists of all the related fields for Kubernetes. It would take additional logic to parse by cluster, nodes, pods, etc. By having one resource type for k8s, parsing the resource labels would be much easier in the Cloud Trace backend as well.

YANYZP: I see. I would allow missing fields for more flexible types of monitored resources

exporter/metric/metric.go
dukeng: I think "k8s" is a deployment service (https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#kubernetes) and is not related to the environment (https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#environment). The environment probably requires a different resource type other than "k8s"

exporter/metric/metric.go
```diff
@@ -74,6 +74,42 @@ type metricExporter struct {
 	startTime time.Time
 }
 
+
+// TODO: put all constants into a separate file in the same package
```
james-bebbington, line 78: Most likely these constants will eventually be included in the Go SDK and we can just reference those values directly, but for now putting them in the exporter is fine (but yes you should move them to another file)

exporter/metric/metric.go
```diff
@@ -74,6 +74,42 @@ type metricExporter struct {
...
+
+// Mappings for the well-known OpenCensus resource label keys
+// to applicable Stackdriver Monitored Resource label keys.
+// A uniquely identifying name for the Kubernetes cluster. Kubernetes
+// does not have cluster names as an internal concept so this may be
+// set to any meaningful value within the environment. For example,
+// GKE clusters have a name which can be used for this label.
+const (
+	CloudType = "cloud"
```
james-bebbington, line 87:
```suggestion
```

exporter/metric/metric.go
james-bebbington: You're right that the Javascript implementaton is incomplete. However, the approach you're currently taking to determine if the resource type is incorrect.

As per Duke's comment, "K8s" is a service rather than a Cloud Provider, so will never show up as a possible value for "cloud.provider". To determine if we are running on a K8s Container, you should be able to just do something like checking for the existance of a `k8s.*` attribute (e.g. `k8s.pod.name`?).

If we wanted to only record info for K8s Containers running on GKE, we could additionally check `cloud.provider == "gcp"`. Note that would likely also include K8s clusters running on GCE directly, I'm not sure if we would want to distinguish those cases but if we did, we'd likely need to establish a new semantic attribute such as `k8s.provider.name`

ymotongpoo:
@james-bebbington I confirmed that the CLA failure is because this PR contains the commit from another user who didn't sign CLA for this PR. I think this is caused in the meddle of merge process for @YANYZP to catch up the upstream (this repo.) One workaroud I come up with is to make a diff between master HEAD and this PR and create a patch and then create the alternative branch with the patch.

exporter/metric/metric.go
ymotongpoo: +1. For the sake of easier removal after Go SDK's introduction of auto detector, moving relevant contents into another file in this PR sounds better.

exporter/metric/constants.go
```diff
@@ -0,0 +1,48 @@
...
+	// Cloud Providers
+	CloudProviderAWS   = "aws"
+	CloudProviderGCP   = "gcp"
+	CloudProviderAZURE = "azure"
+
+	K8SKeyClusterName    = "k8s.cluster.name"
+	K8SKeyNamespaceName  = "k8s.namespace.name"
+	K8SKeyPodName        = "k8s.pod.name"
+	K8SKeyDeploymentName = "k8s.deployment.name"
+)
```
james-bebbington, line 48: Can you please add a newline at the end of each of these files? It's not essential but just a convention to avoid any possible issues with various tools.

exporter/metric/metric_test.go
```diff
@@ -118,3 +118,98 @@ func TestRecordToMpb(t *testing.T) {
 		t.Errorf("%v", aggError)
 	}
 }
+
+
+func TestResourceToMonitoredResourcepb(t *testing.T) {
+	resList := []*resource.Resource{
```
james-bebbington, line 124: Could you please add a couple of tests cases:

* Only Partial Data provided - e.g.
```
kv.String("cloud.provider", "gcp"),
kv.String("k8s.pod.name", "opentelemetry-pod-autoconf"),
```

* Cloud Provider = GCP, but invalid data / unknown resource type - e.g. 
```
kv.String("cloud.provider", "gcp"),
kv.String("cloud.zone", "us-central1-a"),
```

exporter/metric/metric.go
```diff
@@ -74,6 +74,28 @@ type metricExporter struct {
...
 }
 
+// Below are maps with monitored resources fields as keys
+// and OpenTelemetry resources fields as values
+var k8sContainerMap = map[string]string{
+	"location":       CloudKeyZone,
+	"cluster_name":   K8SKeyClusterName,
+	"namespace_name": K8SKeyNamespaceName,
+	"pod_name":       K8SKeyPodName,
+	"container_name": K8SKeyDeploymentName,
```
james-bebbington, line 84: Shouldn't this be "container.name" rather than "k8s.deployment.name"?

exporter/metric/metric.go
```diff
@@ -272,25 +291,92 @@ func (me *metricExporter) recordToMdpb(record *export.Record) *googlemetricpb.Me
 	}
 }
 
+// checkPrefixInMapKeys checks whether the non-empty prefix string
+//  is in the values of the map
+func checkPrefixInMapKeys(labelMap map[string]string, prefix string) bool {
+	if len(prefix) == 0 {
+		return false
+	}
```
james-bebbington, line 299: ^ I'd probably get rid of that condition. I don't think you have any tests that cover that condition, and technically an empty string is a prefix of any key so this should return true anyway.

exporter/metric/metric.go
```diff
@@ -272,25 +291,92 @@ func (me *metricExporter) recordToMdpb(record *export.Record) *googlemetricpb.Me
...
-func (me *metricExporter) resourceToMonitoredResourcepb(_ *resource.Resource) *monitoredrespb.MonitoredResource {
-	// TODO: Implement the process to convert Resource to MonitoringResoruce.
-	// ref. https://cloud.google.com/monitoring/api/resources
-	// This method only returns global resource for now.
-
-	// "global" only accepts "project_id" for label.
-	// https://cloud.google.com/monitoring/api/resources#tag_global
+func (me *metricExporter) resourceToMonitoredResourcepb(res *resource.Resource) *monitoredrespb.MonitoredResource {
+
+	if res == nil || res.Len() == 0 {
```
james-bebbington, line 314: Since you end up doing the same initialization in both branches of code, you could avoid a bit of duplication by doing:

```go
monitoredRes := &monitoredrespb.MonitoredResource{
	Type: "global",
	Labels: map[string]string{
		"project_id": me.o.ProjectID,
	},
}

if res == nil || res.Len() == 0 {
	return monitoredRes 
}

...
monitoredRes.Labels = transformResource(match, resLabelMap)
```

exporter/metric/metric.go
```diff
@@ -272,25 +291,92 @@ func (me *metricExporter) recordToMdpb(record *export.Record) *googlemetricpb.Me
...
+				"project_id": me.o.ProjectID,
+			},
+		}	
+	}
+
+	// labels in Resource are in the form of []kv.KeyValue
+	// convert them into a map of kv.String
+	resLabelList := res.Attributes()
+
+	resLabelMap := make(map[string]string)
```
james-bebbington, line 330: I'm okay if you leave this but in terms of performance, note that it's probably not worth building and allocating this working map vs just iterating through the collection 1-2 extra times. In fact I think with the current code it'd be the same number of iterations.

i.e.
Currently you iterate:
- Once to generate the map
- Once to determine if a "k8s" key exists

Instead you could iterate:
- Once to find CloudKeyProvider
- Once to determine if "k8s" key exists
(then iterate through `input` instead of `match` in `transformResource`)

Having said that, it might be easier to maintain in the future with a map as we add more resource types, and we can benchmark / fine tune this in the future.

Instead, consider at least putting the code to build this map into a separate function to improve readability.

exporter/metric/metric_test.go
james-bebbington: Actually maybe those are both the same case 😆 

Does Cloud Monitoring distinguish between mandatory and optional fields? Do we need to do something in our code to validate that? i.e. if we don't have all the fields for a given resource, should we fallback to global or something? Otherwise Cloud Monitoring may reject the request entirely? (I could be wrong - it wasn't obvious to me what happens from the API docs)

james-bebbington: Sure, but my question is what will happen if we attempt to export data that has the following attributes?

```go
kv.String("cloud.provider", "gcp"),
kv.String("cloud.zone", "us-central1-a"),
kv.String("k8s.cluster.name", "opentelemetry-cluster"),
```

Your mapping code will identify it as type "k8s_container" due to including a "k8s.*" attribute, but it doesn't have all the required (?) attributes. What will happen in this case? (and we should have a test case to cover it)

exporter/metric/metric.go
```diff
@@ -272,25 +291,89 @@ func (me *metricExporter) recordToMdpb(record *export.Record) *googlemetricpb.Me
...
+	resLabelMap := generateResLabelMap(res)
+
+	resTypeStr := "global"
+	match := map[string]string{}
+	
+	if resType, found := resLabelMap[CloudKeyProvider]; found {
+		switch resType {
+		case CloudProviderGCP:
+			if checkPrefixInMapKeys(resLabelMap, "k8s") {
+				resTypeStr = "k8s_container"
```
james-bebbington, line 334: Should put these strings into constant fields

exporter/metric/metric_test.go
YANYZP: I agree. Initially, I had the same opinion and treated types of missing fields as global monitored resources in the same way of Go OpenCensus (https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/resource.go#L162). Now I would strongly recommend keeping consistent with Opencensus since the backend would not receive monitored resources with missing fields. @dukeng

james-bebbington: Okay cool - looks good now.

I think you could have one more test case for a "global" resource type, e.g.:
```go
kv.String("cloud.provider", "gcp"),
kv.String("cloud.zone", "us-central1-a"),
```

exporter/metric/metric_test.go
```diff
@@ -118,3 +118,159 @@ func TestRecordToMpb(t *testing.T) {
...
+			"k8s_container", 
+			map[string]string{
+				"project_id":     "",
+				"location": "us-central1-a",
+				"cluster_name": "opentelemetry-cluster",
+				"namespace_name": "default",
+				"pod_name": "opentelemetry-pod-autoconf",
+				"container_name": "opentelemetry",
+			},
+	    },
```
james-bebbington, line 148: Character spacing is a bit wonky in this file. I think you are using an editor that isn't consistently converting to tabs for you. Can you please fix this?

Also we need to set up a linter to automatically detect these kinds of things. I've added a ticket to track that here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/59

james-bebbington:
@YANYZP this all looks good to me no,w other than the character spacing issue mentioned above.

As well as that, can you follow Yoshi's earlier advice of creating a new branch that doesn't include the commit from a user who didn't sign the CLA? All of your PRs have this problem so you probably need to update the master branch in your forked repo.

exporter/metric/metric.go
```diff
@@ -74,6 +74,41 @@ type metricExporter struct {
 	startTime time.Time
 }
 
+// Below are maps with monitored resources fields as keys
+// and OpenTelemetry resources fields as values
+var k8sContainerMap = map[string]string{
```
dukeng, line 79: - I don't see k8sClusterMap here. Is there any reason you exclude that one? 
- Looks like you impose a strict matching of these resource types (no field should be missing). Is there any reason to throw away partial labels? For example, if we only have "location", "cluster_name" and "pod_name" we can't match this with any predefined resource labels and would return empty. Is it expected?

YANYZP: 1) k8sClusterMap would be included in the next version.
2) I think we should follow the instruction strictly and discard partial labels because
i. according to the definition of labels in MonitoredResource(https://github.com/googleapis/go-genproto/blob/master/googleapis/api/monitoredres/monitored_resource.pb.go#L217), the Labels map is required to contain the values for all the resources label descriptors. In other words, all keys must exist. 
ii. Otherwise, if we allow missing labels, it would greatly increase the difficulty of subdividing the resources types since we need to consider all possible cases of missing label keys.

PR #53, opened 2020-06-24
Export labels

YANYZP:
This PR would be abandoned soon. The function of exporting labels would be implemented in exporter.go. The PR(https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/50/files) would extend this function.

PR #55, opened 2020-06-29
Inject span

YANYZP:
This includes functions of injecting labels into trace spans and initializing label maps by one time


exporter/trace/cloudtrace.go
```diff
@@ -206,11 +212,30 @@ func newContextWithTimeout(ctx context.Context, timeout time.Duration) (context.
 	return context.WithTimeout(ctx, timeout)
 }
 
+// SetResLabelMapForOnce intializes ResLabelMap for once since 
+// monitored resources are static
+func (e * Exporter) SetResLabelMapForOnce(input map[string]string) {
```
james-bebbington, line 217: Even though it's true that monitored resources are static, OT doesn't preclude the possibility of specifying different resource info for different telemetry data.

For example, see this proposal which would allow different resource scopes: https://github.com/open-telemetry/oteps/issues/78

As such, we shouldn't make the assumption of static resources in the exporter. Instead, you should just translate from the Otel resource in the trace data to the expected attribute values, similar to what you did for the metrics resource mapping code in the other PR.

---

Having said that, this does come at the cost of performance. We may end up looking to optimize this somewhat in the future, but I don't think we should do that now.

james-bebbington: Hmm I should have read your other PR first sorry as that makes it more clear what you're trying to achieve here.

Getting the resource from the controller config might actually be a reasonable thing to do (perhaps it would be better to get it from the top-level tracer/meter config). This is a somewhat dependent on how the SDK changes over time, but we could probably fix this if the SDK changed significantly.

Interesting idea - let me think about it some more and get back to you tomorrow.

FYI @nilebox

YANYZP: No problem. In fact, the PR(https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/53/files) needs to be changed because I plan to export labels from the merged exporter instead of from cloudmonitoring.go or metric.go. But I may still use Config to extract resources from push.Option. More details are in my design doc.

james-bebbington: 👍  - commented on design doc

PR #56, opened 2020-06-29
Added create trace pipeline method (Breaking Change: removes implementation of "SpanBatcher")

stevencl1013:
Pipeline method to create exporter and provider in one method. Also updated tests, example, and readme to use pipeline.

This PR also removes the `ExportSpans` method, so the provider can no longer use the `withBatcher` option when creating the exporter.
(NOTE: THIS IS A BREAKING CHANGE)

### DIFF
--- a/example/trace/http/client/client.go
+++ b/example/trace/http/client/client.go
@@ -39,12 +39,11 @@ func initTracer() func() {
 
 	// Create Google Cloud Trace exporter to be able to retrieve
 	// the collected spans.
-	_, flush, err := texporter.NewExportPipeline(
-		texporter.WithProjectID(projectID),
+	_, flush, err := texporter.InstallNewPipeline(
+		[]texporter.Option {texporter.WithProjectID(projectID)},
 		// For the demonstration, use sdktrace.AlwaysSample sampler to sample all traces.
 		// In a production application, use sdktrace.ProbabilitySampler with a desired probability.
-		texporter.WithSDKConfig(&sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
-		texporter.RegisterAsGlobal(),
+		sdktrace.WithConfig(sdktrace.Config{DefaultSampler: sdktrace.AlwaysSample()}),
 	)
 	if err != nil {
 		log.Fatal(err)
--- a/example/trace/http/client/client.go
+++ b/example/trace/http/client/client.go
@@ -22,7 +22,6 @@ import (
 	"os"
 
 	"net/http"
-	"time"
 
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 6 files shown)


example/trace/http/server/server.go
```diff
@@ -28,30 +28,26 @@ import (
...
-	if err != nil {
-		log.Fatal(err)
-	}
-	global.SetTraceProvider(tp)
+	return flush
 }
 
 func main() {
-	initTracer()
+	_ = initTracer()
```
james-bebbington, line 50:
```suggestion
	fn := initTracer()
	defer fn()
```

exporter/trace/cloudtrace.go
```diff
@@ -229,6 +253,44 @@ type Exporter struct {
 	traceExporter *traceExporter
 }
 
+// NewExportPipeline sets up a complete export pipeline with the recommended setup 
+// for trace provider. Returns provider, flush function, and errors.
+func NewExportPipeline(opts ...Option) (apitrace.Provider, func(), error) {
```
james-bebbington, line 258: This doesn't include an option to set the resource of the trace provider as per `WithResource`: https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/provider.go#L211

In general, I think you followed the Jaeger example, but it might be better to look at how a similar function is implemented in the operations metric exporter, see: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/master/exporter/metric/metric.go#L115

It allows you to pass in sdk `push.Options` and applies them directly. You could do something similar here, but for traces, meaning you provide a way for users to specify all current and future sdktrace options (you wouldn't need to add SDKConfig, Resource, etc).

i.e. something like:
```go
func NewExportPipeline(opts []Option, topts ...sdktrace.Option) (*apitrace.Provider, func(), error) {
	...

	tp, err := sdktrace.NewProvider(
		append([]sdktrace.Option{
			sdktrace.WithSyncer(&Exporter{traceExporter: te},
		}, topts...)...,
	)
```

It would also be nice to separate the registration of the trace provider in an `InstallNewPipeline` function similar to what's done in the metric exporter.

stevencl1013:
I keep running into this issue with the timeout test. It passes on my machine when I run `make precommit`, but for some reason in Circle it fails because the rpc code is `Unknown` instead of `DeadlineExceeded`. Not sure what's causing it

example/trace/http/client/client.go
```diff
@@ -35,30 +34,26 @@ import (
...
+func initTracer() func() {
 	projectID := os.Getenv("PROJECT_ID")
 
 	// Create Google Cloud Trace exporter to be able to retrieve
 	// the collected spans.
-	exporter, err := texporter.NewExporter(
-		texporter.WithProjectID(projectID),
+	_, flush, err := texporter.InstallNewPipeline(
+		[]texporter.Option {texporter.WithProjectID(projectID)},
+		// For the demonstration, use sdktrace.AlwaysSample sampler to sample all traces.
```
james-bebbington, line 44: Should it be
```suggestion
		// For this example code we use sdktrace.AlwaysSample sampler to sample all traces.
```

example/trace/http/server/server.go
```diff
@@ -28,30 +28,26 @@ import (
...
+func initTracer() func() {
 	projectID := os.Getenv("PROJECT_ID")
 
 	// Create Google Cloud Trace exporter to be able to retrieve
 	// the collected spans.
-	exporter, err := cloudtrace.NewExporter(
-		cloudtrace.WithProjectID(projectID),
+	_, flush, err := cloudtrace.InstallNewPipeline(
+		[]cloudtrace.Option {cloudtrace.WithProjectID(projectID)},
+		// For the demonstration, use sdktrace.AlwaysSample sampler to sample all traces.
```
james-bebbington, line 38:
```suggestion
		// For this example code we use sdktrace.AlwaysSample sampler to sample all traces.
```

exporter/trace/README.md
```diff
@@ -21,28 +21,31 @@ Add `github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace`
...
-// Create exporter.
+// Create exporter and trace provider pipeline, and register provider.
 projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
-exporter, err := texporter.NewExporter(texporter.WithProjectID(projectID))
+_, flush, err := texporter.InstallNewPipeline(
+        []texporter.Option {
+                texporter.WithProjectID(projectID),
+                // other optional exporter options
+        },
+        // For the demonstration, use sdktrace.AlwaysSample sampler to sample all traces.
```
james-bebbington, line 34:
```suggestion
        // This example code uses sdktrace.AlwaysSample sampler to sample all traces.
```

james-bebbington:
> I keep running into this issue with the timeout test. It passes on my machine when I run `make precommit`, but for some reason in Circle it fails because the rpc code is `Unknown` instead of `DeadlineExceeded`. Not sure what's causing it

That's weird. I just reran the build and it worked 😕

PR #57, opened 2020-07-01
Update metric exporter to v0.7.0 compat

marvinkite:


marvinkite:
@googlebot I signed it!

exporter/metric/metric.go
```diff
@@ -147,7 +148,8 @@ func (me *metricExporter) ExportMetrics(ctx context.Context, cps export.Checkpoi
 // if the descriptor is not registered in Cloud Monitoring yet.
 func (me *metricExporter) exportMetricDescriptor(ctx context.Context, cps export.CheckpointSet) error {
 	mds := make(map[key]*googlemetricpb.MetricDescriptor)
-	aggError := cps.ForEach(func(r export.Record) error {
+	//TODO: check if it is export.PassThroughExporter
```
nilebox, line 151: This seems trivial to implement in this PR?

exporter/metric/metric.go
```diff
@@ -419,17 +419,11 @@ func countToTypeValueAndTimestamp(count aggregator.Count, kind apimetric.NumberK
 				DoubleValue: float64(value),
 			},
 		}, t, nil
-	case apimetric.Uint64NumberKind:
```
nilebox, line 422: Note to myself: support for Uint64 was removed in https://github.com/open-telemetry/opentelemetry-go/issues/851

exporter/metric/metric_test.go
```diff
@@ -75,9 +75,9 @@ func TestRecordToMpb(t *testing.T) {
 
 	desc := apimetric.NewDescriptor("testing", apimetric.ValueRecorderKind, apimetric.Float64NumberKind)
 
-	lvagg := lastvalue.New()
+	lvagg := &lastvalue.Aggregator{}
```
nilebox, line 78: It might be better to use a constructor function here
```suggestion
	lvagg := &lastvalue.New(1)[0]
```

exporter/metric/metric_test.go
```diff
@@ -75,9 +75,9 @@ func TestRecordToMpb(t *testing.T) {
...
 	aggtest.CheckedUpdate(t, lvagg, metric.NewFloat64Number(12.34), &desc)
-	lvagg.Checkpoint(ctx, &desc)
+	_ = lvagg.Update(ctx, 1, &desc)
```
nilebox, line 80:
```suggestion
	err := lvagg.Update(ctx, 1, &desc)
	assert.NoError(t, err)
```

exporter/metric/metric.go
marvinkite: I don't have enough understanding of otel and Google Metric so I just made my best guess here. 

From otel documentation:

```go
const (
	// CumulativeExporter indicates that the Exporter expects a
	// Cumulative Aggregation.
	CumulativeExporter ExportKind = 1 // e.g., Prometheus

	// DeltaExporter indicates that the Exporter expects a
	// Delta Aggregation.
	DeltaExporter ExportKind = 2 // e.g., StatsD

	// PassThroughExporter indicates that the Exporter expects
	// either a Cumulative or a Delta Aggregation, whichever does
	// not require maintaining state for the given instrument.
	PassThroughExporter ExportKind = 4 // e.g., OTLP
)
```

In this file a bit further down:

```go
	mkind := r.Descriptor().MetricKind()

	switch mkind {
	case apimetric.CounterKind, apimetric.UpDownCounterKind:
		kind = googlemetricpb.MetricDescriptor_CUMULATIVE
	case apimetric.ValueObserverKind, apimetric.ValueRecorderKind:
		kind = googlemetricpb.MetricDescriptor_GAUGE
	// TODO: Add support for SumObserverKind and UpDownSumObserverKind.
	// https://pkg.go.dev/go.opentelemetry.io/otel@v0.6.0/api/metric?tab=doc#Kind
	default:
		kind = googlemetricpb.MetricDescriptor_METRIC_KIND_UNSPECIFIED
	}
```

My best guess would be the `PassThroughExporter` I may be wrong this is why I put there the TODO note. Any advise?

marvinkite:
@googlebot I consent. 

There is no other authors on this comment.

exporter/metric/metric.go
james-bebbington: From my understanding, the different exporter Kinds are for getting the SDK to handle converting between cumulative & delta metrics - e.g. Prometheus only supports cumulative metrics so always has to convert any OT metrics that are generated as DELTAs into CUMULATIVEs

While Cloud Monitoring does support DELTA metrics, it doesn't look like this is supported for custom metrics as per: https://cloud.google.com/monitoring/custom-metrics/creating-metrics#auto-creation

This may warrant a wider discussion around opening up support for custom metric DELTAs in Cloud Monitoring. @nilebox & I can take up this discussion, but for now I believe we will need to  change this to `export.CumulativeExporter`

In addition to the above, I think we have another problem in the other code you referenced (not a blocker for merging this PR):

```go
case ..., apimetric.UpDownCounterKind:
		kind = googlemetricpb.MetricDescriptor_CUMULATIVE
```

CUMULATIVEs only support always increasing values, so won't be able to represent up down counters. One possible solution is to just use a Gauge for these types for now, but again we probably need a wider discussion with Cloud Monitoring to determine how to best represent this metric type.

More info: https://cloud.google.com/monitoring/api/v3/kinds-and-types

james-bebbington:
Thanks for your work on this PR! See my comment above re DELTAs vs CUMULATIVEs. Are you able to make that change and run a test app to validate this works correctly?

If not, we should be able to get someone from Google to look into this in the coming weeks.

exporter/metric/metric.go
james-bebbington: To update this, the reason this worked before is that the `push.WithStateful(true)` option above previously did the conversion from DELTA to CUMULATIVE

gzhao408:
@marvinkite @james-bebbington I've rebased and squashed this in #73, left @marvinkite as author

PR #63, opened 2020-07-06
inject labels

YANYZP:
Put the logic of injecting labels into span in the trace_proto.go because
1) The semantic of injecting labels is on the level of building spans, not implementing trace exporter.
2) func protoFromSpanData would be used by both func ExportSpan() and func ExportSpans() in trace.go. It would have  duplicate callings if we put the code in trace.go or cloudtrace.go.

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -72,9 +72,14 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
 }
 
 func injectLabelsFromResources(sd *export.SpanData) {
-	if sd.Resource.Len() > 0 {
-		for _, ele := range sd.Resource.Attributes() {
-			sd.Attributes = append(sd.Attributes, ele)
+	existingAttrs := make(map[kv.KeyValue]bool)
+	for _, attr := range sd.Attributes {
+		existingAttrs[attr] = true	
+	}
+	for _, attr := range sd.Resource.Attributes() {
+		if !existingAttrs[attr] {
+			existingAttrs[attr] = true	
+			sd.Attributes = append(sd.Attributes, attr)
 		}
 	}
 }
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -75,15 +75,16 @@ func injectLabelsFromResources(sd *export.SpanData) {
 	if sd.Resource.Len() == 0 {
 		return
 	}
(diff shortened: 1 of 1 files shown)


YANYZP:
Tested it by client.go on GCE and got non-duplicate results

exporter/trace/trace_proto.go
```diff
@@ -71,11 +71,29 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
 	}
 }
 
+func injectLabelsFromResources(sd *export.SpanData) {
+	if sd.Resource.Len() == 0 {
+		return
+	}
+	uniqueAttrs := make(map[kv.KeyValue]bool)
```
nilebox, line 78: AFAIU the goal is to have a set of unique keys, and skip resource attributes which conflict with the span attributes.
If so, you should use a string key as a key in this map, rather than entire KeyValue struct.
Please correct me if I misunderstood the purpose of this map.

exporter/trace/trace_proto.go
```diff
@@ -71,11 +71,29 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
...
+	for _, attr := range sd.Attributes {
+		uniqueAttrs[attr] = true	
+	}
+	for _, attr := range sd.Resource.Attributes() {
+		if !uniqueAttrs[attr] {
```
nilebox, line 83: It's more readable to use indentation for non-happy path. See article https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

So I suggest to rewrite this block as follows:
```go
if uniqueAttrs[attr] {
  continue // skip resource attributes which conflict with span attributes
}

uniqueAttrs[attr] = true
sd.Attibutes = append(sd.Attributes, attr)
```

exporter/trace/trace_proto.go
nilebox: You can also preallocate the map capacity as
```
make(map[...]bool, len(sd.Attributes)
```
since we now that we will at least add all span attributes to the map

YANYZP: Sounds good. Will do

YANYZP:
Ready for final review

nilebox:
@YANYZP Could you please send a separate PR with unit tests for this code?

YANYZP:
> Could you please send a separate PR with unit tests for this code?

Sure, will do.

PR #65, opened 2020-07-09
covered test for injecting spans

YANYZP:
test for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/63

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -72,7 +72,7 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
 }
 
 // If there are duplicate keys present in the list of attributes,
-// then the last value found for the key is preserved.
+// then the first value found for the key is preserved.
 func injectLabelsFromResources(sd *export.SpanData) {
 	if sd.Resource.Len() == 0 {
 		return
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -71,22 +71,21 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
 	}
 }
 
-
-// If there are duplicate keys present in the list of attributes, 
-// then the first value found for the key is preserved.
+// If there are duplicate keys present in the list of attributes,
+// then the last value found for the key is preserved.
 func injectLabelsFromResources(sd *export.SpanData) {
 	if sd.Resource.Len() == 0 {
 		return
 	}
 	uniqueAttrs := make(map[kv.Key]bool, len(sd.Attributes))
 	for _, attr := range sd.Attributes {
-		uniqueAttrs[attr.Key] = true	
+		uniqueAttrs[attr.Key] = true
 	}
 	for _, attr := range sd.Resource.Attributes() {
 		if uniqueAttrs[attr.Key] {
 			continue // skip resource attributes which conflict with span attributes
 		}
-		uniqueAttrs[attr.Key] = true	
+		uniqueAttrs[attr.Key] = true
 		sd.Attributes = append(sd.Attributes, attr)
 	}
 }
--- a/exporter/trace/trace_proto_test.go
+++ b/exporter/trace/trace_proto_test.go
@@ -28,40 +28,43 @@ func TestInjectLabelsFromResources(t *testing.T) {
 		input    export.SpanData
 		expected export.SpanData
 	}{
+		// empty resource
 		{
-			export.SpanData{
+			input: export.SpanData{
 				Resource: resource.New(),
 				Attributes: []kv.KeyValue{
 					kv.String("a", "1"),
 				},
 			},
-			export.SpanData{
+			expected: export.SpanData{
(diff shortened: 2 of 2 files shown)


YANYZP:
Ready for final review

exporter/trace/trace_proto.go
```diff
@@ -71,22 +71,21 @@ func generateDisplayName(s *export.SpanData, format DisplayNameFormatter) string
 	}
 }
 
-
-// If there are duplicate keys present in the list of attributes, 
-// then the first value found for the key is preserved.
+// If there are duplicate keys present in the list of attributes,
+// then the last value found for the key is preserved.
 func injectLabelsFromResources(sd *export.SpanData) {
```
YANYZP, line 76: Even though the logic is to preserve the first value, the result would be the last value because Resource package would overwrite the values. https://github.com/open-telemetry/opentelemetry-go/blob/v0.7.0/sdk/resource/resource.go#L40

nilebox: You're not invoking the `New` function here, so I don't understand why did you rewrite the comment. In this function the first value is preserved, no?

exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+		{
+			export.SpanData{
+				Resource:   resource.New(kv.String("b", "2")),
+				Attributes: []kv.KeyValue{},
+			},
+			export.SpanData{
+				Resource: resource.New(kv.String("b", "2")),
+				Attributes: []kv.KeyValue{
+					kv.String("b", "2"),
```
nilebox, line 53: This value should be different from the one in the Resource, to be able to test which one wins.

exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,117 @@
...
+			export.SpanData{
+				Resource: resource.New(),
+				Attributes: []kv.KeyValue{
+					kv.String("a", "1"),
+				},
+			},
+			export.SpanData{
+				Resource: resource.New(),
+				Attributes: []kv.KeyValue{
+					kv.String("a", "1"),
```
nilebox, line 41: This is exact copy of the span above. Maybe make values different?

exporter/trace/trace_proto_test.go
nilebox: Sorry, just realized that this is "expected".
Please add field names everywhere to make this easier to read / less error prone, i.e.
```
{
  input: ...
  expected: ....
}
```

exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"testing"
+
+	export "go.opentelemetry.io/otel/sdk/export/trace"
+	"go.opentelemetry.io/otel/sdk/resource"
+
+	"go.opentelemetry.io/otel/api/kv"
+)
+
+func TestInjectLabelsFromResources(t *testing.T) {
+	testcases := []struct {
```
nilebox, line 27: I don't see any test cases with conflict between resource and span attributes, e.g.
```go
			export.SpanData{
				Resource: resource.New(kv.String("a", "resource-value")),
				Attributes: []kv.KeyValue{
					kv.String("a", "attribute-value"),
				},
			},

```

exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,117 @@
...
+					kv.String("b", "2"),
+				},
+			},
+		},
+		{
+			export.SpanData{
+				Resource: resource.New(
+					kv.String("c", "1"),
+					kv.String("b", "1"),
+					kv.String("b", "2"),
```
nilebox, line 77: there's no need to test duplicate keys in `resource.New`, it should be already tested in `opentelemetry-go`.

nilebox: Instead, you may add a test case where `Attributes` slice below has duplicate keys, since you construct it manually.

exporter/trace/trace_proto_test.go
YANYZP: No problem

YANYZP:
Ready for final review

exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,138 @@
...
+					kv.String("b", "3"),
+					kv.String("c", "1"),
+				},
+			},
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.input.Resource.String(), func(t *testing.T) {
```
james-bebbington, line 117: It's a bit weird to make the resource string the test case name here. A couple of your tests have the same resource so presumably the same test case name.

It would be better to make your comment for each case the name instead, i.e.

```go
testcases := []struct {
	name     string
	input    export.SpanData
	expected export.SpanData
}{
	{
		name: empty resource,
		input: ...
...
}

...

t.Run(tc.name, ...
```

YANYZP:
Real Final review

PR #66, opened 2020-07-09
Use mock server in the cloudtrace tests

the-ericwang35:
As part of the project to create a mock server for exporters to test against (see [here](https://docs.google.com/document/d/1ZUwuqKIcTjrR32oxKwJw8XFNThQ8qDm_KvmTzzgiwtQ/edit) for more details, and [here](https://github.com/googleinterns/cloud-operations-api-mock) for the repository), we'd like to integrate our mock server with the Go exporter tests. This PR integrates our mock server into the cloud trace tests.

In the future, we'd like to integrate this into exporters of all languages in order to allow for unified tests.


exporter/trace/cloudtrace_test.go
```diff
@@ -16,94 +16,29 @@ package trace_test
...
-	os.Exit(m.Run())
-}
-
 func TestExporter_ExportSpan(t *testing.T) {
 	// Initial test precondition
-	mockTrace.spansUploaded = nil
-	mockTrace.delay = 0
+	mock := cloudmock.NewCloudMock()
+	defer mock.Shutdown()
+	clientOpt := []option.ClientOption{option.WithGRPCConn(mock.ClientConn())}
```
nilebox, line 41: It's nice that you're able to just pass the client connection from mock to client in Go here, but other languages won't have this luxury, as you mentioned in the description

> In the future, we'd like to integrate this into exporters of all languages in order to allow for unified tests.

Will the mock have a fixed port or some other way of discovering the endpoint after starting it?
No need to change anything, just wanted to clarify the mock "contract"

exporter/trace/cloudtrace_test.go
```diff
@@ -154,14 +90,16 @@ func TestExporter_DisplayNameFormatter(t *testing.T) {
...
+	assert.EqualValues(t, "TEST_FORMAT"+spanName, mock.GetSpan(0).DisplayName.Value)
 }
 
 func TestExporter_Timeout(t *testing.T) {
 	// Initial test precondition
-	mockTrace.spansUploaded = nil
-	mockTrace.delay = 20 * time.Millisecond
+	mock := cloudmock.NewCloudMock()
+	defer mock.Shutdown()
+	mock.SetDelay(20 * time.Millisecond)
```
nilebox, line 101: This will also be impossible to set in other languages.
Ideally there should be a Docker image with flags to make this functionality portable to other languages.

exporter/trace/cloudtrace_test.go
the-ericwang35: Yeah, for other languages we're currently exploring two options. 
1. We have a standalone server mode, similar to [Moto](https://github.com/spulec/moto#stand-alone-server-mode), where the user will spin up the server manually and optionally pass in the address (there's a default address if nothing is passed in). In the exporter test, they'll then point the exporter to the address they choose. After discussions with the Python exporter team, this is what we were planning on doing (they'll spin up the server in their CI and point their exporter to our mock server). 
2. It is possible to call Go functions from other languages (see [here](https://medium.com/learning-the-go-programming-language/calling-go-functions-from-other-languages-4c7d8bcc69bf)), so if we go this route, they could still use this `cloudmock` library to spin up the server, pass the client connection, etc. but this would probably be a bit more difficult than the first option.

exporter/trace/cloudtrace_test.go
the-ericwang35: As mentioned above, we were thinking of doing something like [this](https://medium.com/learning-the-go-programming-language/calling-go-functions-from-other-languages-4c7d8bcc69bf) to allow other languages to use our `cloudmock` library. We could also turn `SetDelay` and other functions into gRPC endpoints on our server so that other languages can still call it. We hadn't really thought of creating a Docker image, but it's definitely something we can consider.

exporter/trace/cloudtrace_test.go
james-bebbington: I would definitely recommend going with option 1 😄

I think we should also consider pros / cons of updating these Go tests to spin up the Fake server out-of-process so its consistent with the other exporters in the future, but this looks good for now.

exporter/trace/cloudtrace_test.go
james-bebbington: I would think having endpoints or command flags to set configuration would be a lot simpler than calling into Go code from another language. From you comment above it doesn't look like you are going that way, as I don't think you can both "spin up the server separately" and "call into Go code" (for a separate process)

Docker would be great for portability / modularity but not essential.

PR #67, opened 2020-07-14
Integrate go exporter metric tests with mock metric api

mabdi3:
As described in the [proposal](https://docs.google.com/document/d/1ZUwuqKIcTjrR32oxKwJw8XFNThQ8qDm_KvmTzzgiwtQ/edit#heading=h.h76dy5b7dqq2), we want to integrate those writing exporters with our mock api instead having to call the google cloud libraries directly. In short, I simply replaced the current client calls with calls to our mock client.

### DIFF
--- a/exporter/metric/metric_test.go
+++ b/exporter/metric/metric_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/googleinterns/cloud-operations-api-mock/cloudmock"
+	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/otel/api/kv"
 	"go.opentelemetry.io/otel/api/metric"
 	apimetric "go.opentelemetry.io/otel/api/metric"
@@ -28,9 +29,10 @@ import (
 	export "go.opentelemetry.io/otel/sdk/export/metric"
 	"go.opentelemetry.io/otel/sdk/metric/aggregator/lastvalue"
 	aggtest "go.opentelemetry.io/otel/sdk/metric/aggregator/test"
+	"go.opentelemetry.io/otel/sdk/metric/controller/push"
 	"go.opentelemetry.io/otel/sdk/resource"
-	"google.golang.org/api/option"
 
+	"google.golang.org/api/option"
 	googlemetricpb "google.golang.org/genproto/googleapis/api/metric"
 )
 
@@ -40,24 +42,75 @@ var (
 	}
 )
 
-func TestDescToMetricType(t *testing.T) {
+func TestExportMetrics(t *testing.T) {
 	cloudMock := cloudmock.NewCloudMock()
 	defer cloudMock.Shutdown()
 
-	clientOpt := []option.ClientOption{option.WithGRPCConn(cloudMock.ClientConn())}
-	o := options{
-		Context:                 context.Background(),
-		MonitoringClientOptions: clientOpt,
-		ProjectID:               "PROJECT_ID_NOT_REAL",
+	clientOpt := option.WithGRPCConn(cloudMock.ClientConn())
+	res := &resource.Resource{}
+	cps := test.NewCheckpointSet(res)
+	ctx := context.Background()
+	desc := apimetric.NewDescriptor("testing", apimetric.ValueRecorderKind, apimetric.Float64NumberKind)
+
+	lvagg := lastvalue.New()
+	aggtest.CheckedUpdate(t, lvagg, metric.NewFloat64Number(12.34), &desc)
+	lvagg.Checkpoint(ctx, &desc)
+	cps.Add(&desc, lvagg, kv.String("a", "A"), kv.String("b", "B"))
+
+	opts := []Option{
+		WithProjectID("PROJECT_ID_NOT_REAL"),
+		WithMonitoringClientOptions(clientOpt),
+		WithMetricDescriptorTypeFormatter(formatter),
 	}
 
-	me, err := newMetricExporter(&o)
+	exporter, err := NewRawExporter(opts...)
 	if err != nil {
-		t.Errorf("Error occurred when creating metric exporter: %v", err)
+		t.Errorf("Error occurred when creating exporter: %v", err)
 	}
 
+	err = exporter.Export(ctx, cps)
+	assert.NoError(t, err)
+}
+
+func TestExportCounter(t *testing.T) {
+	cloudMock := cloudmock.NewCloudMock()
+	defer cloudMock.Shutdown()
+
+	clientOpt := option.WithGRPCConn(cloudMock.ClientConn())
+
+	resOpt := push.WithResource(
+		resource.New(
+			kv.String("test_id", "abc123"),
+		),
+	)
+
+	pusher, err := InstallNewPipeline(
+		[]Option{
+			WithProjectID("PROJECT_ID_NOT_REAL"),
+			WithMonitoringClientOptions(clientOpt),
+			WithMetricDescriptorTypeFormatter(formatter),
+		},
+		resOpt,
+	)
+	assert.NoError(t, err)
+
+	defer pusher.Stop()
+
+	// Start meter
+	ctx := context.Background()
+	meter := pusher.Provider().Meter("cloudmonitoring/test")
+
+	// Register counter value
+	counter := metric.Must(meter).NewInt64Counter("counter-a")
+	clabels := []kv.KeyValue{kv.Key("key").String("value")}
+	counter.Add(ctx, 100, clabels...)
+}
+
+func TestDescToMetricType(t *testing.T) {
 	inMe := []*metricExporter{
-		me,
+		{
+			o: &options{},
+		},
 		{
 			o: &options{
 				MetricDescriptorTypeFormatter: formatter,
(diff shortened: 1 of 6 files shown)


exporter/metric/metric_test.go
```diff
@@ -39,14 +40,20 @@ var (
 )
 
 func TestDescToMetricType(t *testing.T) {
+	cloudMock := cloudmock.NewCloudMock()
+	client := cloudMock.MetricServiceClient
```
nilebox, line 44: This is different from the approach in #66 where `[]option.ClientOption{option.WithGRPCConn(mock.ClientConn())}` is used instead.
Shall we use the same approach in both PRs for consistency (whichever is preferrable)?

exporter/metric/metric_test.go
```diff
@@ -39,14 +40,20 @@ var (
 )
 
 func TestDescToMetricType(t *testing.T) {
+	cloudMock := cloudmock.NewCloudMock()
```
nilebox, line 43: Where do you shut down this mock? See `defer mock.Shutdown()` in #66

exporter/metric/metric_test.go
```diff
@@ -39,14 +40,20 @@ var (
...
+	client := cloudMock.MetricServiceClient
 	inMe := []*metricExporter{
 		{
-			o: &options{},
+			o:         &options{},
+			cloudMock: cloudmock.NewCloudMock(),
```
nilebox, line 48: You have just created a mock above. Why do you need a new mock for every exporter here?

exporter/metric/metric_test.go
```diff
@@ -65,10 +72,14 @@ func TestDescToMetricType(t *testing.T) {
 		if out != w {
 			t.Errorf("expected: %v, actual: %v", w, out)
 		}
+		inMe[i].cloudMock.Shutdown()
```
nilebox, line 75: AFAICT, this is the only use of the `cloudMock` field? Do you actually need it?

exporter/metric/metric_test.go
```diff
@@ -65,10 +72,14 @@ func TestDescToMetricType(t *testing.T) {
...
 	}
 }
 
 func TestRecordToMpb(t *testing.T) {
+	cloudMock := cloudmock.NewCloudMock()
+	client := cloudMock.MetricServiceClient
+	defer cloudMock.Shutdown()
```
nilebox, line 82: this should be one line up, right after creating the mock.

exporter/metric/metric.go
```diff
@@ -66,7 +65,9 @@ type metricExporter struct {
 	// TODO: [ymotongpoo] this map should be goroutine safe with mutex.
 	mdCache map[key]*googlemetricpb.MetricDescriptor
 
-	client *monitoring.MetricClient
+	client monitoringpb.MetricServiceClient
+
+	cloudMock *cloudmock.CloudMock
```
nilebox, line 70: After reading the code, it seems that you don't really need this field, just client should be enough.

exporter/metric/metric_test.go
```diff
@@ -97,6 +108,8 @@ func TestRecordToMpb(t *testing.T) {
 		mdCache: map[key]*googlemetricpb.MetricDescriptor{
 			mdkey: md,
 		},
+		cloudMock: cloudMock,
```
nilebox, line 111: No need to pass the mock, it's unused.

exporter/metric/metric_test.go
mabdi3: I'm not too sure about the usability of the approach in #66 here since the metric exporters here are defined literally by their respective structs and not by using the `newMetricExporter` function defined [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/15b5fb3c8d62367a9f9333ee86b5fbfe0c7e7670/exporter/metric/metric.go#L78). 

Seems to me there is a way to proceed, one involving reverting the changes to `newMetricExporter` and passing in the correct clientOptions to direct requests to our mock api library. After looking closer, it seems newMetricExporter is the more "commerical" method of creating an exporter where one would actually want to connect to the google cloud instance instead of our mock api. Correct me if I'm misunderstanding, but I shouldn't have touched that file entirely.

exporter/metric/metric_test.go
mabdi3: Nice catch, added a shutdown call below.

exporter/metric/metric_test.go
nilebox: Mocking should be still as close to the actual code being used as possible, so it's preferable to use constructor functions over directly creating structs when possible.

If I understand correctly, using `newMetricExporter` will increase the code coverage in tests, and will test the actual code executed when the metric exporter is being initialized in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/8fcf64591886d64b6c07975ceed0828fb63a8920/exporter/metric/cloudmonitoring.go#L73

It's fine to merge the change as is, and then you may consider creating another PR switching to `newMetricExporter`.

mabdi3: I agree, thanks a lot.

exporter/metric/metric.go
```diff
@@ -80,15 +79,8 @@ func newMetricExporter(o *options) (*metricExporter, error) {
...
-	clientOpts := append(o.MonitoringClientOptions, option.WithUserAgent(userAgent))
-	ctx := o.Context
-	if ctx == nil {
-		ctx = context.Background()
-	}
-	client, err := monitoring.NewMetricClient(ctx, clientOpts...)
-	if err != nil {
-		return nil, err
-	}
+	cloudMock := cloudmock.NewCloudMock()
```
nilebox, line 82: This is a production code, not test. You shouldn't be using mocks in the production code, so please revert this change.

Instead you should probably have a test invoking `newMetricExporter` with `clientOpts` similar to approach in #66. @the-ericwang35 could you please assist Mohamed with that if you have time?

mabdi3: > This is a production code, not test. You shouldn't be using mocks in the production code, so please revert this change.
> 
> Instead you should probably have a test invoking `newMetricExporter` with `clientOpts` similar to approach in #66.

I agree, as stated in the conversation [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/67#discussion_r456017469). I'm assuming you missed the change/my comment when you suggested merging the code as is. I reverted all the changes on `metric.go`, and instead used `newMetricExporter` where appropriate.

exporter/metric/metric_test.go
```diff
@@ -39,10 +41,23 @@ var (
 )
 
 func TestDescToMetricType(t *testing.T) {
+	cloudMock := cloudmock.NewCloudMock()
```
nilebox, line 44: This test only invokes the `descToMetricType()` function, i.e. it doesn't export anything, so setting up a mock seems unnecessary (i.e. you can just revert changes in this test).

exporter/metric/metric_test.go
```diff
@@ -73,6 +88,16 @@ func TestRecordToMpb(t *testing.T) {
 	cps := test.NewCheckpointSet(res)
 	ctx := context.Background()
 
+	cloudMock := cloudmock.NewCloudMock()
```
nilebox, line 91: Similarly, this test only invokes the `recordToMpb` and doesn't actually export anything, so adding mock is redundant.

exporter/metric/metric_test.go
```diff
@@ -38,6 +43,40 @@ var (
...
+	defer pusher.Stop()
+
+	// Start meter
+	ctx := context.Background()
+	meter := pusher.Provider().Meter("cloudmonitoring/test")
+
+	// Register counter value
+	counter := metric.Must(meter).NewInt64Counter("counter-a")
+	clabels := []kv.KeyValue{kv.Key("key").String("value")}
+	counter.Add(ctx, 100, clabels...)
```
nilebox, line 77: Would be nice to add a function for metrics similar to [GetNumSpans()](https://github.com/googleinterns/cloud-operations-api-mock/blob/d3cb41cc27b43ba96ca5125afc13de3161783e94/cloudmock/cloud_mock.go#L89-L90) to be able to validate that the correct data appeared on the other end.

exporter/metric/metric_test.go
```diff
@@ -28,8 +29,12 @@ import (
 	"go.opentelemetry.io/otel/sdk/metric/aggregator/lastvalue"
 	aggtest "go.opentelemetry.io/otel/sdk/metric/aggregator/test"
 	"go.opentelemetry.io/otel/sdk/resource"
+	"google.golang.org/api/option"
 
 	googlemetricpb "google.golang.org/genproto/googleapis/api/metric"
+
```
nilebox, line 35: The import sorting/grouping seems random, please fix

exporter/metric/metric_test.go
nilebox: Alternatively, you may add another test which will construct the exporter via https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4866baabe07e06199c33b0a3ca8aca46bb9441d7/exporter/metric/cloudmonitoring.go#L49-L50 and invoke https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4866baabe07e06199c33b0a3ca8aca46bb9441d7/exporter/metric/metric.go#L174 directly and assert that there is no error

exporter/metric/metric_test.go
mabdi3: Done

exporter/metric/metric_test.go
```diff
@@ -20,15 +20,20 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/googleinterns/cloud-operations-api-mock/cloudmock"
+	"github.com/stretchr/testify/assert"
+
```
nilebox, line 25: nit: these imports should be together with the one below (i.e. no empty line)

james-bebbington: Surprising that the linter didn't pick this up. Maybe we need to add some more CI steps

PR #75, opened 2020-07-31
Create example using mock server

stevencl1013:
Can be used for testing without needing to run in a GCP environment or authenticate.


example/trace/mock-server/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/googleinterns/cloud-operations-api-mock/cloudmock"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+
+	cloudtrace "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace"
+)
+
+func main() {
+	// create new mock server
+	mock := cloudmock.NewCloudMock()
```
nilebox, line 33: IMO it would be more useful for the end users if we documented a way to start a standalone mock server, and setting up their pipeline to forward spans to it.
e.g.
> "For local testing, you can start a mock server as described in https://github.com/googleinterns/cloud-operations-api-mock#stand-alone-server-mode and set the host/port in the client as follows: {code snippet}

We don't expect end users to write code for mocking inside of their applications.

@james-bebbington what do you think?

james-bebbington: I was just writing up a very similar comment - I totally agree with your suggestion

Issue #23, opened 2020-05-11
Add Readme.md for examples


rghetia:
Create single readme or one for each example in example directory.
Readme should contain
- Brief description about the example.
- how to run
- Expected output.

ymotongpoo:
README.md for the exporter example was added. Need one for Trace exporter.

james-bebbington:
We should include documentation for how to use multiple providers, see https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/76#issuecomment-667746838

punya:
> We should include documentation for how to use multiple providers, see [#76 (comment)](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/76#issuecomment-667746838)

I decided to document usage of multiple providers in a Go-style example func, rather than the README. It'll ensure the code keep compiling (samples in markdown files inevitably get outdated).

PR #76, opened 2020-07-31
Use trace provider from pipeline instead of getting it from global

stevencl1013:
Updated trace tests to use the trace provider returned from `InstallNewPipeline()` instead of always calling `global.TraceProvider()`.

IMO it looks cleaner this way, but if this is unnecessary or if there's some reason to use `global.TraceProvider()`, feel free to close this PR.


example/trace/http/client/client.go
```diff
@@ -49,13 +48,13 @@ func initTracer() func() {
...
 
 func main() {
-	flush := initTracer()
+	tp, flush := initTracer()
 	defer flush()
-	tr := global.TraceProvider().Tracer("cloudtrace/example/client")
+	tr := tp.Tracer("cloudtrace/example/client")
 
 	client := http.DefaultClient
 	ctx := correlation.NewContext(context.Background(),
```
james-bebbington, line 60: Not related to this PR. but do attributes we add to the correlation context show up anywhere? My understanding is that this is not implemented in the SDK yet, so perhaps we should remove it from the examples for now.

james-bebbington:
On second thoughts, if we are updating the README and examples, we need to think about what we would prefer to demonstrate to users. Using the global (singleton) trace provider is expected to be the most common case, and it's only rare / advanced users that would need to create multiple trace providers. I think we should at least mention both options in the README.

WDYT @nilebox?

nilebox:
> I think we should at least mention both options in the README.

Agreed. We should initially provide an example with a global trace provider since it's the most common case.

Then we can mention that if a user has more than one trace providers configured, they need to explicitly fetch and reference them. And follow with a short code snippet showing the difference.

james-bebbington:
Closing for now, improving docs tracked in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/23

Issue #70, opened 2020-07-22
Automate the release process for Go modules


nilebox:
The current manual process is error prone: we need to create additional tags for Go module subdirectories but often forget about that (see #38, #69).

See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/38#issuecomment-641433905 with links to existing `opentelemetry-go` scripts which we can reuse to address this.

PR #80, opened 2020-08-04
Added scripts to automate release process and related docs

james-bebbington:
Issue: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/70

I brought these scripts / docs over from opentelemetry-go and made the following changes:

- Removed part of pre_release script that updated the version used in code by the OpenTelemetry SDK. We don't have any equivalent of that in this repo yet.
- Updated pre_release script to update `opentelemetry-operations-go` dependencies instead of `otel` SDK dependencies
- Removed references to updating CHANGELOG file (can add this later if needed)

Other than that it was just a few minor changes re updating directory names, etc, but for the most part the scripts were able to be brought over as is.

Left for future work:

- Add script to automate updating the Otel SDK version
- Consider automatically enforcing pinning the version of the exporter to the SDK version. On a related note, should I make the next release v0.9.0 to match the SDK?
- Consider adding a CHANGELOG file (for now the changes are just documented as release notes)

See related PR: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/79

### DIFF
--- a/exporter/metric/version.go
+++ b/exporter/metric/version.go
@@ -14,7 +14,8 @@
 
 package metric
 
-// Version is the current release version of OpenTelemetry in use.
+// Version is the current release version of the OpenTelemetry
+// Operations Metric Exporter in use.
 func Version() string {
-	return "0.10.0"
+	return "0.2.1"
 }
--- a/exporter/metric/version.go
+++ b/exporter/metric/version.go
@@ -1,0 +1,20 @@
+// Copyright 2020, Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metric
+
+// Version is the current release version of OpenTelemetry in use.
+func Version() string {
+	return "0.10.0"
+}
--- a/exporter/trace/version.go
+++ b/exporter/trace/version.go
@@ -14,7 +14,8 @@
 
 package trace
 
-// Version is the current release version of OpenTelemetry in use.
+// Version is the current release version of the OpenTelemetry
+// Operations Trace Exporter in use.
 func Version() string {
 	return "0.2.1"
 }
--- a/exporter/trace/version.go
+++ b/exporter/trace/version.go
@@ -1,0 +1,20 @@
+// Copyright 2020, Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+// Version is the current release version of OpenTelemetry in use.
+func Version() string {
+	return "0.2.1"
+}
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,69 @@
+# Release Process
+
+## Pre-Release
+
+Update go.mod for submodules to depend on the new release which will happen in the next step.
+
+1. Run the pre-release script. It creates a branch `pre_release_<new tag>` that will contain all release changes.
+
+    ```
+    ./pre_release.sh -t <new tag>
+    ```
+
+2. Verify the changes.
+
+    ```
+    git diff master
+    ```
+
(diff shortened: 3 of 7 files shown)


nilebox:
> Removed part of pre_release script that updated the version used in code by the OpenTelemetry SDK. We don't have any equivalent of that in this repo yet.

Should we add it though? (via a separate issue / PR).
Java exporters do specify the exporters version in [TraceTranslator.java#L30-L33](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/b835980ef1e13e4e59ba22b865ef933c1e518470/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java#L30-L33)
And the [requirements doc](https://docs.google.com/document/d/1eeGeSYlE2F-Dc-5_yLXnIlj7e--BYFeWlzWy6qBV2cg/edit#heading=h.aoqug1blgfpl) also has the attribute `g.co/agent/component` documented.

nilebox:
> On a related note, should I make the next release v0.9.0 to match the SDK?

We could do that, but that would be different from [SemVer's](https://semver.org/) `vMINOR.MAJOR.PATCH` requirements, where we need to increment the "MINOR" version every time we add a new feature (e.g. add support for metric views), or "MAJOR" when we introduce breaking changes (e.g. rename a public function).

On the other hand, Otel Go SDK itself hasn't really followed that convention, introducing breaking changes in the "minor" releases, though that will probably change after GA / `v1.0.0`

exporter/trace/version.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+// Version is the current release version of OpenTelemetry in use.
+func Version() string {
+	return "0.2.1"
```
nilebox, line 19: Why 0.2.1 here but 0.10.0 in metrics?

exporter/metric/version.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metric
+
+// Version is the current release version of OpenTelemetry in use.
```
nilebox, line 17: I think the comment should be changed to "OpenTelemetry Metrics Exporter" or something similar?

james-bebbington:
> Java exporters do specify the exporters version in [TraceTranslator.java#L30-L33](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/b835980ef1e13e4e59ba22b865ef933c1e518470/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java#L30-L33)
> And the [requirements doc](https://docs.google.com/document/d/1eeGeSYlE2F-Dc-5_yLXnIlj7e--BYFeWlzWy6qBV2cg/edit#heading=h.aoqug1blgfpl) also has the attribute `g.co/agent/component` documented.

Have updated the script to set the VERSION number for trace & metric exporters. I'm not using this for anything yet but have created https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/82 to track using this as specified in the requirements doc.

> We could do that, but that would be different from [SemVer's](https://semver.org/) `vMINOR.MAJOR.PATCH` requirements, where we need to increment the "MINOR" version every time we add a new feature (e.g. add support for metric views), or "MAJOR" when we introduce breaking changes (e.g. rename a public function).
> 
> On the other hand, Otel Go SDK itself hasn't really followed that convention, introducing breaking changes in the "minor" releases, though that will probably change after GA / `v1.0.0`

I guess it's less important to keep versions pinned if we create the attribute you mentioned above.

FYI @aabmass in case you had any thoughts / wanted to cross reference against what you are doing in Python / JS.

james-bebbington:
In the pre_release script, I had to move the code that creates a new branch above the code that updates the VERSION number otherwise creating a new branch would fail because there are uncommitted changes ... and of course this change should happen in the new branch anyway (see the change to pre_release.sh here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/80/commits/4b8c4760810fd28c361a458d6b7240b962cea9e8). I'm not sure how this works in the Go repo; let me know if you notice I'm missing something here.

aabmass:
We are pinning the SDK commit only for development, so that we can build in parallel without waiting for a release. For the actual release, I switched it to a real release (e.g.`opentelemetry-api==0.11b0`). Using `==` because there are a lot of breaking changes. After GA, I plan to remove this. OT python is not really following semver before GA either, AFAICT.

> On a related note, should I make the next release v0.9.0 to match the SDK?

Python is doing this, but all of the other exporters/instrumentation packages are doing so as well. It also simplifies the release scripts a little if they match.

> Consider adding a CHANGELOG file (for now the changes are just documented as release notes)

The python operations repo has changelogs, but JS operations is just pulling a changelog from Git like you are doing here.

aabmass:
@james-bebbington do you have any thoughts on having your release script push to master without a PR?

james-bebbington:
> @james-bebbington do you have any thoughts on having your release script push to master without a PR?

Not sure. The extra cost of creating a PR is fairly trivial. I'll leave it like this for now but consider changing it later. If we did that we'd presumably need to change the branch permissions to allow pushing straight to upstream master which I'd prefer not to?

PR #83, opened 2020-08-04
Add pass through of status message

qnten:
When setting the the status of a span with `span.SetStatus(codes.Unknown, "Error message")`, the error message didn't get send to the server. This PR adds the status message to the grpc request if a message is given.

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -114,7 +114,10 @@ func protoFromSpanData(s *export.SpanData, projectID string, format DisplayNameF
 		sp.ParentSpanId = s.ParentSpanID.String()
 	}
 	if s.StatusCode != codes.OK {
-		sp.Status = &statuspb.Status{Code: int32(s.StatusCode)}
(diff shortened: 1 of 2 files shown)


exporter/trace/trace_proto.go
```diff
@@ -114,7 +114,10 @@ func protoFromSpanData(s *export.SpanData, projectID string, format DisplayNameF
 		sp.ParentSpanId = s.ParentSpanID.String()
 	}
 	if s.StatusCode != codes.OK {
-		sp.Status = &statuspb.Status{Code: int32(s.StatusCode)}
+		sp.Status = &statuspb.Status{Code: int32(s.StatusCode), Message: s.StatusMessage}
+	}
+	if s.StatusCode == codes.OK && s.StatusMessage != "" {
+		sp.Status = &statuspb.Status{Code: int32(s.StatusCode), Message: s.StatusMessage}
```
james-bebbington, line 120: Can this just be one condition:
```go
if s.StatusCode != codes.OK || s.StatusMessage != "" {
	...
}
```

PR #87, opened 2020-08-11
Release script fixes

james-bebbington:
- Minor changes to `tag.sh` (see comments)
- Added `gojq`, needed by `verify_examples.sh`

### DIFF
--- a/tag.sh
+++ b/tag.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 readonly PROGNAME=$(basename "$0")
-readonly PROGDIR=$(readlink -m "$(dirname "$0")")
+readonly PROGDIR=$(pwd)
 
 readonly EXCLUDE_PACKAGES="tools"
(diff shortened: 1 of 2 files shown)


tag.sh
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 readonly PROGNAME=$(basename "$0")
-readonly PROGDIR=$(readlink -m "$(dirname "$0")")
+readonly PROGDIR=$(pwd)
```
james-bebbington, line 18: I'm not super familiar with sym links. `readlink` didn't work as expected for me on Mac, so I just switched to using `pwd`. I'm not too sure why we need this at all, but let me know if I should not check this in / am missing something

tag.sh
```diff
@@ -126,7 +126,7 @@ git_tag() {
     local tag="$1"
     local commit="$2"
 
-    git tag -a "$tag" -s -m "Version $tag" "$commit"
+    git tag -a "$tag" -m "Version $tag" "$commit"
```
james-bebbington, line 129: I was having issues using `gpg` so disabled signing of the tags for now. If we think this is important, I'm happy to rollback this change

tag.sh
nilebox: Might be some legacy stuff due to $GOPATH/src being a symlink when the repo is checked out in Travis or Circle CI (I recall something related to that) or similar. Getting current dir via `pwd` seems reasonable.

PR #90, opened 2020-08-27
Update otel to v0.11.0

marvinkite:


example/trace/http/client/client.go
```diff
@@ -57,31 +59,29 @@ func main() {
...
+		label.String("username", "donuts"),
 	)
 
 	var body []byte
 
-	err := tr.WithSpan(ctx, "say hello",
-		func(ctx context.Context) error {
-			req, _ := http.NewRequest("GET", "http://localhost:7777/hello", nil)
-
-			ctx, req = httptrace.W3C(ctx, req)
```
nilebox, line 71: Looks like you have removed this bit. Is it now handled by the SDK itself or other reason?

nilebox (approved):

example/trace/http/client/client.go
marvinkite: This is how the example in OTEL. The client [transport is wrapped](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/master/instrumentation/net/http/example/client/client.go#L61). So no need for explicitly calling these.

PR #91, opened 2020-08-31
Update metrics exporter to otel v0.11.0

liiling:
Update metrics exporter and example to otel v0.11.0


exporter/metric/cloudmonitoring.go
```diff
@@ -26,7 +26,7 @@ import (
 	export "go.opentelemetry.io/otel/sdk/export/metric"
 	"golang.org/x/oauth2/google"
 
-	monitoring "cloud.google.com/go/monitoring/apiv3"
+	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
```
nilebox, line 29: This doesn't seem related to updating the otel dependency. Looks like you just updated all dependencies?

We have automation for specifically updating the SDK dependency: please update the version number in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ba72e89a40870af14a0bc190e62174ab3e86335a/Makefile#L136
and run `make update-otel`.

nilebox (changes requested):
Thanks for making this change! Could you just update the OTel SDK dependency as mentioned in the comment?

exporter/metric/cloudmonitoring.go
liiling: Thanks for the review! Just updated the SDK dependency as requested. 

For the cloud monitoring API, I updated it because of failure in CircleCI build (it complained that the API is deprecated).

nilebox (approved):

PR #92, opened 2020-09-02
Async metrics

### LINKED ISSUE #89: unable to use OpenTelemetry asynchronous metric instruments
Hi! I'm trying to create a demo using OpenTelemetry with GCP monitoring backend.

When instrumenting with OpenTelemetry's async metric instuments, only ValueObserver seems to work, while SumObserver and UpDownSumObserver both give me the following error:
> Metrics Accumulator error: rpc error: code = InvalidArgument desc = Request was missing field metricDescriptor.metricKind: The descriptor does not have the metric kind set.

Is there any plan on fixing this? I'd be happy to contribute.

### LINKED ISSUE #94: Challenged by 'hello world'
I'm unable to get the Cloud Monitoring exporter working:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/master/exporter/metric/README.md

There are appear to be a couple of issues in addition to my likely incompetence; I'm reasonable familiar with Cloud Monitoring and OpenCensus and this is my first dib into OpenTelemetry. Code at end based upon the README example. The code works with the `stdout` exporter.

I've exported:

+ `GOOGLE_APPLICATION_CREDENTIALS` to a known working Service Account with `roles/monitoring.admin`
+ `PROJECT` to a project with Stackdriver enabled

The module appears to have a bug using `InstallNewPipeline(opts, nil)`:

```console
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x972274]

goroutine 1 [running]:
go.opentelemetry.io/otel/sdk/metric/controller/push.New(0xbfa380, 0xc0001f08c0, 0xbebb60, 0xc000010740, 0xc00000fba0, 0x2, 0x2, 0x1)
	/home/dazwilkin/go/pkg/mod/go.opentelemetry.io/otel/sdk@v0.11.0/metric/controller/push/push.go:56 +0x84
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric.NewExportPipeline(0xc0002dfc50, 0x1, 0x1, 0xc0002dfc88, 0x1, 0x1, 0x4d1c01, 0xc00000f800, 0xc0002dfb98)
	/home/dazwilkin/go/pkg/mod/github.com/!google!cloud!platform/opentelemetry-operations-go/exporter/metric@v0.11.0/metric.go:166 +0x1cf
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric.InstallNewPipeline(0xc0002dfc50, 0x1, 0x1, 0xc0002dfc88, 0x1, 0x1, 0x7, 0x6, 0xc00a20)
	/home/dazwilkin/go/pkg/mod/github.com/!google!cloud!platform/opentelemetry-operations-go/exporter/metric@v0.11.0/metric.go:147 +0x6d
main.main()
	/home/dazwilkin/Projects/go-open-telemetry/main.go:25 +0xe1
exit status 2
```

This occurs in `metric.go` line 166 because the append results in len(opts)==2 with the second value being void:

```golang
pusher := push.New(
	checkpointer,
	exporter,
	append([]push.Option{
		push.WithPeriod(period),
	}, popts...)...,
)
```

Which then breaks `push.go` line 56 on the second apply:

```golang
for _, opt := range opts {
	opt.Apply(c)
}
```

I can fudge this by creating an non-nil popts:

```golang
opts := []cloudmonitoring.Option{
	cloudmonitoring.WithProjectID(projectID),
}
popts := []push.Option{}
pusher, err := cloudmonitoring.InstallNewPipeline(opts, popts...)
```

But this then results in a second failure (which bemuses me);

```console
2020/09/09 14:49:49 [main] ProjectID: [[REDACTED]]
2020/09/09 14:49:49 Error during exporting TimeSeries: currently the aggregator is not supported: &{{0 0} 1 [1]}
2020/09/09 14:49:49 rpc error: code = InvalidArgument desc = Request was missing field timeSeries.
```

### Code

```golang
package main

import (
	"context"
	"log"
	"math/rand"
	"os"
	"time"

	"go.opentelemetry.io/otel/api/metric"
	"go.opentelemetry.io/otel/label"
	"go.opentelemetry.io/otel/sdk/metric/controller/push"

	cloudmonitoring "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
)

func main() {
	projectID := os.Getenv("PROJECT")
	if projectID == "" {
		log.Fatal("Environment variable `PROJECT` is required and was unset")
	}

	log.Printf("[main] ProjectID: %s", projectID)
	opts := []cloudmonitoring.Option{
		cloudmonitoring.WithProjectID(projectID),
	}
	popts := []push.Option{}
	pusher, err := cloudmonitoring.InstallNewPipeline(opts, popts...)
	if err != nil {
		log.Fatal(err)
	}
	defer pusher.Stop()

	meter := global.Meter("cloudmonitoring/example")

	counter := metric.Must(meter).NewInt64Counter("counter.foo")

	lemonsKey := label.Key("ex.com/lemons")
	labels := []label.KeyValue{
		lemonsKey.Int(10),
		label.Key("key").String("value"),
		label.String("dog", "Freddie"),
	}

	ctx := context.Background()

	s := rand.NewSource(time.Now().UnixNano())
	r := rand.New(s)

	counter.Add(ctx, r.Int63n(100), labels...)
}
```

liiling:
- Set MetricKind for SumObserver and UpDownSumObserver (Fixes #89)
(Note: Currently both are set to CUMULATIVE to be consistent with Counter and UpDownCounter, though perhaps UpDownCounter and UpDownSumObserver should be set to GAUGE instead?)
- Change the default aggregation for metrics other than ValueObserver and ValueRecorder to Sum. Otherwise the casting to `sum` always causes the following error when using the Counter, UpDownCounter, SumObserver and UpDownSumObserver (Fixes #94):
```
error during exporting TimeSeries: currently the aggregator is not supported: &{{0 0} 1 [1]}
rpc error: code = InvalidArgument desc = Request was missing field timeSeries.
```


exporter/metric/metric.go
```diff
@@ -418,12 +418,10 @@ func recordToMdpbKindType(r *export.Record) (googlemetricpb.MetricDescriptor_Met
 
 	var kind googlemetricpb.MetricDescriptor_MetricKind
 	switch mkind {
-	case apimetric.CounterKind, apimetric.UpDownCounterKind:
+	case apimetric.CounterKind, apimetric.UpDownCounterKind, apimetric.SumObserverKind, apimetric.UpDownSumObserverKind:
```
nilebox, line 421: @james-bebbington could you please check if this mapping is correct?

exporter/metric/metric.go
```diff
@@ -562,7 +560,7 @@ func lastValueToTypedValueAndTimestamp(lv *lastvalue.Aggregator, kind apimetric.
 	return tv, t, nil
 }
 
-func sumToTypedValueAndTimestamp(sum *sum.Aggregator, kind apimetric.NumberKind, start, end int64) (*monitoringpb.TypedValue, *monitoringpb.TimeInterval, error) {
+func sumToTypedValueAndTimestamp(sum *array.Aggregator, kind apimetric.NumberKind, start, end int64) (*monitoringpb.TypedValue, *monitoringpb.TimeInterval, error) {
```
nilebox, line 563: the function name should also change to `arrayToTypedValueAndTimestamp` probably?

nilebox (approved):

exporter/metric/metric.go
```diff
@@ -499,7 +497,7 @@ func (me *metricExporter) recordToTypedValueAndTimestamp(r *export.Record) (*mon
 	case apimetric.CounterKind, apimetric.UpDownCounterKind, apimetric.SumObserverKind, apimetric.UpDownSumObserverKind:
 		// CUMULATIVE measurement should have the same start time and increasing end time.
 		// c.f. https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#MetricKind
-		if sum, ok := agg.(*sum.Aggregator); ok {
+		if sum, ok := agg.(*array.Aggregator); ok {
```
nilebox, line 500: @liiling could you try adding a unit test reproducing this issue for both existing metric types (Counter) and new (SumObserver) to make sure we didn't break anything and fixed the issue?

exporter/metric/metric.go
james-bebbington: We do still need to figure out how to handle `UpDownCounters` and `UpDownObservers` as I don't believe these will show up correctly in Cloud Monitoring as `Cumulatives` (which are assumed to always go up).

I think it's okay to merge this for now as `SumObservers` should definitely be treated as `Cunmulatives` and that is the primary change in this PR. Could you please add a TODO to the code to stating that we need to confirm how UpDownCounters should be handled.

exporter/metric/metric.go
james-bebbington: Also, I don't think this change is correct. Looking at https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/metric/selector/simple/simple.go it appears that the default aggregation for these types is still always `sum`.

The default aggregation for Value Observers can be configured, so we will need to rethink how we handle that in the future, but that is covered by the existing TODO comment

liiling: Thanks for the review! I agree that the change seems strange... Will add test cases to reproduce the issue as suggested!

james-bebbington: Sorry for the delayed response.

Yea that's interesting. I hadn't noticed the custom selector we are using in this repo sorry. I'm not sure I fully understand the justification in the comments there as to why we are using array: "array gives the option to use Sum, Count, Max, Min, Quantile and Points (most flexible)". This doesn't seem to apply given that we always extract the sum here anyway.

I would probably recommend dropping that custom selector as I think we're better to follow what the SDK is doing and wait until a "views" or "default aggregations" implementation is implemented there before deciding what to do in this repo. If we drop that custom selector, we should be able to just change this code back to `sum` as per what the SDK's `simple.go` selector does.

In either case, at some point I think we should change this to switch on the aggregation type instead of the metric type, as per https://github.com/open-telemetry/opentelemetry-go/blob/master/exporters/metric/prometheus/prometheus.go#L223

@ymotongpoo what are your thoughts on this?

ymotongpoo: Thank you for raising this to the discussion. The reason why I used `array` instead of `sum` is exactly because of what @james-bebbington pointed out. At the moment I implemented the selector.go, the predefined selectors didn't provide the option to fetch LastValue as noted in the comment here.
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/84f7c391be/exporter/metric/metric.go#L487-L489

I haven't checked the latest status of predefined selectors in SDK, but if they support the LastValue now, then we can remove the custom selector entirely and use the standard one instead. Otherwise, as James says, Views API or default aggregations for LastValue would be the alternative.

james-bebbington: Oh I see what you mean - looks like we do still need this selector for now to select LastValue 👍  (doesn't look like there is a selector in this SDK for do this)

Lets just change the default aggregator in `selector.go` to sum for now then as @liiling suggested

liiling: @james-bebbington Thanks for the update! The suggested changes are now implemented.

james-bebbington (approved):
LGTM

exporter/metric/metric.go
nilebox: FYI I brought up the issue with all SDK selectors using histogram for `ValueObserver`, and the agreement is that we may upstream a new selector which will use last value instead.
Then we'll be able to remove custom selector from our repo.

There is also a similar issue about Prometheus using histogram for exporting gauges: https://github.com/open-telemetry/opentelemetry-go/issues/1139#issuecomment-690750369

Issue #95, opened 2020-09-09
Cloud Monitoring metrics example in README doesn't work


DazWilkin:
The sample as-is appears to not work:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/master/exporter/metric/README.md

+ See #94 
+ There is no (longer) a `kv` package (now `label`?)
+ `meter := pusher.Provider().Meter("...")` vs. `global.Meter("...")` as used by other [examples](https://github.com/open-telemetry/opentelemetry-go/tree/master/example)

Recommend: ensuring this sample is part of the package's sources so that it can be tested when the exporter is built.

nilebox:
Thanks for reporting the issue @DazWilkin 
Feel free to send a PR fixing it if you have time for that.

> Recommend: ensuring this sample is part of the package's sources so that it can be tested when the exporter is built.

We do have a full code example actually: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/master/example/metric
Does it work for you?

DazWilkin:
Thanks @nilebox I'll try that tomorrow

DazWilkin:
@nilebox thanks for the prompt response.

Your example works partially for me.

Cloud Monitoring receives the observer's values but not the counter's.

```console
go run github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric
go: finding module for package github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric
go: downloading github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric v0.11.0
go: found github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric in github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric v0.11.0
2020/09/10 08:42:20 Most recent data: counter 141, observer 12.89
2020/09/10 08:42:30 Most recent data: counter 182, observer 16.89
2020/09/10 08:42:40 Most recent data: counter 136, observer 16.84
2020/09/10 08:42:50 Most recent data: counter 165, observer 15.84
2020/09/10 08:43:00 Most recent data: counter 198, observer 13.49
2020/09/10 08:43:10 Most recent data: counter 122, observer 14.49
2020/09/10 08:43:11 Error during exporting TimeSeries: currently the aggregator is not supported: &{{0 0} 1044 [100 122 136 141 165 182 198]}
[...]

PR #96, opened 2020-09-10
Revised Monitoring example

DazWilkin:
The example in the Monitoring README is insufficient.

I believe this is a working example.

I removed the curious explanation of importing modules. In my experience (and editor), I can simply reference the imports in code and `go.mod` is revised automatically (or will be revised on `go run` or `go build`). So, I simplified this and made the imports exhaustive.

I was unable to use `popts == nil` and so have used the example `resource.New(...)` approach here too.

The Google Project ID is required if run off-cloud and so I've added it here. I noticed, after-the-fact, that the Trace README correctly specifies `WithProjectID(os.Getenv("GOOGLE_CLOUD_PROJECT"))` and so, both READMEs are now consistent. However, neither piece of documentation explains the requirement to `export GOOGLE_CLOUD_PROJECT="..."` and likely should.

### DIFF
--- a/exporter/metric/README.md
+++ b/exporter/metric/README.md
@@ -31,14 +31,7 @@ projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
 opts := []mexporter.Option{
     mexporter.WithProjectID(projectID),
 }
-popts:= []push.Option{
-    push.WithResource(
-        resource.New(
-            label.String("author", "google"),
-            label.String("application", "example"),
-        ),
-    ),
-}
+popts:= []push.Option{}
 
 // Create exporter (collector embedded with the exporter).
 pusher, err := mexporter.InstallNewPipeline(opts, popts)
--- a/exporter/metric/README.md
+++ b/exporter/metric/README.md
@@ -16,6 +16,7 @@ Google Cloud Monitoring is a managed service provided by Google Cloud Platform. 
 After you import the metric exporter package, then register the exporter to the application, and start sending metrics. If you are running in a GCP environment, the exporter will automatically authenticate using the environment's service account. If not, you will need to follow the instructions in [Authentication](#Authentication).
 
 ```go
+package main
 
 import (
     "go.opentelemetry.io/otel/api/metric"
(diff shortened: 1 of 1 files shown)


exporter/metric/README.md
```diff
@@ -13,26 +13,35 @@ Google Cloud Monitoring is a managed service provided by Google Cloud Platform.
...
+    "go.opentelemetry.io/otel/label"
+    "go.opentelemetry.io/otel/sdk/metric/controller/push"
+    "go.opentelemetry.io/otel/sdk/resource"
+
+    mexporter "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
 )
 
 // Initialize exporter option.
-opts := []mexporter.Option{}
+projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
```
nilebox, line 30: As I commented in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/95#issuecomment-690782197, this shouldn't be necessary when the application is running in GCP environment.

So I would prefer to omit this in the initial example, and then add a short paragraph describing how to set / override project ID if needed.

exporter/metric/README.md
```diff
@@ -13,26 +13,35 @@ Google Cloud Monitoring is a managed service provided by Google Cloud Platform.
...
+opts := []mexporter.Option{
+    mexporter.WithProjectID(projectID),
+}
+popts:= []push.Option{
+    push.WithResource(
```
nilebox, line 35: Not sure if this is useful for the minimal example. Maybe we should fix the `InstallNewPipeline` to allow null popts instead? Alternatively, just make an empty slice.

exporter/metric/README.md
```diff
@@ -42,16 +51,18 @@ defer pusher.Stop()
 ctx := context.Background()
 meter := pusher.Provider().Meter("cloudmonitoring/example")
 
-counter := metric.Must(meter).NewInt64Counter("counter-foo")
-labels := []kv.KeyValue{kv.Key("key").String("value")}
+counter := metric.Must(meter).NewInt64Counter("counter.foo")
```
nilebox, line 54: Any specific reason for replacing `-` with `.` in the metric name?

Google Cloud Monitoring is actually using `/` and `_` as separators normally, e.g. see metrics in https://cloud.google.com/monitoring/api/metrics_kubernetes

exporter/metric/README.md
DazWilkin: OK. Will revise. Would you like me to make the change to the Trace `README` example too to keep them consistent? if so, I'll submit the Trace change as a separate PR.

exporter/metric/README.md
DazWilkin: Yes, supporting `nil` is best ;-)

If I use `v0.11.0` then an empty slice generates an error:

```
2020/09/11 10:04:55 Error during exporting TimeSeries: currently the aggregator is not supported: &{{0 0} 123 [123]}
2020/09/11 10:04:55 rpc error: code = InvalidArgument desc = Request was missing field timeSeries.
```

If I use `v0.11.1-0.20200910093037-ef5e24b7fcfc` then an empty slice works. Will change to proactively permit.

exporter/metric/README.md
DazWilkin: I'd thought I'd seen `counter.foo` used in another example and so inadvertently copied. Will revert.

DazWilkin: Reverted: 0fee3ade2a5087fdfaf7df92d8577c358e499175

exporter/metric/README.md
DazWilkin: Fixed: 7e67ff1fca11b098d4a7410d207cbf18fd75ac37

exporter/metric/README.md
DazWilkin: Fixed: 2edc48082e9cc22b0d0c2614a612ba8fa2f6f76a

nilebox (approved):

PR #97, opened 2020-09-14
Working example and consistent with Monitoring README

DazWilkin:
Per https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/95#issuecomment-690782197 revision to the Trace example to remove the redundant `WithProjectID(...)` if running on GCP. The function is now explained under "Authentication".

The example referenced `tracer.WithSpan(...)` which appears incorrect; I looked at the GCP [example](https://cloud.google.com/trace/docs/setup/go-ot) and revised this with that example's self-executing function that includes a `tracer.Start(...)`

Feel free to ignore this PR; I offered to ensure both examples are consistent and this does that.

### DIFF
--- a/exporter/trace/README.md
+++ b/exporter/trace/README.md
@@ -15,45 +15,57 @@ Google Cloud Trace is a managed service provided by Google Cloud Platform. The e
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


nilebox (approved):
Looks good, thank you!

Issue #98, opened 2020-09-17
Writing timeseries fails due to invalid argument error


shouichi:
We tried to instrument http handlers using `go.opentelemetry.io/contrib/instrumentation/net/http` and got the following error.

```
2020/09/17 05:36:44 rpc error: code = InvalidArgument desc = Request was missing field timeSeries.
2020/09/17 05:37:44 rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Field timeSeries[0].metric.labels had an invalid value of "http.scheme": Label key contains invalid characters.: timeSeries[0]; Field timeSeries[1].metric.labels had an invalid value of "http.server_name": Label key contains invalid characters.: timeSeries[1]; Field timeSeries[2].metric.labels had an invalid value of "http.host": Label key contains invalid characters.: timeSeries[2]
```

The version we use is 7f2025af52931b2969faf09e998dd0731e8c9266.

nilebox:
> Field timeSeries[0].metric.labels had an invalid value of "http.scheme": Label key contains invalid characters.

Looks like we need to rename the label keys, e.g. replace `.` with `_` in the exporter, since these format limitations are specific to Cloud Monitoring APIs.

nilebox:
As far as I understand, the renaming can be done in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/7f2025af52931b2969faf09e998dd0731e8c9266/exporter/metric/metric.go#L462

@shouichi would you be interested in sending a PR making this change? If you do so, please also make sure to add a unit test illustrating the issue.

shouichi:
Sure, I'll take a look 👍

PR #99, opened 2020-09-18
Normalize label keys

shouichi:
Close https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/98.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -459,7 +459,7 @@ func (me *metricExporter) recordToMpb(r *export.Record) *googlemetricpb.Metric {
 	iter := r.Labels().Iter()
 	for iter.Next() {
 		kv := iter.Label()
-		labels[string(kv.Key)] = kv.Value.AsString()
+		labels[normalizeLabelKey(string(kv.Key))] = kv.Value.AsString()
 	}
 
 	return &googlemetricpb.Metric{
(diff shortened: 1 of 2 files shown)


exporter/metric/metric.go
```diff
@@ -604,3 +604,15 @@ func aggToTypedValue(kind apimetric.NumberKind, value apimetric.Number) (*monito
...
+// https://github.com/googleapis/googleapis/blob/c4c562f89acce603fb189679836712d08c7f8584/google/api/metric.proto#L149
+//
+// > The label key name must follow:
+// >
+// > * Only upper and lower-case letters, digits and underscores (_) are
+// >   allowed.
+// > * Label name must start with a letter or digit.
+// > * The maximum length of a label name is 100 characters.
+func normalizeLabelKey(s string) string {
+	return strings.ReplaceAll(s, ".", "_")
```
nilebox, line 617: This looks good as the first step, though Ideally we would also need to make sure that the label key is valid after this conversion (with all the other limitations you listed above), and might be useful to return an error if it's still invalid after "normalization".

nilebox (approved):

exporter/metric/metric.go
shouichi: Yes. We try to adopt this exporter to our production and we might send another PR(s) in the future. Thanks for the quick merge 👍

shouichi:
Tried 45880ec6f71fcaf0b3830da8361be24a715646d8 and confirmed that writing timeseries worked 🎉

exporter/metric/metric.go
martinraison: This is still relevant today. Especially with third-party library instrumentation, it's not always trivial to guarantee all metric names follow Google's conventions. For now I'm replacing all characters outside `[A-Za-z0-9]` with `_` and truncating to 100 characters

PR #101, opened 2020-10-01
Update otel to v0.12.0

yas-okadatech:


bogdandrutu:
A merge and release of this PR would be really appreciated by the opentelemetry-collector-contrib repo

mrothroc:
It would be great to have this.  I'll just note that you can't use go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
to trace gRPC calls in google cloud without this PR because that module requires 0.12.0

exporter/trace/cloudtrace.go
```diff
@@ -233,26 +233,23 @@ type Exporter struct {
...
 	exporter, err := NewExporter(opts...)
 	if err != nil {
 		return nil, nil, err
 	}
-	tp, err := sdktrace.NewProvider(append(topts, sdktrace.WithSyncer(exporter))...)
-	if err != nil {
-		return nil, nil, err
-	}
+	tp := sdktrace.NewTracerProvider(append(topts, sdktrace.WithSyncer(exporter))...)
 	return tp, exporter.traceExporter.Flush, nil
```
james-bebbington, line 253: Would this be somewhat cleaner if we returned `exporter.Shutdown` here instead or `exporter.traceExporter.Flush`?

james-bebbington (approved):
LGTM other than one minor comment.

Since this PR is blocking people I will merge it now, but can you please check the comment below and fix in a new PR if correct?

PR #105, opened 2020-10-06
Fix to use the correct version to build user agent

shouichi:


exporter/metric/option.go
```diff
@@ -25,9 +25,7 @@ import (
 	apioption "google.golang.org/api/option"
 )
 
-var (
-	userAgent = fmt.Sprintf("opentelemetry-go %s; metric-exporter %s", otel.Version(), version)
-)
+var userAgent = fmt.Sprintf("opentelemetry-go %s; metric-exporter %s", otel.Version(), Version())
```
nilebox, line 28: Unrelated to version fix, but this should probably have `google-cloud-metric-exporter` or similar instead of just `metric-exporter` to match the `google-cloud-trace-exporter` string used for tracing (see below).

@james-bebbington @aabmass do we have this documented anywhere to make sure we use consistent agent for other languages as well?

nilebox (approved):

shouichi:
@nilebox Totally unrelated but... Could you add `hacktoberfest-accepted` label to my PRs or add `hacktoberfest` topic to this repository so that they are accepted to hacktoberfest? Thanks!

exporter/metric/option.go
james-bebbington: The trace user agent is document but the metric one is not. @aabmass - can you add that to the requirements doc?

Agree that `google-cloud-metric-exporter` would make sense here

james-bebbington (commented):
We should really have test cases to cover this:

- For traces it should be relatively easy to add a check to `cloudtrace_test.go` to validate the user agent _attribute_ (`g.co/agent`) is set correctly.
- For metrics it doesn't look like we have tests using the mocks yet, so we can leave this for now. But can you change the text to `google-cloud-metric-exporter` as commented above?

james-bebbington:
I'm going to merge this for now since I want to do a release and we should include this. Have created https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/112 to track adding tests.

Issue #107, opened 2020-10-19
Frequent Internal Errors when exporting metrics with different labels at runtime


IsaacPD:
Frequently when attempting to collect metrics I encounter this error

> 2020/10/19 15:37:59 rpc error: code = Internal desc = Internal error encountered. Please retry after a few seconds. If internal errors persist, contact support at https://cloud.google.com/support/docs.

and then in metric explorer the metrics are temporarily unavailable until the next time I run my program and don't see this error.

I believe this issue is similar to https://github.com/googleapis/python-monitoring/issues/7 as the issue seems to be related to whenever I try to write time series that have different labels one run after the other, and metric descriptor creation requests are always sent despite them existing already.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ec19c7e5b68f792f5f6e6c6e652a4b033726d407/exporter/metric/metric.go#L178-L187

This comment https://github.com/googleapis/python-monitoring/issues/7#issuecomment-620365463 mentions that appropriate metric descriptors are created on creation of time series, and that it is not necessary to create metric descriptors with a separate call. The fix to the issue I linked has a different fix in which they check whether the metric descriptor exists first and if it does not then they create it.

I've tried not making the call to ```exportMetricDescriptors``` and that worked for me, so I believe that is the issue.

Steps to reproduce:
Change example/metric/example.go to use different labels at runtime depending on os.Args as such
```go
	of := newObservedFloat(12.34)
	i, _ := strconv.ParseUint(os.Args[1], 10, 32)
	callback := func(_ context.Context, result metric.Float64ObserverResult) {
		v := of.get()
		result.Observe(v, olabels[i])
	}

	metric.Must(meter).NewFloat64ValueObserver("observer-a", callback)

	// Add measurement once an every 10 second.
	timer := time.NewTicker(10 * time.Second)
	for range timer.C {
		rand.Seed(time.Now().UnixNano())

		r := rand.Int63n(100)
		cv := 100 + r
		counter.Add(ctx, cv, clabels[i])

		r2 := rand.Int63n(100)
		ov := 12.34 + float64(r2)/20.0
		of.set(ov)
		log.Printf("Most recent data: counter %v, observer %v", cv, ov)
	}
```
Run ```go run example.go 1 && go run example.go 0```
An error will appear after a couple of runs if not on the first

dashpole:
[...]

PR #108, opened 2020-10-22
Check for the existence of metric descriptors before creating

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/107

The recommended use of the stackdriver metrics API for a given metric is:

1. CreateMetricDescriptor 
2. CreateTimeSeries
3. CreateTimeSeries
4. CreateTimeSeries
5. CreateTimeSeries
...

However, the OpenTelemetry APIs don't have any sort of "Initialization" functions for metrics, only `Export`.  Each time Export is called, we must assume it could be the first time we have seen that metric, and attempt to create a metric descriptor first.  We are currently doing:

1. CreateMetricDescriptor 
2. CreateTimeSeries
3. CreateMetricDescriptor 
4. CreateTimeSeries
5. CreateMetricDescriptor 
6. CreateTimeSeries
7. CreateMetricDescriptor 
8. CreateTimeSeries

However, that seems to cause problems for stackdriver, as shown in the linked issue.

Instead, we can first check to see if the metric descriptor already exists:

1. CreateMetricDescriptor 
2. CreateTimeSeries
3. GetMetricDescriptor 
4. CreateTimeSeries
5. GetMetricDescriptor 
6. CreateTimeSeries

This doesn't appear to cause the same problems that CreateMetricDescriptor + CreateTimeSeries does.

Note to reviewers:  I wasn't able to unit test this specific behavior because the mock stackdriver service doesn't return previously created metric descriptors in GetMetricDescriptor calls.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -216,17 +216,33 @@ func (me *metricExporter) exportMetricDescriptor(ctx context.Context, cps export
 	// the descriptor does not exist in global cache (me.mdCache).
 	// See details in #26.
 	for kmd, md := range mds {
-		req := &monitoringpb.CreateMetricDescriptorRequest{
-			Name:             fmt.Sprintf("projects/%s", me.o.ProjectID),
-			MetricDescriptor: md,
-		}
-		_, err := me.client.CreateMetricDescriptor(ctx, req)
+		err := me.createMetricDescriptorIfNeeded(ctx, md)
 		if err != nil {
 			return err
 		}
 		me.mdCache[kmd] = md
 	}
 	return nil
+}
+
+func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md *googlemetricpb.MetricDescriptor) error {
+	mdReq := &monitoringpb.GetMetricDescriptorRequest{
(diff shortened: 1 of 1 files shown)


exporter/metric/metric.go
```diff
@@ -216,11 +216,22 @@ func (me *metricExporter) exportMetricDescriptor(ctx context.Context, cps export
 	// the descriptor does not exist in global cache (me.mdCache).
 	// See details in #26.
 	for kmd, md := range mds {
+		mdReq := &monitoringpb.GetMetricDescriptorRequest{
```
nilebox, line 219: Could you extract this code into a separate `getOrCreateMetricDescriptor` function to make it easier to maintain and test?

Also, do we need to cover this change with a test, or is it already covered by a mock E2E test?

dashpole: Updated.  It will get _and_ create the metric descriptor, so I named it createMetricExporterIfNeeded

It is technically covered by the mock E2E test, but the mock doesn't seem to do anything with the requests.

nilebox (approved):

PR #111, opened 2020-10-28
Better bundle config and error management

james-bebbington:
Changes:
- Allow BundleByteThreshold & BundleByteLimit to be configured
- Set default BundleByteLimit to unlimited (was 50KB, or 1KB * bundle count threshold)
- Log oversized item errors separately to buffer full errors

Tests:
- Change the cloudtrace_test file so it runs in the trace package
- Add a test for oversized errors

### DIFF
--- a/exporter/trace/trace.go
+++ b/exporter/trace/trace.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"sync"
 	"time"
 
@@ -41,11 +42,11 @@ type traceExporter struct {
 	client *traceclient.Client
 }
 
-const defaultBufferedByteLimit = 8 * 1024 * 1024
 const defaultBundleDelayThreshold = 2 * time.Second
 const defaultBundleCountThreshold = 50
-const bundleByteThresholdMultiplier = 300
-const bundleByteLimitMultiplier = 1000
+const defaultBundleByteThreshold = 15000
+const defaultBundleByteLimit = 0
+const defaultBufferedByteLimit = 8 * 1024 * 1024
 
 func newTraceExporter(o *options) (*traceExporter, error) {
 	client, err := traceclient.NewClient(o.Context, o.TraceClientOptions...)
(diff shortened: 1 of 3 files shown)


exporter/trace/trace.go
```diff
@@ -144,38 +159,55 @@ func (e *traceExporter) Flush() {
...
+	var bufferFullMsg, oversizedItemMsg, separator string
+
+	if bufferFull > 0 {
+		bufferFullMsg = fmt.Sprintf("buffer full: %v", bufferFull)
+	}
+	if oversized > 0 {
+		oversizedItemMsg = fmt.Sprintf("oversized item: %v", oversized)
 	}
+	if bufferFull > 0 && oversized > 0 {
+		separator = ", "
```
nilebox, line 209: We could just use https://golang.org/pkg/strings/#Join instead

exporter/trace/trace.go
```diff
@@ -41,42 +41,57 @@ type traceExporter struct {
...
+
 	if o.BundleCountThreshold > 0 {
 		b.BundleCountThreshold = o.BundleCountThreshold
 	} else {
 		b.BundleCountThreshold = defaultBundleCountThreshold
 	}
-	b.BundleByteThreshold = b.BundleCountThreshold * bundleByteThresholdMultiplier
-	b.BundleByteLimit = b.BundleCountThreshold * bundleByteLimitMultiplier
+
+	if o.BundleByteThreshold > 0 {
```
dashpole, line 77: Doesn't block this PR, but should we return errors for invalid configuration instead of silently using the default?

dashpole (approved):

exporter/trace/trace.go
jsuereth: We should chat about that.  I think for something like it we should do our best to ignore config errors and limp along not making your app worse.

However, this would be an undiscoverable config error, which should be fixed in some fashion...

jsuereth (approved):

exporter/trace/trace.go
nilebox: We could just pass these parameters as is to `bundler.NewBundler` and let it handle zero values and other validation.

exporter/trace/trace.go
```diff
@@ -41,42 +41,57 @@ type traceExporter struct {
 	client *traceclient.Client
 }
 
-const defaultBufferedByteLimit = 8 * 1024 * 1024
 const defaultBundleDelayThreshold = 2 * time.Second
 const defaultBundleCountThreshold = 50
-const bundleByteThresholdMultiplier = 300
```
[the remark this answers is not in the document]
james-bebbington: I'm not sure. I believe Steven did some benchmarking to determine what values worked best, so I am keen to leave these defaults that are probably better tuned for our use case than the bundler library defaults.

I don't know he used multipliers for bundleByteThreshold / bundleByteLimit though. I just changed these to be more explicit (and removed the bundle byte limit).

exporter/trace/trace.go
james-bebbington: Doesn't look the bundler library does any validation of its own. I think in all their use cases, negative values will behave the same as zero which is arguably an idiomatic way to handle things, e.g. They use `DelayThreshold` in `time.AfterFunc` which ends up hitting this condition in the Go source: https://github.com/golang/go/blob/5f3354d1bf2e6a61e4b9e1e31ee04b99dfe7de35/src/time/sleep.go#L29

Still, I'd be happy to make this return errors instead but would do in a follow up PR

james-bebbington: Or maybe that makes more sense to be something that is done by the code that reads config, e.g. in the Collector

jsuereth: Sorry for not responding sooner. I think the important part here is from a user perspective, there is a correctable notification of failures and values aren't silently converted in a way that's confusing.

Likely that means work in whatever reads the config and passes to this function.

PR #119, opened 2020-11-19
Map opentelemetry service.name to Service in SD UI

dashpole:
I'm not sure if we should rely on the `g.co/gae/app/module` label to set "Service" in the UI, but it does make traces more useable...

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -47,12 +47,16 @@ const (
 	URLAttribute        = "http.url"
 	UserAgentAttribute  = "http.user_agent"
 	StatusCodeAttribute = "http.status_code"
+	ServiceAttribute    = "service.name"
 
 	labelHTTPHost       = `/http/host`
 	labelHTTPMethod     = `/http/method`
 	labelHTTPStatusCode = `/http/status_code`
 	labelHTTPPath       = `/http/path`
 	labelHTTPUserAgent  = `/http/user_agent`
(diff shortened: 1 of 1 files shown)


exporter/trace/trace_proto.go
```diff
@@ -47,12 +47,14 @@ const (
...
 	UserAgentAttribute  = "http.user_agent"
 	StatusCodeAttribute = "http.status_code"
+	ServiceAttribute    = "service.name"
 
 	labelHTTPHost       = `/http/host`
 	labelHTTPMethod     = `/http/method`
 	labelHTTPStatusCode = `/http/status_code`
 	labelHTTPPath       = `/http/path`
 	labelHTTPUserAgent  = `/http/user_agent`
+	labelService        = `g.co/gae/app/module`
```
nilebox, line 57: Not sure if this is a good idea to support this label across all resources - this seems specific to AppEngine?

nilebox: Actually, the mapping seems reverse (i.e. when this label is present, it will be translated to `service.name` on Cloud Monitoring side), which is okay probably.

nilebox (approved):

punya:
@dashpole is this ready to merge?

exporter/trace/trace_proto.go
punya: Seems worth adding a comment so that the next person reading this code doesn't have the same question.

dashpole: Comment Added

PR #120, opened 2020-11-23
Update to otel 0.14

adlerhurst:
Hi guys

As i was testing otel 0.14 i updated your project. Maybe the changes work with you.

Greets from Switzerland


exporter/metric/metric.go
```diff
@@ -21,8 +21,9 @@ import (
 	"strings"
 	"time"
 
-	"go.opentelemetry.io/otel/api/global"
-	apimetric "go.opentelemetry.io/otel/api/metric"
+	global "go.opentelemetry.io/otel"
```
dashpole, line 24: No need to rename.  Lets stick with otel.

exporter/trace/trace_proto.go
```diff
@@ -27,8 +27,8 @@ import (
 	tracepb "google.golang.org/genproto/googleapis/devtools/cloudtrace/v2"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
 
+	opentelemetry "go.opentelemetry.io/otel"
```
dashpole, line 30: can we change to use "otel"?

exporter/metric/error.go
```diff
@@ -20,7 +20,8 @@ import (
 
 	"go.opentelemetry.io/otel/sdk/export/metric/aggregation"
 
-	apimetric "go.opentelemetry.io/otel/api/metric"
+	apimetric "go.opentelemetry.io/otel/metric"
```
dashpole, line 23: to be consistent, can we switch to using `metric` instead of `apimetric`?

exporter/metric/metric.go
```diff
@@ -21,8 +21,9 @@ import (
...
+	apimetric "go.opentelemetry.io/otel/metric"
```
dashpole, line 25: It would also be nice to use `metric`, rather than `apimetric` here.

dashpole (commented):
Thanks for the help!  Just a few nits.

exporter/metric/metric.go
```diff
@@ -21,8 +21,10 @@ import (
 	"strings"
 	"time"
 
-	global "go.opentelemetry.io/otel"
-	apimetric "go.opentelemetry.io/otel/metric"
+	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
```
dashpole, line 24: please keep non-opentelemetry imports grouped at the bottom.  This also includes the protobuf import.

adlerhurst:
hi @dashpole 

thanks for your review. i've corrected the files.

I've also changed the used otel version in.
https://github.com/adlerhurst/opentelemetry-operations-go/blob/20c83ede56cac1b365904bda780e79bba9c9cf46/exporter/trace/version.go#L20
and
https://github.com/adlerhurst/opentelemetry-operations-go/blob/20c83ede56cac1b365904bda780e79bba9c9cf46/exporter/metric/version.go#L20

exporter/metric/metric.go
adlerhurst: changed it in all files

dashpole (approved):

dashpole:
closed/reopened to poke tests

PR #123, opened 2020-12-22
Added a set function to set MonitoringClientOptions in option

Kun17:
Currently there is no need to set MonitoringClientOptions in trace exporter option.

If user wants to export to stackdriver monitoring, there is no way to pass credentials. 

So here we are!


Kun17:
> @googlebot I signed it!

@googlebot I signed it!

PR #129, opened 2021-01-12
Update otel dependency to v0.15.0

dashpole:


aabmass (approved):
LGTM.

On a side note, should we mark the `*.sum` files as generated to hide the noise by default https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-how-changed-files-appear-on-github

I'm not too familiar with Go tooling though so feel free to ignore

dashpole:
I usually still skim go.sum to make sure it isn't pulling in any dependencies I don't want to have.

PR #135, opened 2021-01-28
Fix lint error about int conversion

punya:
Previous version of code would have produced a rune rather
than a decimal representation.


aabmass (approved):
🚢

PR #137, opened 2021-01-28
Stagger interval end time when exporting

### LINKED ISSUE #130: Invalid start time on a time series data point
I wrote the following script to prototype writing and exporting metrics with OpenTelemetry Go. However, I get errors complaining about an invalid start time every time I try to run:

```
func main() {
	// Initialize exporter options
	formatter := func(d *metric.Descriptor) string {
		return fmt.Sprintf("MY.METRIC.DOMAIN/%s", d.InstrumentationName())
	}
 
	keyOpt := apioption.WithCredentialsFile(os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"))
	endpointOpt := apioption.WithEndpoint("MY.MONITORING.ENDPOINT:PORT")
 
	opts := []mexporter.Option{
		mexporter.WithProjectID("MY-PROJECT-ID"),
		mexporter.WithMetricDescriptorTypeFormatter(formatter),
		mexporter.WithMonitoringClientOptions(keyOpt, endpointOpt),
	}
	popts:= []push.Option{}
 
	// Create Exporter
	pusher, err := mexporter.InstallNewPipeline(opts, popts...)
	if err != nil {
		log.Fatalf("mexporter.InstallNewPipeline: %v", err)
	}
	defer pusher.Stop()
 
	// Start meter
	ctx := context.Background()
	meter := pusher.MeterProvider().Meter("MY/CUSTOM/METRIC")
 
	counter := metric.Must(meter).NewInt64Counter("response_count")
	labels := []label.KeyValue{
		label.Key.String("response_code", "200"),
		label.Key.String("response_code_class", "2XX"),
	}
	counter.Add(ctx, 1, labels...)
}
```

```
InvalidArgument desc = Field timeSeries[0].points[0].interval.start_time had an invalid value of "2021-01-14T10:49:57-08:00": The start time must be before the end time (2021-01-14T10:49:57-08:00) for the non-gauge metric '[...]'
```

As per my offline conversation with Nail, this seems to be a result of the start and end time being equal, I guess as a result of being registered at the same time. There's a provision in the [Java libraries](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/97ef862a34ee84e2603600fad80ca4ca7516b680/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java#L134-L142) to ensure they are staggered with endTime as +1 to startTime, but it seems like no such provision is included in the go libraries.

punya:
* Use canonical utilities to convert to timestamp PBs, which also improves precision (seconds to millis).
* Meet [GCP requirement](https://cloud.google.com/monitoring/api/ref_v3/rpc/google.monitoring.v3#timeinterval) that the end time of a new interval must be at least a millisecond after the end time of the
previous interval, for all non-gauge types.

Fixes #130

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -34,6 +34,7 @@ import (
 	"go.opentelemetry.io/otel/sdk/resource"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
+	"github.com/golang/protobuf/ptypes"
 	googlepb "github.com/golang/protobuf/ptypes/timestamp"
 	"google.golang.org/api/option"
 	googlemetricpb "google.golang.org/genproto/googleapis/api/metric"
@@ -507,7 +508,6 @@ func (me *metricExporter) recordToTypedValueAndTimestamp(r *export.Record) (*mon
 	agg := r.Aggregation()
 	nkind := r.Descriptor().NumberKind()
 	ikind := r.Descriptor().InstrumentKind()
-	now := time.Now().Unix()
 
 	// TODO: Ignoring the case for Min, Max and Distribution to simply
 	// the first implementation.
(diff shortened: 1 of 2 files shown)


dashpole (commented):
Great find!  It would be nice to have a unit test for this if possible

dashpole (approved):

PR #140, opened 2021-02-03
Fix verify-examples

punya:
* Add missing tools
* Remove GOPATH usage


verify_examples.sh
```diff
@@ -19,16 +19,6 @@ set -euo pipefail
 cd $(dirname $0)
 TOOLS_DIR=$(pwd)/.tools
 
-if [ -z "${GOPATH}" ] ; then
```
aabmass, line 22: `$GOPATH` isn't needed?

aabmass (approved):

verify_examples.sh
punya: We use Go modules, so no.

PR #141, opened 2021-02-05
Add k8s_container metric example

dashpole:
This writes to metric: `kubernetes.io/container/ephemeral_storage/used_bytes`


dashpole:
Waiting for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/142

example/container_metric/example.go
```diff
@@ -0,0 +1,101 @@
...
+	// Recorder metric ephemeral storage metric
+	vr, err := pusher.MeterProvider().Meter("cloudmonitoring/example").
+		NewInt64ValueRecorder("container/ephemeral_storage/used_bytes")
+	if err != nil {
+		log.Fatalf("failed to establish pipeline: %v", err)
+	}
+	valuerecorder := vr.Bind()
+	defer valuerecorder.Unbind()
+
+	// Add measurement once an every 10 second.
```
punya, line 92:
```suggestion
	// Add measurement once every 10 seconds.
```

example/container_metric/example.go
```diff
@@ -0,0 +1,101 @@
...
+	of.f = v
+}
+
+func (of *observedInt) get() int64 {
+	of.mu.RLock()
+	defer of.mu.RUnlock()
+	return of.f
+}
+
+func newObservedInt(v int64) *observedInt {
```
punya, line 50: Is this struct/func used anywhere?

punya (commented):
The code looks reasonable. Is it worth adding a bit of documentation (either in the README or as a comment in example.go) saying why this example is valuable? E.g., is this individual metric something that people care about; is it because it teaches you how to follow GCM semantic conventions regarding K8s; or something else altogether?

Is this example something that's worth including for other languages too?

example/container_metric/example.go
```diff
@@ -0,0 +1,101 @@
...
+	// prepare credential file following the instruction described in this doc.
+	// https://pkg.go.dev/golang.org/x/oauth2/google?tab=doc#FindDefaultCredentials
+	opts := []mexporter.Option{
+		mexporter.WithMetricDescriptorTypeFormatter(func(desc *metric.Descriptor) string {
+			return fmt.Sprintf("kubernetes.io/%s", desc.Name())
+		}),
+		mexporter.WithInterval(10 * time.Second),
+	}
+
+	// These are the semantic conventions for containers that map to the k8s_container resource
```
nilebox, line 67: Wouldn't it make more sense to use https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/gke.go here as an example closer to what we recommend to our users?

dashpole: we could potentially override all of the env vars for GKE, but the GCE ones won't be set unless the example is run inside GCE.  My hope was to have an example I could reference when I needed to test GKE metrics locally

example/container_metric/example.go
dashpole: removed

dashpole:
added a readme

punya (approved):

example/container_metric/example.go
nilebox: > but the GCE ones won't be set unless the example is run inside GCE.

GKE nodes are running on GCE VMs (well, for "GKE on GCP" specifically, not on-prem or AWS), so I thought they would be able to reach GCP metadata server and fetch GCE metadata from there?

The purpose of an example is usually for the end user to copy-paste and modify it to their needs, so my fear here is that instead of using auto-detection mechanism they will get the idea that they have to manually provide all the attributes every time.

Maybe consider adding a comment or adding some flag to the application to run in the "local" mode vs "production" mode on GCE/GKE?

punya:
@dashpole when can we merge this?

example/container_metric/example.go
dashpole: Yeah, I think it might be better just to keep this example for myself to reference, and keep existing examples for users.

PR #142, opened 2021-02-05
allow Google Inc copyright holder

dashpole:
See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/141.  It doesn't doesn't seem to let me add Google Inc in the same PR as a new file, so adding a separate change here.


aabmass:
Should the copyright in #141 just be "Google"? I also see "Google LLC" in some guidelines elsewhere as well

dashpole:
Ooh, Google LLC passes!

PR #148, opened 2021-02-27
Fix user agent strings

### LINKED ISSUE #112: Add tests that the user agent string is set correctly for traces & metrics
See comment: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/105#pullrequestreview-502680035

jsuereth:
- [X] Ensure user agent string is correct in Trace Exporter
- [X] Add user agent string test for trace exporter
- [x] Add user agent string test for metrics exporter.

Fixes #112

### DIFF
--- a/exporter/metric/metric_test.go
+++ b/exporter/metric/metric_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020, Google Inc.
+// Copyright 2020-2021, Google Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/exporter/metric/metric_test.go
+++ b/exporter/metric/metric_test.go
@@ -17,6 +17,7 @@ package metric
 import (
 	"context"
 	"fmt"
+	"net"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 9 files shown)


exporter/metric/metric_test.go
```diff
@@ -422,3 +429,76 @@ func TestTimeIntervalPassthru(t *testing.T) {
...
+
+	exporter, err := NewRawExporter(opts...)
+	if err != nil {
+		t.Errorf("Error occurred when creating exporter: %v", err)
+	}
+
+	err = exporter.Export(ctx, cps)
+	assert.NoError(t, err)
+
+	// Now check for user agent string in the buffer from one call.
```
punya, line 501: Is this missing a check for the other call?

exporter/metric/metric_test.go
```diff
@@ -422,3 +429,76 @@ func TestTimeIntervalPassthru(t *testing.T) {
...
+func (m *mock) CreateMetricDescriptor(ctx context.Context, req *monitoringpb.CreateMetricDescriptorRequest) (*googlemetricpb.MetricDescriptor, error) {
+	return m.createMetricDescriptor(ctx, req)
+}
+
+func TestExportMetricsWithUserAgent(t *testing.T) {
+	server := grpc.NewServer()
+	t.Cleanup(server.Stop)
+
+	// Channel to shove user agent strings into
+	ch := make(chan []string, 2)
```
punya, line 452: I assume this is allocated with length 2 so that you can save two calls into it, then check both. Consider interleaving the calls and checks instead -- it's a bit clearer to see that *each* call sets the UA properly. If you go that route, you can reduce the buffer to length 1.

punya (approved):

exporter/metric/metric_test.go
jsuereth: Do you think there's value in checking both?   What're the odds you misconfigure JUST ONE grpc method user agent without misconfiguring both?

exporter/metric/metric_test.go
jsuereth: I updated it to be as you state, I think the test is a bit less readable and I was a bit concerned about whether it flakily passes as I'm not sure how to know when a channel is "done" in Go without adding an explicit signal. I don't think there's a huge value in testing both UAs, but it's done :)

PR #149, opened 2021-03-03
Remove bundler in favor of BatchSpanProcessor + to comply w/ Specification

### LINKED ISSUE #100: Check if we need to make changes to support ForceFlush in batch processor
See recent change https://github.com/open-telemetry/opentelemetry-go/pull/1166 (issue: https://github.com/open-telemetry/opentelemetry-go/issues/1072)

jsuereth:
Fixes #100 

- Remove bundler and associated config/tests
- Remove metrics-options from trace-exporter config
- Update returned `flush` method from "pipeline" generation for `shutdown` method (only exposed sdk method in 1.0 spec)
- Add configuration for batch processor to Exporter-Pipeline helper method
- Update examples to denote `flush`=>`shutdown` semantic change

### DIFF
--- a/exporter/trace/cloudtrace.go
+++ b/exporter/trace/cloudtrace.go
@@ -67,41 +67,10 @@ type options struct {
 	// Optional.
 	OnError func(err error)
 
-	// MonitoringClientOptions are additional options to be passed
-	// to the underlying Stackdriver Monitoring API client.
-	// Optional.
-	MonitoringClientOptions []option.ClientOption
-
 	// TraceClientOptions are additional options to be passed
 	// to the underlying Stackdriver Trace API client.
 	// Optional.
 	TraceClientOptions []option.ClientOption
-
-	// BundleDelayThreshold determines the max amount of time
-	// the exporter can wait before uploading view data or trace spans to
-	// the backend.
-	// Optional. Default value is 2 seconds.
-	BundleDelayThreshold time.Duration
-
-	// BundleCountThreshold determines how many view data events or trace spans
-	// can be buffered before batch uploading them to the backend.
-	// Optional. Default value is 50.
-	BundleCountThreshold int
-
-	// BundleByteThreshold is the number of bytes that can be buffered before
-	// batch uploading them to the backend.
-	// Optional. Default value is 15KB.
-	BundleByteThreshold int
-
-	// BundleByteLimit is the maximum size of a bundle, in bytes. Zero means unlimited.
-	// Optional. Default value is unlimited.
-	BundleByteLimit int
-
-	// BufferMaxBytes is the maximum size (in bytes) of spans that
-	// will be buffered in memory before being dropped.
-	//
-	// If unset, a default of 8MB will be used.
-	BufferMaxBytes int
 
 	// DefaultTraceAttributes will be appended to every span that is exported to
 	// Stackdriver Trace.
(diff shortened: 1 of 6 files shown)


punya (approved):

exporter/trace/cloudtrace.go
```diff
@@ -67,41 +67,14 @@ type options struct {
 	// Optional.
 	OnError func(err error)
 
-	// MonitoringClientOptions are additional options to be passed
-	// to the underlying Stackdriver Monitoring API client.
-	// Optional.
-	MonitoringClientOptions []option.ClientOption
```
dashpole, line 73: Should this just be renamed to TraceClientOptions instead of removed?  The actual option.ClientOption doesn't appear to be specific to GCM

exporter/trace/cloudtrace.go
```diff
@@ -267,34 +171,47 @@ type Exporter struct {
...
 func NewExporter(opts ...Option) (*Exporter, error) {
 	o := options{Context: context.Background()}
 	for _, opt := range opts {
 		opt(&o)
 	}
+	return NewExporterWithOptions(&o)
+}
+
+// NewExporter creates a new Exporter thats implements trace.Exporter.
+// This version takes a single options struct.  This shouldn't be called
```
dashpole, line 212: If this shouldn't be called by user code, should this function be package-local (private)?

exporter/trace/trace.go
```diff
@@ -62,72 +61,31 @@ func newTraceExporter(o *options) (*traceExporter, error) {
...
-	protoSpan := protoFromSpanSnapshot(sd, e.projectID, e.o.DisplayNameFormatter)
-	protoSize := proto.Size(protoSpan)
-	err := e.bundler.Add(protoSpan, protoSize)
-	e.checkBundlerError(err)
+func (e *traceExporter) ConvertSpan(_ context.Context, sd *export.SpanSnapshot) *tracepb.Span {
+	return protoFromSpanSnapshot(sd, e.projectID, e.o.DisplayNameFormatter)
+}
+
+func (e *traceExporter) Shutdown(ctx context.Context) error {
+	// return e.client.Close()
```
dashpole, line 83: nit: stray, commented out code.

exporter/trace/cloudtrace.go
jsuereth: If you look right below this line, there already was a `TraceClientOptions` that had the exact same set of types.

exporter/trace/cloudtrace.go
jsuereth: Oh my, just learned about capitals vs. not capitals.... will fix.

exporter/trace/trace.go
jsuereth: Good catch.  I need to turn this one for reals, and the tests needed some tweaking (they double-close the connection)

exporter/trace/cloudtrace.go
```diff
@@ -267,34 +171,44 @@ type Exporter struct {
...
-// for trace provider. Returns provider, flush function, and errors.
+// for trace provider. Returns provider, shutdown function, and errors.
 func NewExportPipeline(opts []Option, topts ...sdktrace.TracerProviderOption) (trace.TracerProvider, func(), error) {
-	exporter, err := NewExporter(opts...)
+	// TODO(suereth): Don't flesh options twice.
+	o := options{Context: context.Background()}
+	for _, opt := range opts {
+		opt(&o)
+	}
+	exporter, err := newExporterWithOptions(&o)
```
dashpole, line 190: `exporter, err := NewExporter(opts...)` seems simpler.  Why is this change needed?

exporter/trace/cloudtrace.go
```diff
@@ -267,34 +171,44 @@ type Exporter struct {
...
 	if err != nil {
 		return nil, nil, err
 	}
-	tp := sdktrace.NewTracerProvider(append(topts, sdktrace.WithSyncer(exporter))...)
-	return tp, exporter.traceExporter.Flush, nil
+	tp := sdktrace.NewTracerProvider(
+		append(topts,
+			sdktrace.WithBatcher(exporter, o.BatchSpanProcessorOptions...))...)
+	return tp, func() {
```
dashpole, line 197: seems like `Shutdown` takes a context on purpose... Should we return a `func(context.Context)` instead of just a `func()`?

exporter/trace/cloudtrace.go
jsuereth: I need access to the opts here, so I decided NOT to recreate them more than once in this flow.

exporter/trace/cloudtrace.go
jsuereth: I think we evaluate that when we evaluate the convenience API.  I want to minimize churn in this change right now.  We have tickets to re-evaluate the convenience API.

PR #150, opened 2021-03-05
Fix #143 - Span status should map rpc status.

### LINKED ISSUE #143: OpenTelemetry span status code does not map directly to google.rpc.Code
Found a similar bug in JS and Python exporters (https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/105)

Currently the code's int value is copied into the span pb if it's not `Ok`:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/f9b9a4b9eab7cc3e9d30c5e8cc9c8e768d60116d/exporter/trace/trace_proto.go#L117-L119

This is fine when it's `Ok`, but the actual error code mapping doesn't make sense:
- OTel [`Error Code = 1`](https://github.com/open-telemetry/opentelemetry-go/blob/0aadfb27bc9e0d6fc0008001720aa5d466c06156/codes/codes.go#L27)
- google.rpc.Code has [`CANCELLED = 1`](https://github.com/googleapis/googleapis/blob/e28aa8570dc86136fe677202c976f4ee00d77f61/google/rpc/code.proto#L41)

jsuereth:
- Update mapping of status code to match trace exporter specification
- Update tests to abide by OTel Specification on "message"
- Update status code mapping to send status in more scenarios

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -30,6 +30,7 @@ import (
 	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 	tracepb "google.golang.org/genproto/googleapis/devtools/cloudtrace/v2"
+	codepb "google.golang.org/genproto/googleapis/rpc/code"
 	statuspb "google.golang.org/genproto/googleapis/rpc/status"
 )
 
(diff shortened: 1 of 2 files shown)


exporter/trace/trace_proto.go
```diff
@@ -115,11 +116,14 @@ func protoFromSpanSnapshot(s *export.SpanSnapshot, projectID string, format Disp
...
 		sp.ParentSpanId = s.ParentSpanID.String()
 	}
-	if s.StatusCode != codes.Ok {
-		sp.Status = &statuspb.Status{Code: int32(s.StatusCode), Message: s.StatusMessage}
-	}
-	if s.StatusCode == codes.Ok && s.StatusMessage != "" {
-		sp.Status = &statuspb.Status{Code: int32(s.StatusCode), Message: s.StatusMessage}
+	if s.StatusCode == codes.Ok {
+		sp.Status = &statuspb.Status{Code: int32(codepb.Code_OK)}
+	} else if s.StatusCode == codes.Unset {
```
aabmass, line 121: Looks like when `s.StatusCode == nil` it would hit the `else` block, but it should go here.

jsuereth: This isn't a pointer, so it can't be nil.

aabmass: oops didn't notice that. I guess the Go SDK defaults to `Unset`

aabmass (approved):

PR #152, opened 2021-03-08
Upgrade to OpenTelemetry 0.18.

charleskorn:
Also updated `Makefile` so that `make update-otel` updates all OpenTelemetry modules.


bogdandrutu:
@jsuereth can you take care of this and do a release, we are behind in the collector and have a hard time updating because of this dependency :(

punya (approved):
@charleskorn thanks for the contribution!

Issue #156, opened 2021-03-19
Unable to set DefaultTraceAttributes of the trace exporter 


filip-ph-johansson:
Reviewing the code in [github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace@v0.18.0/cloudtrace.go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/trace/cloudtrace.go) there's no constructor Option for setting the 
[DefaultTraceAttributes](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/trace/cloudtrace.go#L81) of the Exporter in 
```
// NewExporter creates a new Exporter thats implements trace.Exporter.
func NewExporter(opts ...Option) (*Exporter, error) {
	o := options{Context: context.Background()}
	for _, opt := range opts {
		opt(&o)
	}
	return newExporterWithOptions(&o)
}
```

punya:
The constructor options were created in https://github.com/open-telemetry/opentelemetry-go/pull/160/commits/07c9f9ec68f47bc20930eb5e1a0e40d66c0020c9. It looks like `DefaultTraceAttributes` was left out by mistake (@ymotongpoo can confirm).

@filip-ph-johansson, would you be interested in contributing a construction option to make `DefaultTraceAttributes` configurable?

filip-ph-johansson:
Ok, thanks for your feedback.
Sure, it should be a quick fix.

jsuereth:
Two additions to this:

1. Are the trace attributes being added specific to just the Stackdriver exporter or are they more associated with the resource?  I'd encourage you to use: OTEL_RESOURCE_ATTRIBUTES env variable (or similar config) for this: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration.
2. As an FYI - We're looking into pushing resource/attribute annotations-as-configuration into core OTEL capabilities.   Happy to have it on our exporter, but longer term we'd like this to just be a (consistent) feature of OTEL.

punya:
Note that handling of OTEL_RESOURCE_ATTRIBUTES is not fully spec-compliant at the moment: https://github.com/open-telemetry/opentelemetry-go/issues/1689.

jsuereth:
Yeah, I think we need to provide it here as an interim solution.   Ideally we can push this use case upstream over time though.

[...]

PR #160, opened 2021-03-24
Add ability to set the default trace attributes

filip-ph-johansson:
Fix [#156](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/156)

This adds the missing WithDefaultTraceAttributes option.

### DIFF
--- a/exporter/trace/cloudtrace.go
+++ b/exporter/trace/cloudtrace.go
@@ -78,7 +78,7 @@ type options struct {
 
 	// DefaultTraceAttributes will be appended to every span that is exported to
 	// Stackdriver Trace.
-	DefaultTraceAttributes map[string]interface{}
+	DefaultTraceAttributes []attribute.KeyValue
 
 	// Context allows you to provide a custom context for API calls.
 	//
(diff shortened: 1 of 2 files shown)


exporter/trace/cloudtrace.go
```diff
@@ -256,20 +262,21 @@ func (e *Exporter) Shutdown(ctx context.Context) error {
 	return e.traceExporter.Shutdown(ctx)
 }
 
-func (e *Exporter) sdWithDefaultTraceAttributes(sd *export.SpanSnapshot) *export.SpanSnapshot {
-	newSD := *sd
-	for k, v := range e.traceExporter.o.DefaultTraceAttributes {
+func createKeyValueAttributes(attr map[string]interface{}) []attribute.KeyValue {
+	kv := make([]attribute.KeyValue, len(attr))
```
punya, line 266: It looks like you are trying to reserve capacity in the slice, not create a slice with several zero elements.

```suggestion
	kv := make([]attribute.KeyValue, 0, len(attr))
```

punya (changes requested):
Thanks for the contribution!

exporter/trace/cloudtrace.go
filip-ph-johansson: True, I missed that. 
Fixed.

punya (approved):

Issue #166, opened 2021-04-22
Using resource options produces "Request was missing field timeSeries" error in the logs


geoah:
Hey there,

It seems that adding resource options in the metrics exporter results in some not very informative error logs every time it tries to flush.

```
rpc error: code = InvalidArgument desc = Request was missing field timeSeries.
```

This can be easily replicated by the following service assuming it has access to GCP.
```go
	mopts := []mexporter.Option{
		mexporter.WithInterval(10 * time.Second),
	}
	ropts := basic.WithResource(
		resource.NewWithAttributes(
			attribute.String("instance_id", "abc123"),
			attribute.String("application", "example-app"),
		),
	)
	_, err := mexporter.InstallNewPipeline(mopts, ropts)
	if err != nil {
		panic(err)
	}

	<-make(chan struct{})
```

It will start logging the error mentioned above every 10 seconds.
First log takes a bit longer for some reason (up to 30 seconds).

Note: using `ropts := []basic.Option{}` does not result in the same logs.

---

I'd be really grateful if someone could help out about whether this is expected behaviour, a big, or misconfiguration on our end.

Thank you very much in advance :)

punya:
@geoah thanks for the bug report. Let me try this out and get back to you shortly.

geoah:
@punya if you have any ideas on what this might be related to and can point me to the right direction I could take a look as well. I just really can't pinpoint to why or where this is happening.

ryanlaycock:
Hey, interestingly I seem to be getting the same error:

`2021/05/13 18:38:57 rpc error: code = InvalidArgument desc = Request was missing field timeSeries.`

even when using the **basic options**:

```
        opts := []mexporter.Option{
		mexporter.WithProjectID(projectID),
		mexporter.WithInterval(10 * time.Second),
	}
	ropts:= []basic.Option{}

	// Create exporter (collector embedded with the exporter).
	pusher, err := mexporter.InstallNewPipeline(opts, ropts...)
	if err != nil {
		return MetricExporter{}, err
	}
```

It happens every 10 seconds in this configuration so presume it is every time it tries to export.

punya:
I dropped the ball on this for a few days, but I'm picking it up again shortly. Thanks for your patience!

soellman:
I was also seeing this before:
1. I set the monitored resource type (see #170)
2. I fixed my service account's IAM permissions

[...]

PR #171, opened 2021-05-25
Don't export timeSeries if there are no records in cps

ryanlaycock:
Added a check to ensure there are records in the cps when exporting, otherwise return and don't export. 

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/166 that caused the error `rpc error: code = InvalidArgument desc = Request was missing field timeSeries.` everytime metrics were exported but no values were added to the metricExporter.


punya (approved):
Thanks @ryanlaycock !

dashpole (approved):

PR #173, opened 2021-06-22
Update to otel-go v1.0.0-RC1

0xc0d:
Since OpenTelemetry has been updated to v1.0.0-RC1 I am sending this PR.

Changes are:

- exporter now satisfies otel Exporter interface which needs a slice of `ReadOnlySpan`
 ```go
ExportSpans(context.Context, []*SpanSnapshot) error
ExportSpans(context.Context, []ReadOnlySpan) error 
```
- ~`WithDefaultTraceAttributes` function and `DefaultTraceAttributes` are removed since it is not possible to add attributes to `ReadOnlySpan`~
- `injectLabelsFromResources` is removed since you can't inject anything to a `ReadOnlySpan`. Instead, `attributeWithLabelsFromResources` returns all attributes with labels from resources.
- Tests for `injectLabelsFromResources` are removed since mocking `ReadOnlySpan` [could lead to a problem in future](https://github.com/open-telemetry/opentelemetry-go/blob/694c9a413da927dc999803c2a6f7f2a4a4465837/sdk/trace/span.go#L85). Although it might be an option to change `attributeWithLabelsFromResources`'s signature to get an interface like this:
```go
type AttributeGetter interface {
    Attributes() []attribute.KeyValue
    Resource() *resource.Resource
}
```
With this we can mock an interface to be able to test this function.


example/trace/http/client/client.go
```diff
@@ -58,19 +57,29 @@ func main() {
...
+	}
+	member, err := baggage.NewMember("username", "donuts")
+	if err != nil {
+		panic(err)
+	}
+	b, err = b.SetMember(member)
+	if err != nil {
+		panic(err)
+	}
+	ctx := baggage.ContextWithBaggage(context.Background(), b)
```
punya, line 72: Slightly more concise:

```suggestion
	member, err := baggage.NewMember("username", "donuts")
	if err != nil {
		panic(err)
	}
	b, err := baggage.New(member)
	if err != nil {
		panic(err)
	}
	ctx := baggage.ContextWithBaggage(context.Background(), b)
```

exporter/trace/cloudtrace.go
```diff
@@ -142,21 +138,13 @@ func WithTimeout(t time.Duration) func(o *options) {
...
 // WithDisplayNameFormatter sets the way span's display names will be
-// generated from SpanSnapshot
+// generated from ReadOnlySpan
 func WithDisplayNameFormatter(f DisplayNameFormatter) func(o *options) {
 	return func(o *options) {
 		o.DisplayNameFormatter = f
 	}
 }
 
-// WithDefaultTraceAttributes sets the attributes that will be appended
```
punya, line 152: It should still be possible to keep this functionality by modifying the Google Cloud Trace proto at export time, even if it's no longer possible to modify the OTel `ReadOnlySpan`.

punya (changes requested):
Thanks for contributing this @0xc0d !

exporter/trace/cloudtrace.go
0xc0d: Yes, I see. I will add this to the `ConvertSpan` function.

0xc0d: I have change `protoFromReadOnlySpan` function signature to accept a list of default trace attributes. This values will be added to the attribute set within the function.
What do you think?

punya (approved):

exporter/trace/cloudtrace.go
punya: This approach makes sense, thank you!

PR #178, opened 2021-07-01
Send span links to Cloud Trace

### LINKED ISSUE #176: Implement links in trace exporter
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/150c8dab3e9a7f336690514cfea6bf2f7b8c7826/exporter/trace/trace_proto.go#L181

aabmass:
Fixes #176

### DIFF
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -186,26 +186,27 @@ func protoFromReadOnlySpan(s sdktrace.ReadOnlySpan, defaultTraceAttributes []att
 }
 
 func linksProtoFromLinks(links []trace.Link) *tracepb.Span_Links {
-	var linksPb *tracepb.Span_Links
-
-	if numLinks := len(links); numLinks > 0 {
-		linksPb = &tracepb.Span_Links{}
-		numLinksToKeep := numLinks
-		if numLinksToKeep > maxNumLinks {
-			numLinksToKeep = maxNumLinks
-		}
-
-		for _, link := range links[:numLinksToKeep] {
-			linkPb := &tracepb.Span_Link{
-				TraceId: link.SpanContext.TraceID().String(),
-				SpanId:  link.SpanContext.SpanID().String(),
-				Type:    tracepb.Span_Link_TYPE_UNSPECIFIED,
-			}
-			copyAttributes(&linkPb.Attributes, link.Attributes)
-			linksPb.Link = append(linksPb.Link, linkPb)
-		}
-		linksPb.DroppedLinksCount = clip32(numLinks - numLinksToKeep)
+	numLinks := len(links)
+	if numLinks == 0 {
+		return nil
 	}
+
+	linksPb := &tracepb.Span_Links{}
+	numLinksToKeep := numLinks
+	if numLinksToKeep > maxNumLinks {
+		numLinksToKeep = maxNumLinks
+	}
+
+	for _, link := range links[:numLinksToKeep] {
+		linkPb := &tracepb.Span_Link{
+			TraceId: link.SpanContext.TraceID().String(),
+			SpanId:  link.SpanContext.SpanID().String(),
+			Type:    tracepb.Span_Link_TYPE_UNSPECIFIED,
+		}
+		copyAttributes(&linkPb.Attributes, link.Attributes)
+		linksPb.Link = append(linksPb.Link, linkPb)
+	}
+	linksPb.DroppedLinksCount = clip32(numLinks - numLinksToKeep)
 
 	return linksPb
 }
--- a/exporter/trace/trace_proto.go
+++ b/exporter/trace/trace_proto.go
@@ -26,6 +26,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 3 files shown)


exporter/trace/trace_proto_test.go
```diff
@@ -0,0 +1,79 @@
...
+				attribute.String("hello", "world"),
+			},
+		}
+		linksPb := linksProtoFromLinks([]trace.Link{link})
+
+		assert.NotNil(t, linksPb)
+		assert.EqualValues(t, linksPb.DroppedLinksCount, 0)
+		assert.Len(t, linksPb.Link, 1)
+		assert.Equal(t, spanContext.TraceID().String(), linksPb.Link[0].TraceId)
+		assert.Equal(t, spanContext.SpanID().String(), linksPb.Link[0].SpanId)
```
dashpole, line 52: also assert that the type is unspecified.

exporter/trace/trace_proto.go
```diff
@@ -178,11 +180,36 @@ func protoFromReadOnlySpan(s sdktrace.ReadOnlySpan, defaultTraceAttributes []att
...
 	}
 
-	// TODO(ymotongpoo): add implementations for Links
+	sp.Links = linksProtoFromLinks(s.Links())
 
 	return sp
 }
 
+func linksProtoFromLinks(links []trace.Link) *tracepb.Span_Links {
+	var linksPb *tracepb.Span_Links
```
dashpole, line 189: nit: you can instantiate a return value in the function definition, and avoid the extra declaration (and you don't need to specify anything when you return):
```golang
func linksProtoFromLinks(links []trace.Link) (linksPb *tracepb.Span_Links) {
    ...
    return
}
```

dashpole (approved):
Looks good overall.

exporter/trace/trace_proto.go
bshi: (Sorry for the fly-by)

Please don’t use naked returns for this purpose. 

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

exporter/trace/trace_proto.go
```diff
@@ -178,11 +180,36 @@ func protoFromReadOnlySpan(s sdktrace.ReadOnlySpan, defaultTraceAttributes []att
 		sp.TimeEvents.DroppedAnnotationsCount = clip32(droppedAnnotationsCount)
 	}
 
-	// TODO(ymotongpoo): add implementations for Links
+	sp.Links = linksProtoFromLinks(s.Links())
```
0xc0d, line 183:
```suggestion
	spanLinks := s.Links()

	numLinks := len(spanLinks)
	numLinksToKeep := numLinks
	if numLinksToKeep > maxNumLinks {
		numLinksToKeep = maxNumLinks
	}

	sp.Links = linksProtoFromLinks(spanLinks[:numLinksToKeep])
	sp.Links.DroppedLinksCount = clip32(numLinks - numLinksToKeep)
```

exporter/trace/trace_proto.go
```diff
@@ -178,11 +180,36 @@ func protoFromReadOnlySpan(s sdktrace.ReadOnlySpan, defaultTraceAttributes []att
...
+				Type:    tracepb.Span_Link_TYPE_UNSPECIFIED,
+			}
+			copyAttributes(&linkPb.Attributes, link.Attributes)
+			linksPb.Link = append(linksPb.Link, linkPb)
+		}
+		linksPb.DroppedLinksCount = clip32(numLinks - numLinksToKeep)
+	}
+
+	return linksPb
+}
```
0xc0d, line 211: Isn't it out of `linksProtoFromLinks`'s concern to drop extra links?
```suggestion
func linksProtoFromLinks(links []trace.Link) *tracepb.Span_Links {
	if len(links) == 0 {
		return nil
	}

	linksPb := &tracepb.Span_Links{
		Link: make([]*tracepb.Span_Link, 0, len(links)),
	}
	for _, link := range links {
		linkPb := &tracepb.Span_Link{
			TraceId: link.SpanContext.TraceID().String(),
			SpanId:  link.SpanContext.SpanID().String(),
			Type:    tracepb.Span_Link_TYPE_UNSPECIFIED,
		}
		copyAttributes(&linkPb.Attributes, link.Attributes)
		linksPb.Link = append(linksPb.Link, linkPb)
	}

	return linksPb
}
```

0xc0d (changes requested):

exporter/trace/trace_proto.go
aabmass: Btw, I think ultimately we should remove all of the dropping logic from the exporter and pass [span limits](https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v0.20.0/trace#WithSpanLimits) to the SDK in `NewExportPipeline()`.

exporter/trace/trace_proto.go
aabmass: I restructured this a bit

exporter/trace/trace_proto.go
0xc0d: Well, the function name says it is supposed to convert span links to protobuf span links, not dropping them. What you see is not what you get something.

But, yes I agree not to drop anything unless it is explicitly there.

jsuereth (approved):

exporter/trace/trace_proto.go
aabmass: @jsuereth made a good point offline that setting these limits will effect ALL exporters if the user has set up multiple pipelines. So I will keep the dropping and tweak this function name

PR #180, opened 2021-07-02
Set span-kind field in trace_proto.go

tbarker25:


exporter/trace/trace_proto.go
```diff
@@ -292,3 +295,24 @@ func clip32(x int) int32 {
...
+		// SpanKindUnspecified is an unspecified SpanKind and is not a
+		// valid SpanKind. SpanKindUnspecified should be replaced with
+		// SpanKindInternal if it is received.
+		return tracepb.Span_INTERNAL
+	case trace.SpanKindServer:
+		return tracepb.Span_SERVER
+	case trace.SpanKindClient:
+		return tracepb.Span_CLIENT
+	case trace.SpanKindProducer:
+		return tracepb.Span_CONSUMER
```
0xc0d, line 311:
```suggestion
		return tracepb.Span_PRODUCER
```

0xc0d (changes requested):

exporter/trace/trace_proto.go
jsuereth: +1

jsuereth (changes requested):

exporter/trace/trace_proto.go
tbarker25: Good catch. Fixed.

exporter/trace/trace_proto.go
```diff
@@ -292,3 +295,24 @@ func clip32(x int) int32 {
...
+	case trace.SpanKindServer:
+		return tracepb.Span_SERVER
+	case trace.SpanKindClient:
+		return tracepb.Span_CLIENT
+	case trace.SpanKindProducer:
+		return tracepb.Span_PRODUCER
+	case trace.SpanKindConsumer:
+		return tracepb.Span_CONSUMER
+	default:
+		log.Printf("warning: unknown span kind %v", kind)
```
dashpole, line 315: This doesn't need to happen in this PR, since it probably spans the entire operations-go repo, but I wonder if we should be using the [otel error handler](https://github.com/open-telemetry/opentelemetry-go/blob/d616df61f5d163589228c5ff3be4aa5415f5a884/handler.go) here instead of logging.  Printing directly in libraries can result in log spam that users don't have any way to turn off

dashpole: Actually, it looks like we have a different way of allowing users to handle errors already: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/150c8dab3e9a7f336690514cfea6bf2f7b8c7826/exporter/trace/trace.go#L101.  Can we use that here?

tbarker25: Discussed offline. I removed the logging call as other switch-case statements in this file don't have error-handling for an unknown kind.

Agree that logging in libraries is a bad idea.

dashpole (approved):

aabmass (approved):

aabmass:
@jsuereth is this good to go now?

0xc0d (approved):

jsuereth (approved):

PR #182, opened 2021-07-12
Introduce ReadOnlySpan that callers can implement

### LINKED ISSUE #181: The new github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace will be impossible to be used in otel-collector
Right now after the upgrade to 1.0.0-RC1 this library requires `sdktrace.ReadOnlySpan` to be passed to the exporter, which otel-collector will not be able to produce since `sdktrace.ReadOnlySpan` are private interface (has a private method).

This is very urgent since it blocks the otel-collector release see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4146

punya:
It's compatible with sdktrace.ReadOnlySpan, but doesn't have the
"private" marker. This means we can adapt the GCP exporter in the
collector to produce instances of this interface.

Fixes #181.

### DIFF
--- a/exporter/trace/cloudtrace.go
+++ b/exporter/trace/cloudtrace.go
@@ -37,7 +37,7 @@ type Option func(*options)
 
 // DisplayNameFormatter is is a function that produces the display name of a span
 // given its ReadOnlySpan
-type DisplayNameFormatter func(sdktrace.ReadOnlySpan) string
+type DisplayNameFormatter func(ReadOnlySpan) string
 
 // options contains options for configuring the exporter.
 type options struct {
--- a/exporter/trace/cloudtrace_test.go
+++ b/exporter/trace/cloudtrace_test.go
@@ -116,7 +116,7 @@ func TestExporter_DisplayNameFormatter(t *testing.T) {
 	clientOpt := []option.ClientOption{option.WithGRPCConn(mock.ClientConn())}
 
 	spanName := "span1234"
-	format := func(s sdktrace.ReadOnlySpan) string {
+	format := func(s ReadOnlySpan) string {
 		return "TEST_FORMAT" + s.Name()
 	}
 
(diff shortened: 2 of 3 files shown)


dashpole (approved):
I think it is questionable whether this belongs in the exporter at all.  For example, otelhttp has a [WithSpanNameFormatter](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithSpanNameFormatter) that users should use instead.  But this is a reasonable solution if we want to keep that option.

punya:
@dashpole I'm also dubious about whether this is the right approach, but I want to unblock upgrading the GCP collector exporter, whose [logic](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6c4c89634b05a375af9d073c0ce77fc41e7de4de/exporter/googlecloudexporter/spandata.go#L68-L81) (excerpt below) would be hard to express using `With...` methods.

```go
	sd := &sdktrace.SpanSnapshot{
		SpanContext:              apitrace.NewSpanContext(sc),
		Parent:                   apitrace.NewSpanContext(parentSc),
		SpanKind:                 pdataSpanKindToOTSpanKind(span.Kind()),
		StartTime:                startTime,
		EndTime:                  endTime,
		Name:                     span.Name(),
		Attributes:               pdataAttributesToOTAttributes(span.Attributes(), resource),
		Links:                    pdataLinksToOTLinks(span.Links()),
		MessageEvents:            pdataEventsToOTMessageEvents(span.Events()),
		DroppedAttributeCount:    int(span.DroppedAttributesCount()),
		DroppedMessageEventCount: int(span.DroppedEventsCount()),
		DroppedLinkCount:         int(span.DroppedLinksCount()),
		Resource:                 r,
	}
```

tbarker25 (approved):

PR #184, opened 2021-07-14
Add YAMLs for e2e tests.

tbarker25:
BUG=191643395


cloudbuild-e2e-cloud-run.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2021 Google LLC
```
aabmass, line 1: Delete this one for now, cloud run requires the "push" mode which isn't implemented yet for Go

aabmass (commented):
Just need to create the triggers and these tests should run on this PR

cloudbuild-e2e-cloud-run.yaml
tbarker25: Done

aabmass (approved):
Awesome LGTM

PR #187, opened 2021-07-20
Remove 'our' ReadOnlySpan interface

punya:
Instead of creating a new interface, we can embed the existing
interface to inherit the private func.


dashpole (approved):
nice

punya:
Thanks @bogdandrutu for the suggestion.

PR #196, opened 2021-07-20
Use the otel ErrorHandler for handling errors.

### LINKED ISSUE #189: Use opentelemetry-go's error handling for the trace exporter
OpenTelemetry provides global error handlers.  As a user, I would prefer to be able to override the error handler once, and have all of my OpenTelemetry errors handled.  If I don't want to use globals, I would still prefer to instantiate a single error handler, and then pass it to all relevant OpenTelemetry components.

Changes needed:
* Instead of [logging errors by default](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/552dfa270c46b4bf6b1763281e824c6cc23ef15f/exporter/trace/cloudtrace.go#L165), we should use the global error handler by default.
* Replace the [WithOnError option](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/552dfa270c46b4bf6b1763281e824c6cc23ef15f/exporter/trace/cloudtrace.go#L116) with a `WithErrorHandler` option, which takes an `otel.ErrorHandler` instead of a func(error).

dashpole:
Fixes #189.

This also unexports the field in options that contains the error handler.


punya (approved):

punya:
@dashpole would you mind resolving conflicts?

exporter/trace/trace.go
```diff
@@ -98,7 +98,7 @@ func (e *traceExporter) uploadSpans(ctx context.Context, spans []*tracepb.Span)
 	if err != nil {
 		// TODO(ymotongpoo): handle detailed error categories
 		// span.SetStatus(codes.Unknown)
-		e.o.handleError(err)
+		e.o.handleError(fmt.Errorf("failed to export to Stackdriver: %v", err))
```
punya, line 101:
```suggestion
		e.o.handleError(fmt.Errorf("failed to export to Google Cloud Monitoring: %w", err))
```

PR #199, opened 2021-07-26
Update to OTel-Go v1.0.0-RC2 / v0.22.0

punya:


exporter/trace/trace_proto.go
```diff
@@ -189,7 +189,7 @@ func protoFromReadOnlySpan(s sdktrace.ReadOnlySpan, defaultTraceAttributes []att
 // Converts OTel span links to Cloud Trace links proto in order. If there are
 // more than maxNumLinks links, the first maxNumLinks will be taken and the rest
 // dropped.
-func linksProtoFromLinks(links []trace.Link) *tracepb.Span_Links {
+func linksProtoFromLinks(links []sdktrace.Link) *tracepb.Span_Links {
```
jsuereth, line 192: Is this one of those private things that will break the collector?

jsuereth (approved):

exporter/trace/trace_proto.go
punya: This func is un-exported, so I can't imagine how it would break the collector. That's a good call-out in general though :+1:

Issue #110, opened 2020-10-27
Implement propagators


bshi:
Presently, trace context via HTTP header `X-Cloud-Trace-Context` (or the equivalent gRPC header) doesn't appear to be extracted from inbound requests.

OpenCensus equivalent: https://pkg.go.dev/go.opencensus.io/exporter/stackdriver/propagation

https://pkg.go.dev/go.opentelemetry.io/otel/propagators

glassonion1:
I thought i needed too.
However, I thought that it should be implemented in open [telemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib) instead of exporter.

I tried implementing http propagator in my library below.
https://github.com/glassonion1/logz/blob/main/internal/propagation/http_format.go

punya:
Related to #116.

punya:
Related to https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/254 -- we should make sure we comply with the v2 spec.

amammay:
just so i can have some ease of re-use in some of my own projects i went ahead and created a package based off @glassonion1 implementation with just some slight tweaks to work with open telemetry v0.20.0

https://github.com/amammay/propagationgcp

PR #200, opened 2021-07-28
Add XCTC propagator

ymotongpoo:
Basically this is a port of [Python version](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/opentelemetry-propagator-gcp/src/opentelemetry/propagators/cloud_trace_propagator/__init__.py)

Solves #110


propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+		sc.TraceID().String(),
+		sc.SpanID().String(),
+		sc.TraceFlags().String(),
+	)
+	carrier.Set(TraceContextHeaderName, header)
+}
+
+// Extract extacts a context from the carrier if the header contains Google Cloud Trace header format.
+func (p *cloudTraceFormatPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
+	if ctx == nil {
```
dashpole, line 76: It doesn't look like we do this check in any of the upstream propagators.  E.g. https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/trace_context.go#L73

dashpole (commented):
Looks good overall.

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+		ctx = context.Background()
+	}
+
+	header := p.getHeaderValue(carrier)
+	if header != "" {
```
punya, line 81: Is this supposed to be `==` rather than `!=`?

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+		switch n {
+		case "trace_id":
+			traceID = match[i]
+		case "span_id":
+			spanID = match[i]
+		case "trace_flags":
+			traceFlags = match[i]
+		}
+	}
+	if traceID == strings.Repeat("0", 32) || spanID == "0" {
```
punya, line 101: Please add a comment explaining this behavior.

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+
+	tid, err := trace.TraceIDFromHex(traceID)
+	if err != nil {
+		return ctx
+	}
+	sid, err := trace.SpanIDFromHex(spanID)
+	if err != nil {
+		return ctx
+	}
+	tf := trace.TraceFlags(0x01)
```
punya, line 113: What do these flag values mean?

propagator/propagator.go
ymotongpoo: yes, you are right! thanks for catching this
c.f. https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/opentelemetry-propagator-gcp/src/opentelemetry/propagators/cloud_trace_propagator/__init__.py#L65-L66

propagator/propagator.go
g-easy: The only defined flag value is 1, which means the request is sampled (i.e. it will be recorded in a backend). Non-sampled requests can still propagate IDs.

https://cloud.google.com/trace/docs/setup#force-trace

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	if traceID == strings.Repeat("0", 32) || spanID == "0" {
+		return ctx
+	}
+
+	tid, err := trace.TraceIDFromHex(traceID)
+	if err != nil {
+		return ctx
+	}
+	sid, err := trace.SpanIDFromHex(spanID)
```
g-easy, line 109: SpanID is decimal. :(

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Inject injects a context to the carrier following Google Cloud Trace format.
+func (p *cloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
+	span := trace.SpanFromContext(ctx)
+	sc := span.SpanContext()
+
+	header := fmt.Sprintf("%s/%s;o=%s",
+		sc.TraceID().String(),
+		sc.SpanID().String(),
```
g-easy, line 68: SpanID needs to be decimal in this header format. :(

propagator/propagator.go
```diff
@@ -0,0 +1,132 @@
...
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// TraceContextHeaderName is the HTTP header field for Google Cloud Trace
+// https://cloud.google.com/trace/docs/setup#force-trace
+const TraceContextHeaderName = "x-cloud-trace-context"
+
+// TraceContextHeaderFormat is the regular expression pattan for valid Cloud Trace header value
+const TraceContextHeaderFormat = "(?P<trace_id>[0-9a-f]{32})/(?P<span_id>[0-9]{1,20});o=(?P<trace_flags>[0-9])"
```
g-easy, line 32: The ";o=[01]" part is optional. (o defaults to zero)

propagator/propagator.go
ymotongpoo: XCTC's `TRACE_TRUE` should correspond to OTel's `trace_flag`.

https://cloud.google.com/trace/docs/setup#force-trace
https://www.w3.org/TR/trace-context/#sampled-flag

propagator/propagator.go
ymotongpoo: This is the non-recording Span defined in the spec.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk

c.f. https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/opentelemetry-propagator-gcp/src/opentelemetry/propagators/cloud_trace_propagator/__init__.py#L76-L77

propagator/propagator.go
ymotongpoo: thanks! removed the lines.

propagator/propagator.go
ymotongpoo: thanks! I missed the point. I update not to make ";o=[01]" part mandatory.

propagator/propagator.go
ymotongpoo: thanks for pointing this out

propagator/propagator.go
ymotongpoo: thanks for pointing this. I fixed.

ymotongpoo:
I found some issues, so I'll add some changes on top of this.

propagator/propagator.go
```diff
@@ -0,0 +1,147 @@
...
+	header := fmt.Sprintf("%s/%d;o=%s",
+		sc.TraceID().String(),
+		sid,
+		sc.TraceFlags().String(),
+	)
+	carrier.Set(TraceContextHeaderName, header)
+}
+
+// Extract extacts a context from the carrier if the header contains Google Cloud Trace header format.
+func (p *cloudTraceFormatPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
```
jsuereth, line 84: QQ:  How does the Go SDK handle precedence rules here?

If a trace is propogated via XCTC and another via W3C,  which one wins?   In Java, we had to add EXPLICIT "If a trace is already available in Context, don't do anything" here due to lacking specification around this scenario of mulitple propagators.    Can you confirm whether the SDK handles the case of multiple propagators or whether or not you need to supress this extract if a previous propagator has already pulled context?

ymotongpoo: Good question. As far as I confirmed the upstream contrib propagators for b3 and jaeger, they don't have such a validation, so I followed those.

* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/05ef436536dc499361b6c9d3546662a99c2f918f/propagators/jaeger/jaeger_propagator.go#L84-L157
* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/05ef436536dc499361b6c9d3546662a99c2f918f/propagators/b3/b3_propagator.go#L133-L161

My understanding is that the only global propagator always wins and other format will be ignored if we don't specify the format with `otel.NewCompsiteTextMapPropagator`.

propagator/propagator.go
```diff
@@ -0,0 +1,152 @@
...
+		switch n {
+		case "trace_id":
+			traceID = match[i]
+		case "span_id":
+			spanID = match[i]
+		case "trace_flags":
+			traceFlags = match[i]
+		}
+	}
+	// non-reording Span
```
aabmass, line 108:
```suggestion
	// non-recording Span
```

propagator/propagator.go
```diff
@@ -0,0 +1,152 @@
...
+	return CloudTraceFormatPropagator{}
+}
+
+func (p CloudTraceFormatPropagator) getHeaderValue(carrier propagation.TextMapCarrier) string {
+	header := carrier.Get(TraceContextHeaderName)
+	if header != "" {
+		return header
+	}
+
+	for _, key := range carrier.Keys() {
```
aabmass, line 53: It might be worth adding a comment explaining that this is to handle different casings of the header name if not initially found. It's not immediately obvious otherwise

aabmass (commented):
Thanks so much for doing this Yoshi!

propagator/propagator.go
ymotongpoo: Yes, you are right! I added the comment here.

propagator/propagator.go
jsuereth: Right, but that doesn't mean all of these don't suffer from this issue (which we've seen consistently show up in all SDKs).  It looks like Go has a "last one wins" philsophy basaed on this code: https://github.com/open-telemetry/opentelemetry-go/blob/2b1bb29ef89708de6a0676d3bf7aa6162594cf35/propagation/propagation.go#L95

This means *any* documentation you have around XCTC should list XCTC *before* w3c trace context so that we ensure we're prioritizing attaching to the new standard vs. the legacy traces.   This should also help with Google's load balancers "cutting" traces by providing a new XCTC when you cross a LB/infra boundary.

punya:
I just noticed that #200 is ready to merge, so I'll get that in and update this PR afterwards.

propagator/propagator.go
```diff
@@ -0,0 +1,158 @@
...
+	"strconv"
+	"strings"
+
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// TraceContextHeaderName is the HTTP header field for Google Cloud Trace
+// https://cloud.google.com/trace/docs/setup#force-trace
+const TraceContextHeaderName = "x-cloud-trace-context"
```
punya, line 31: Do the constants in this file need to be exported?

propagator/propagator.go
```diff
@@ -0,0 +1,158 @@
...
+	return CloudTraceFormatPropagator{}
+}
+
+func (p CloudTraceFormatPropagator) getHeaderValue(carrier propagation.TextMapCarrier) string {
+	header := carrier.Get(TraceContextHeaderName)
+	if header != "" {
+		return header
+	}
+
+	// Currently, header name can be in different cases (eg. x-cloud-trace-context),
```
punya, line 53: [According to the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#:~:text=The%20Get%20function%20is%20responsible%20for%20handling%20case%20sensitivity.%20If%20the%20getter%20is%20intended%20to%20work%20with%20a%20HTTP%20request%20object%2C%20the%20getter%20MUST%20be%20case%20insensitive.)
> The Get function is responsible for handling case sensitivity. If the getter is intended to work with a HTTP request object, the getter MUST be case insensitive.

so the following loop isn't necessary.

propagator/propagator.go
```diff
@@ -0,0 +1,158 @@
...
+// Inject injects a context to the carrier following Google Cloud Trace format.
+func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
+	span := trace.SpanFromContext(ctx)
+	sc := span.SpanContext()
+
+	// https://cloud.google.com/trace/docs/setup#force-trace
+	// Trace ID: 32-char hexadecimal value representing a 128-bit number.
+	// Span ID: decimal representation of the unsigned interger.
+	sidHex := sc.SpanID().String()
+	sid, err := strconv.ParseUint(sidHex, 16, 64)
```
punya, line 75: This is still parsing the span ID in base 16, though the doc says it's supposed to be decimal.

propagator/propagator.go
```diff
@@ -0,0 +1,158 @@
...
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// TraceContextHeaderName is the HTTP header field for Google Cloud Trace
+// https://cloud.google.com/trace/docs/setup#force-trace
+const TraceContextHeaderName = "x-cloud-trace-context"
+
+// TraceContextHeaderFormat is the regular expression pattan for valid Cloud Trace header value
+const TraceContextHeaderFormat = "(?P<trace_id>[0-9a-f]{32})/(?P<span_id>[0-9]{1,20})(;o=(?P<trace_flags>[0-9]))?"
```
punya, line 34: This is used with `FindStringSubmatch`, so it should be anchored `^...$`.

propagator/propagator.go
punya: The conversion from XCTC's trace flag to OTel's trace flag is explicit in this function, but implicit in the other one (i.e. they are assumed to have the same semantics). Why?

propagator/propagator.go
ymotongpoo: because `sc.SpanID().String()` returns hex value, it needs to be parsed in base 16. this is intended.

propagator/propagator.go
ymotongpoo: Agreed. Added the head and tail notions.

propagator/propagator.go
ymotongpoo: I feel at least the header is good to be exposed for reference purpose. For now I moved other constants and variables for headers such as regexp format and compiled pattern to private.

propagator/propagator.go
ymotongpoo: Thanks. Added test for getHeaderValue and confirmed the behavior.

propagator/propagator.go
ymotongpoo: the conversion is just to convert the `string` value to `trace.TraceFlags`, i.e. `byte`.  Am I misunderstanding your point?

propagator/propagator.go
aabmass: `sc.SpanID()` is already `[8]byte`, so you could avoid the hex conversion. I suppose it should be BE, so `binary.BigEndian.Uint64(sc.SpanID()[:])` maybe?

propagator/propagator.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// TraceContextHeaderName is the HTTP header field for Google Cloud Trace
+// https://cloud.google.com/trace/docs/setup#force-trace
+const TraceContextHeaderName = "x-cloud-trace-context"
+
+// traceContextHeaderFormat is the regular expression pattan for valid Cloud Trace header value
```
aabmass, line 33:
```suggestion
// traceContextHeaderFormat is the regular expression pattern for valid Cloud Trace header value
```

propagator/propagator.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		log.Printf("CloudTraceFormatPropagator: invalid trace id %#v: %v", traceID, err)
+		return ctx
+	}
+	sidUint, err := strconv.ParseUint(spanID, 10, 64)
+	if err != nil {
+		log.Printf("CloudTraceFormatPropagator: on span ID conversion: %v", err)
+		return ctx
+	}
+	sid, err := trace.SpanIDFromHex(fmt.Sprintf("%016x", sidUint))
```
aabmass, line 117: Same comment as above about the hex conversion, would using `binary.BigEndian.PutUint64()` or something similar work?

aabmass (approved):
Left some nits but LGTM

propagator/propagator.go
ymotongpoo: @punya @aabmass I forgot the use of `encoding/binary`. Yes, you are right. I changed the code to use it.

propagator/propagator.go
punya: I'm probably missing something here, which is okay because I see that Aaron has approved the code after a careful review. I'll learn more about it elsewhere.

PR #202, opened 2021-08-06
Update release tool to handle split versions

### LINKED ISSUE #185: Handle split versions in release tooling and docs
From now until the full API/SDK are stable, otel-go will release a mix of module versions (currently v0.21.0 and v1.0.0-RC1). We should improve our tooling to handle this properly, otherwise our releases will be more time-consuming and error-prone.

punya:
Fixes #185.


aabmass (approved):
LGTM

PR #203, opened 2021-08-10
Change to use encoding/binary for SpanID conversion

ymotongpoo:
In the context of the update for XCTC (#200), I didn't push the changes around SpanID conversion.

* Change to use `binary/encoding` instead of `trace.SpanID().String()` and `trace.SpanIDFromHex()`
* Add comments public functions/methods

### DIFF
--- a/propagator/propagator.go
+++ b/propagator/propagator.go
@@ -55,6 +55,7 @@ func (p CloudTraceFormatPropagator) getHeaderValue(carrier propagation.TextMapCa
 }
 
 // Inject injects a context to the carrier following Google Cloud Trace format.
+// In this method, SpanID is expected to be stored in big endian.
 func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
 	span := trace.SpanFromContext(ctx)
 	sc := span.SpanContext()
@@ -78,6 +79,7 @@ func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagat
 }
 
 // Extract extacts a context from the carrier if the header contains Google Cloud Trace header format.
+// In this method, SpanID in carrier is decimal, and it is converted to trace.SpanID in big endian.
 func (p CloudTraceFormatPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
 	header := p.getHeaderValue(carrier)
 	if header == "" {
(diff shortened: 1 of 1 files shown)


propagator/propagator.go
```diff
@@ -56,11 +62,9 @@ func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagat
 	// https://cloud.google.com/trace/docs/setup#force-trace
 	// Trace ID: 32-char hexadecimal value representing a 128-bit number.
 	// Span ID: decimal representation of the unsigned interger.
-	sidHex := sc.SpanID().String()
-	sid, err := strconv.ParseUint(sidHex, 16, 64)
-	if err != nil {
-		return
-	}
+	ary := sc.SpanID()
+	sid := binary.BigEndian.Uint64(ary[:])
```
punya, line 66: Is the endian-ness a publicly documented part of the SpanID API?

aabmass: I was also not sure on this hoping for other's inputs, but my guess is "not applicable". OTel spec says:

> [`SpanId` A valid span identifier is an 8-byte array with at least one non-zero byte.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#:~:text=SpanId%20A%20valid%20span%20identifier%20is%20an%208-byte%20array%20with%20at%20least%20one%20non-zero%20byte.)

Since it's just defined as a byte array, there is no endianness or MSB/LSB right? The XCTC context spec just says

> [`SPAN_ID` is the decimal representation of the (unsigned) span ID.](https://cloud.google.com/trace/docs/setup#:~:text=SPAN_ID%20is%20the%20decimal%20representation%20of%20the%20(unsigned)%20span%20ID.)

My take is it doesn't matter as long as we are consistent across implementations on how we interpret the byte array as uint64. Wdyt?

punya: I agree, it doesn't matter as long as we're consistent across implementations. I'm happy to approve and merge this PR, but in the near future we should create a doc about XCTC-OTel requirements so that we can actually keep this consistent.

punya (approved):

aabmass (approved):
LGTM. We should probably spec out how the OTel span ID's should be converted from bytes to base-10 uint64s for XCTC somewhere, and define some standard test cases that get included in each language.

ymotongpoo:
@punya @aabmass thanks for the review and comments. Given the needs to spec out the conversion process, I added the comments on `Inject` and `Extract` for now.

PR #204, opened 2021-08-12
Add SpanContextFromRequest

### LINKED ISSUE #116: Need SpanContextFromRequest
Need an analog of [this](https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/stackdriver/propagation/http.go#L43) convenience function for the GCP specific header `X-Cloud-Trace-Context`

ymotongpoo:
Fixes #116 

* Extrace spanContextFromHeader out of Extrace and use it as common logic for Extract and SpanContextFromReqeust
* Change return type of `New()` from `propagation.TextMapPropagator` to `CloudTraceFormatPropagator` so that the caller call `SpanContextFromRequest` without type casting
* Fixes a bug where TraceFlags is not properly set when `;o=TRACE_TRUE` part is not set.

### DIFF
--- a/propagator/propagator.go
+++ b/propagator/propagator.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


ymotongpoo:
Oops, this introduces degradation. I'll create another pull request that patches against #203

PR #205, opened 2021-08-12
Add SpanContextFromRequest

### LINKED ISSUE #116: Need SpanContextFromRequest

ymotongpoo:
Fixes #116. Merge this after merging #203.

* Extrace spanContextFromHeader out of Extrace and use it as common logic for Extract and SpanContextFromReqeust
* Change return type of `New()` from `propagation.TextMapPropagator` to `CloudTraceFormatPropagator` so that the caller call `SpanContextFromRequest` without type casting
* Fixes a bug where TraceFlags is not properly set when `;o=TRACE_TRUE` part is not set.

### DIFF
--- a/propagator/propagator.go
+++ b/propagator/propagator.go
@@ -55,6 +55,7 @@ func (p CloudTraceFormatPropagator) getHeaderValue(carrier propagation.TextMapCa
 }
 
 // Inject injects a context to the carrier following Google Cloud Trace format.
+// In this method, SpanID is expected to be stored in big endian.
 func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
 	span := trace.SpanFromContext(ctx)
 	sc := span.SpanContext()
@@ -78,6 +79,7 @@ func (p CloudTraceFormatPropagator) Inject(ctx context.Context, carrier propagat
 }
 
 // Extract extacts a context from the carrier if the header contains Google Cloud Trace header format.
+// In this method, SpanID in carrier is decimal, and it is converted to trace.SpanID in big endian.
 func (p CloudTraceFormatPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
 	header := p.getHeaderValue(carrier)
 	if header == "" {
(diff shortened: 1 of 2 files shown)


propagator/propagator.go
```diff
@@ -131,7 +132,27 @@ func (p CloudTraceFormatPropagator) Extract(ctx context.Context, carrier propaga
 		TraceFlags: tf,
 		Remote:     true,
 	}
-	sc := trace.NewSpanContext(scConfig)
+	return trace.NewSpanContext(scConfig), true
+}
+
+// SpanContextFromRequest extracts a trace.SpanContext from the HTTP request req.
+// In this method, SpanID is expected to be stored in big endian.
+func (p CloudTraceFormatPropagator) SpanContextFromRequest(req *http.Request) (sc trace.SpanContext, ok bool) {
```
dashpole, line 140: should we have this return an error, so users can debug if they want to?

Extract can print the error (since it can't return it), and we can return it here.

ymotongpoo: @dashpole fair enough. I replaced bool with err.

propagator/propagator_test.go
```diff
@@ -289,9 +289,9 @@ func TestSpanContextFromRequest(t *testing.T) {
 			req := httptest.NewRequest("GET", "http://example.com", nil)
 			req.Header.Set(tc.key, fmt.Sprintf("%s/%s%s", tc.traceID, tc.spanID, tc.flagPart))
 
-			sc, ok := propagator.SpanContextFromRequest(req)
-			if !ok {
-				t.Errorf("%v: SpanContextFromRequest returned false", tc.name)
+			sc, err := propagator.SpanContextFromRequest(req)
+			if err != nil {
+				t.Errorf("%v: SpanContextFromRequest returned false: %v", tc.name, err)
```
dashpole, line 294: nit: update the test error message, since we can't return false anymore.

dashpole (approved):
lgtm after nit

propagator/propagator_test.go
ymotongpoo: oops, thanks for catching this. updated the message.

Issue #210, opened 2021-09-29
exporter/metric uses deprecated (and removed) testing method of go.opentelemetry.io/otel/sdk/metric


mishas:
Trying to use github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric together with the latest version of go.opentelemetry.io/otel/sdk/metric (v0.23.0) fails with the following error:

```
go: finding module for package go.opentelemetry.io/otel/sdk/export/metric/metrictest
example.com/pkg/metric imports
	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric tested by
	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric.test imports
	go.opentelemetry.io/otel/sdk/export/metric/metrictest: module go.opentelemetry.io/otel/sdk/export/metric@latest found (v0.23.0), but does not contain package go.opentelemetry.io/otel/sdk/export/metric/metrictest
```

Tried with versions v0.20.1, v0.21.0, v0.22.0 and v1.0.0-RC1 (all fail with the same error)

PR #212, opened 2021-10-06
Update opentelemetry-go to v1.0.1 (trace) and v0.24.0 (metrics)

dashpole:
No changes to trace-related things, but plenty of naming changes to metrics instruments (e.g. ValueObserver -> GaugeObserver) and there is now a split between the "controller" and the "processor", which means there are now two nested `ForEach` loops.

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/210


exporter/metric/metric.go
```diff
@@ -255,16 +261,19 @@ func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md
...
 	tss := []*monitoringpb.TimeSeries{}
 
-	aggError := cps.ForEach(export.CumulativeExportKindSelector(), func(r export.Record) error {
-		ts, err := me.recordToTspb(&r)
-		if err != nil {
-			return err
-		}
-		tss = append(tss, ts)
-		return nil
+	// TODO do something with the Library?
```
punya, line 267: Consider opening issues to track these TODOs (and mentioning the issue in the TODO).

PR #220, opened 2021-10-19
[Collector e2e] Use json fixtures and local grpc server for assertions

aabmass:
- Update integration tests to load OTLP input, metric descriptor, and time series requests from JSON fixture files.
- Use a fake gRPC server to intercept requests to GCM. Compare these against expectation fixtures with `protocmp`
  - Ignore some fields when testing expectations, such as project ID (`name`). More ignores may be easily added if they cause flakes
  - Only write to GCM after the expectation part passes

Added JSON fixtures for a basic counter and the tests are passing 🙂

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ integrationtest:
 	set -e; for dir in $(ALL_GO_MOD_DIRS); do \
 	  echo "go test ./... + race in $${dir}"; \
 	  (cd "$${dir}" && \
-	    EXPORT_TO_GCP=true $(GOTEST) -tags=integrationtest -run=TestIntegration ./...); \
+	    $(GOTEST) -tags=integrationtest -run=TestIntegration ./...); \
 	done
 
 .PHONY: test-386
--- a/exporter/collector/metrics_integration_test.go
+++ b/exporter/collector/metrics_integration_test.go
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build integrationtest
+// +build integrationtest
+
 package collector_test
 
 import (
 	"context"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


exporter/collector/internal/integrationtest/testcase.go
```diff
@@ -0,0 +1,137 @@
...
+	t *testing.T,
+	startTime time.Time,
+	endTime time.Time,
+) pdata.Metrics {
+	bytes, err := ioutil.ReadFile(m.OTLPInputFixturePath)
+	require.NoError(t, err)
+	metrics, err := otlp.NewJSONMetricsUnmarshaler().UnmarshalMetrics(bytes)
+	require.NoError(t, err)
+
+	for i := 0; i < metrics.ResourceMetrics().Len(); i++ {
```
aabmass, line 69: Is there a more elegant way to update the timestamps?

dashpole:
My rule of thumb is: If it is "easy" to run, it should be triggered by `make test`.  If it is "hard" to run, it should have a special make target.  If you want to make sure the mock test runs before the test against the real cloud monitoring, you can just make the `make integrationtest` target run the `test` target before it runs.  The simplest way might be to have two different tests that share the core test function library, but have different files (with different build rules) to perform setup differently.

exporter/collector/metrics_integration_test.go
```diff
@@ -96,56 +93,72 @@ func TestIntegrationMetrics(t *testing.T) {
...
+				)
+				require.Emptyf(
+					t,
+					diff,
+					"Expected CreateTimeSeries request and actual GCM request differ:\n%v",
+					diff,
+				)
+			})
+
+			t.Run("Real Export", func(t *testing.T) {
```
dashpole, line 146: I think you could split this out into a separate file with the integration build rule like you had before if you wanted, rather than use env vars.

Makefile
```diff
@@ -89,7 +89,7 @@ integrationtest:
 	set -e; for dir in $(ALL_GO_MOD_DIRS); do \
 	  echo "go test ./... + race in $${dir}"; \
 	  (cd "$${dir}" && \
-	    $(GOTEST) -tags=integrationtest -run=TestIntegration ./...); \
+	    EXPORT_TO_GCP=true $(GOTEST) -tags=integrationtest -run=TestIntegration ./...); \
```
dashpole, line 92: if you wanted to ensure `test` passes before running `integrationtest`, I think you could do (above): `integrationtest: test`.

dashpole (approved):
This works for me.  Comments are optional

Makefile
aabmass: I'll leave like this for now since CircleCI is already covering that. It'll keep the build time down for the Cloud Build

exporter/collector/metrics_integration_test.go
aabmass: I also don't have a strong opinion. What do you prefer?

aabmass: Split it into two files, `metrics_test.go` and `metrics_integration_test.go` with build tag

dashpole (approved):
Looks great!

Issue #218, opened 2021-10-18
Support non-string Attributes in the metrics exporter


ronilichtman:
When sending a non-string Attribute (for example an Int Attribute) the exporter sends null/empty string to GCP.

See attached image.

Attributes added as follows:

```golang
errorCodeKey = attribute.Key("error_code")
errCode := 123
counterMetric.Add(ctx, 1, errorCodeKey.Int(errCode))
```

![image](https://user-images.githubusercontent.com/12766460/137716640-7cacb073-498f-45b6-804e-af99c948fd00.png)

PR #228, opened 2021-11-03
Add support for non-string metric labels

ronilichtman:
This is a follow-up to #217 which added support for monitored resource labels. In this PR we add support for metric labels.

Context: #218

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -498,7 +498,7 @@ func (me *metricExporter) recordToMpb(r *export.Record, library instrumentation.
 	iter := r.Labels().Iter()
 	for iter.Next() {
 		kv := iter.Label()
-		labels[normalizeLabelKey(string(kv.Key))] = kv.Value.AsString()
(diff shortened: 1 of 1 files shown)


dashpole:
Would you mind adding a unit test?

ronilichtman:
Hey David, I initially hoped to leverage the existing `TestRecordToMpb` test from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/metric/metric_test.go#L154, but couldn't get it to work.

Specifically, I changed the the strings that it expects (i.e. the "b_b" string to "b_b1234") and the test always passes essentially ignoring the `want` variable. I'm unsure how to integrate into that test, do you mind if I write a `TestRecordToMpb_NonStringAttributes` case which looks a bit different?

dashpole:
Feel free to write a new test that is specifically for this change.

dashpole:
i'll add the test myself in a follow-up.

dashpole (approved):

PR #230, opened 2021-11-08
Add ops-agent metric tests to integration tests for the exporter.

jsuereth:
Ensure existing behavior for ops-agent self + host metrics remains as-is during rewrite.

cc @lingshi, @aabmass

### DIFF
--- a/exporter/collector/testdata/fixtures/ops_agent_self_metrics_expect.json
+++ b/exporter/collector/testdata/fixtures/ops_agent_self_metrics_expect.json
@@ -1,0 +1,60 @@
+{
+  "createTimeSeriesRequests": [
+    {
+      "timeSeries": [
+        {
+          "metric": {
+            "type": "agent.googleapis.com/agent/memory_usage"
+          },
+          "resource": {
+            "type": "gce_instance",
+            "labels": {
+              "instance_id": "1447596667076884700",
+              "zone": "us-central1-a"
+            }
+          },
+          "metricKind": "GAUGE",
+          "valueType": "DOUBLE",
+          "points": [
+            {
+              "interval": {
+                "endTime": "1970-01-01T00:00:00Z"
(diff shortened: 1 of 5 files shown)


aabmass:
I'll look into the test failure, looks like something small

exporter/collector/testdata/fixtures/ops_agent_self_metrics_expect.json
```diff
@@ -0,0 +1,60 @@
...
+    {
+      "timeSeries": [
+        {
+          "metric": {
+            "type": "agent.googleapis.com/agent/memory_usage"
+          },
+          "resource": {
+            "type": "gce_instance",
+            "labels": {
+              "instance_id": "1447596667076884700",
```
aabmass, line 12: All of the generated resources are missing the `project_id` label. I'm guessing the OpenCensus exporter adds this label from the configured/detected credentials on the collector (and maybe you ran the script on your dev laptop) instead of from the metric's resource. This seems like a bug to me in that exporter?

jsuereth: I did run the script from my laptop, and you're right.   Should I add that back in manually?

aabmass: Interesting, so tests are passing now because

1. Unit tests (which test the expectations) are running in CircleCI where there are no Application Default Credentials, so the exporter doesn't set `project_id` and it matches up anyway. This will break test runs on cloudtop/GCE VMs, but I can fix it in a follow up PR.
2. Integration test are passing because the exporter uses the configured/default project ID regardless.

> Should I add that back in manually?

No idts. I say we merge this as is :)

aabmass (approved):
Pretty cool, LGTM!

PR #232, opened 2021-11-22
Fix bug where test gRPC server channels were buffering

aabmass:
There was a bug popping up occasionally where the fake gRPC server that captures requests for integration tests was missing requests. The cause was that the receiving side would occasionally hit the `default` case before the sending goroutines had a chance to run.

The fix is to use slices + mutex instead to ensure we capture all of the points before sending a gRPC response.

### DIFF
--- a/exporter/collector/internal/integrationtest/metric_test_server.go
+++ b/exporter/collector/internal/integrationtest/metric_test_server.go
@@ -92,7 +92,7 @@ func (f *fakeMetricServiceServer) CreateTimeSeries(
 	ctx context.Context,
 	req *monitoringpb.CreateTimeSeriesRequest,
 ) (*emptypb.Empty, error) {
-	go f.appendCreateTimeSeriesReq(req)
+	f.appendCreateTimeSeriesReq(req)
 	return &emptypb.Empty{}, nil
 }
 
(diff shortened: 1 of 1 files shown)


exporter/collector/internal/integrationtest/metric_test_server.go
```diff
@@ -78,23 +84,23 @@ func (m *MetricsTestServer) Serve() error {
...
-	createTimeSeriesChan       chan<- *monitoringpb.CreateTimeSeriesRequest
+	appendCreateMetricDescriptorReq func(*monitoringpb.CreateMetricDescriptorRequest)
+	appendCreateTimeSeriesReq       func(*monitoringpb.CreateTimeSeriesRequest)
 }
 
 func (f *fakeMetricServiceServer) CreateTimeSeries(
 	ctx context.Context,
 	req *monitoringpb.CreateTimeSeriesRequest,
 ) (*emptypb.Empty, error) {
-	go func() { f.createTimeSeriesChan <- req }()
```
aabmass, line 89: I think increasing the channel buffer size and making this run in the same goroutine would also fix the issue, but idk how big the buffer may need to be. If it's too small, this solution would deadlock

aabmass: Ah yes sorry, the slice append should be in the same goroutine, let me update.

If I leave it as a channel, it can be in the same goroutine but AFAIK it will block if the buffer is full and cause deadlock

dashpole: Yeah, I think this works much better.

dashpole (approved):

PR #242, opened 2021-12-13
[Metrics Rewrite] support for pdata Sum points

aabmass:
- Separated metric mapping portions into methods on struct `metricMapper`, which holds config and has all pure functions. This makes it easy to test without creating a GCM client
- Implement OTLP Sum point mapping in `sumPointToTimeSeries()`
- Tests

Note that in the OC based exporter, delta sums were converted into gauges ([test fixture](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/9bc1f49ebe000b0b3b1aa5b7f201e7996effdcd8/exporter/collector/testdata/fixtures/delta_counter_metrics_expect.json#L15)) while here we are making cumulatives with short time window. This will be a breaking change

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -19,7 +19,6 @@ package collector
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"unicode"
 
@@ -209,7 +208,7 @@ func (m *metricMapper) sumPointToTimeSeries(
 // metricNameToType maps OTLP metric name to GCM metric type (aka name)
 func (m *metricMapper) metricNameToType(name string) string {
 	// TODO
-	return fmt.Sprintf("workload.googleapis.com/%v", name)
+	return "workload.googleapis.com/" + name
 }
 
 func attributesToLabels(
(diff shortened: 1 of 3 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -19,20 +19,29 @@ package collector
...
 type metricsExporter struct {
 	cfg    *Config
 	client *monitoring.MetricClient
+	mapper metricMapper
+}
+
+// metricMapper is the part that transforms metrics. Separate from metricsExporter since it has
+// all pure functions.
+type metricMapper struct {
+	cfg *Config
```
dashpole, line 44: cfg doesn't appear to be used anywhere.

aabmass: You're right, i added it preemptively. It will effect some of the mapping logic like metric name. I can remove it from this PR until it is needed if you prefer

dashpole: no, thats fine.

dashpole (approved):

PR #243, opened 2021-12-14
[Metrics Rewrite] attribute to label mapping

aabmass:
Check out the test cases to see how special cases are handled. This uses pdata's [`AttributeValue.AsString()`](https://pkg.go.dev/go.opentelemetry.io/collector/model/pdata#AttributeValue.AsString) which appears to convert to string using JSON rules. Differences from [OC based exporter](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/42e7e58efdb937e8477f827d3fba022212335dbc/sanitize.go#L26):

- This code does not truncate attrib keys longer than 100 characters
- This code does not prepend `key` to keys starting with `_`.

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -19,6 +19,8 @@ package collector
 
 import (
 	"context"
+	"strings"
+	"unicode"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
 	"go.opentelemetry.io/collector/component"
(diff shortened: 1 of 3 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -173,7 +175,35 @@ func attributesToLabels(
...
+	ls := make(labels, attrs.Len())
+	attrs.Range(func(k string, v pdata.AttributeValue) bool {
+		ls[sanitizeKey(k)] = v.AsString()
+		return true
+	})
+	return ls
+}
+
+// Replaces non-alphanumeric characters to underscores. Note, this does not truncate label keys
+// longer than 100 characters or prepend "key" when the first character is "_" like OpenCensus
```
dashpole, line 187: Should we start a list somewhere of the changes between OC and otel exporters?

dashpole (approved):

dashpole:
Side point: it would be neat to have the exporter specification in OSS so we can reference it in comments in implementations.

exporter/collector/metricsexporter.go
aabmass: Yup, let me add some markdown

aabmass: Added `exporter/collector/breaking-changes.md`

PR #245, opened 2021-12-15
Add workload-metrics test fixture

dashpole:
Does it matter that it isn't using the workload.googleapis.com prefix?  It sounds like we may introduce special behavior for that prefix in the future, so it might be nice to be able to set config for a fixture.


aabmass:
> Does it matter that it isn't using the workload.googleapis.com prefix?

If we duplicate one of the fixture files and explicitly prepend the OTLP metric names with `workload.googleapis.com/`, we could cover both right?

aabmass:
Test flaked (probably because of metric descriptor propagation behind the scenes, should be a one off thing), restarted it

aabmass (approved):
LGTM thanks!

dashpole:
The fixture comes in without workload.googleapis.com, and the exporter is expected to prepend it.  Adding it to the incoming and outgoing test case won't exercise the exporter, so I don't think that would be what we want

Issue #225, opened 2021-10-29
Support CreateServiceTimeSeries


dashpole:
Similarly to https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/294, we should support CreateServiceTimeSeries for service metrics.  This needs to happen before the collector switches to use this repo.

jsuereth:
Fixed by #259

PR #247, opened 2021-12-20
Add logic to translate metric descriptors and initial flow

jsuereth:
- If you enable any of the end-to-end integration tests, you'll see metric descriptors sent to the dummy service.
- Adds "known domains" configuration, so in the event we add new metric domain types (or system metrics define them), the metric name mapping logic can be configured.
- Adds "CreateDefaults" method to metric config structure.  This allows us to write method which use configuration without worrying about nil checking repeatedly.  Note: There's probably a better "go" way to do this, let me know.
- Updates createTimeSeries to call `CreateTimeSeries`.  We'll need to figure out `CreateServiceTimeSeries` later.
- Adds metric name/type/display name mappings (for updated version).
- Add simple label-mapping (no label-descriptions possible)
- Add constants for Summary mapping.

Not in this PR:
- Add "legacy" flag for metric naming conventions that:
  - uses external.googleapis.com/OpenCensus/
  - sets display name to original metric name (or last part of the path).

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -30,7 +30,6 @@ import (
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 	"go.opentelemetry.io/collector/model/pdata"
 	"google.golang.org/genproto/googleapis/api/label"
-	"google.golang.org/genproto/googleapis/api/metric"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
@@ -43,7 +42,7 @@ type metricsExporter struct {
 	client *monitoring.MetricClient
 	mapper metricMapper
 	// A channel that receives metric descriptor and sends them to GCM once.
-	mds chan *metric.MetricDescriptor
+	mds chan *metricpb.MetricDescriptor
 }
 
 // metricMapper is the part that transforms metrics. Separate from metricsExporter since it has
@@ -71,7 +70,7 @@ func (me *metricsExporter) Shutdown(context.Context) error {
 
 // Updates config object to include all defaults for metric export.
 func (cfg *Config) SetMetricDefaults() {
-	if cfg.MetricConfig.KnownDomains == nil || len(cfg.MetricConfig.KnownDomains) == 0 {
+	if len(cfg.MetricConfig.KnownDomains) == 0 {
 		cfg.MetricConfig.KnownDomains = domains
 	}
 	// TODO - if in legacy mode, this should be
@@ -113,7 +112,7 @@ func newGoogleCloudMetricsExporter(
 		cfg:    cfg,
 		client: client,
 		mapper: metricMapper{cfg},
-		mds:    make(chan *metric.MetricDescriptor),
+		mds:    make(chan *metricpb.MetricDescriptor),
 	}
 
 	// Fire up the metric descriptor exporter.
@@ -160,7 +159,7 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
 
 // Reads metric descriptors from the md channel, and reports them (once) to GCM.
 func (me *metricsExporter) exportMetricDescriptorRunner(ctx context.Context) {
-	mdCache := make(map[string]*metric.MetricDescriptor)
+	mdCache := make(map[string]*metricpb.MetricDescriptor)
 	for {
 		select {
 		case md := <-me.mds:
@@ -181,7 +180,7 @@ func (me *metricsExporter) exportMetricDescriptorRunner(ctx context.Context) {
 }
 
 // Helper method to send metric descriptors to GCM.
-func (me *metricsExporter) exportMetricDescriptor(ctx context.Context, md *metric.MetricDescriptor) error {
+func (me *metricsExporter) exportMetricDescriptor(ctx context.Context, md *metricpb.MetricDescriptor) error {
 	// export
 	req := &monitoringpb.CreateMetricDescriptorRequest{
 		// TODO set Name field with project ID from config or ADC
@@ -469,17 +468,17 @@ func (m *metricMapper) labelDescriptors(pm pdata.Metric) []*label.LabelDescripto
 	return result
 }
 
-func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metric.MetricDescriptor {
+func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metricpb.MetricDescriptor {
 	sumName := fmt.Sprintf("%s%s", pm.Name(), SummarySumSuffix)
 	countName := fmt.Sprintf("%s%s", pm.Name(), SummaryCountPrefix)
 	percentileName := fmt.Sprintf("%s%s", pm.Name(), SummaryPercentileSuffix)
 	labels := m.labelDescriptors(pm)
-	return []*metric.MetricDescriptor{
+	return []*metricpb.MetricDescriptor{
 		{
 			Type:        m.metricNameToType(sumName),
 			Labels:      labels,
-			MetricKind:  metric.MetricDescriptor_CUMULATIVE,
-			ValueType:   metric.MetricDescriptor_DOUBLE,
+			MetricKind:  metricpb.MetricDescriptor_CUMULATIVE,
+			ValueType:   metricpb.MetricDescriptor_DOUBLE,
 			Unit:        pm.Unit(),
 			Description: pm.Description(),
 			DisplayName: sumName,
@@ -487,8 +486,8 @@ func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metric.Metri
 		{
 			Type:        m.metricNameToType(countName),
 			Labels:      labels,
-			MetricKind:  metric.MetricDescriptor_CUMULATIVE,
-			ValueType:   metric.MetricDescriptor_INT64,
+			MetricKind:  metricpb.MetricDescriptor_CUMULATIVE,
+			ValueType:   metricpb.MetricDescriptor_INT64,
 			Unit:        pm.Unit(),
 			Description: pm.Description(),
 			DisplayName: countName,
@@ -501,8 +500,8 @@ func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metric.Metri
 					Key:         "percentile",
 					Description: "the value at a given percentile of a distribution",
 				}),
-			MetricKind:  metric.MetricDescriptor_GAUGE,
-			ValueType:   metric.MetricDescriptor_DOUBLE,
+			MetricKind:  metricpb.MetricDescriptor_GAUGE,
+			ValueType:   metricpb.MetricDescriptor_DOUBLE,
 			Unit:        pm.Unit(),
 			Description: pm.Description(),
 			DisplayName: percentileName,
@@ -511,7 +510,7 @@ func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metric.Metri
 }
 
 // Extract the metric descriptor from a metric data point.
-func (m *metricMapper) metricDescriptor(pm pdata.Metric) []*metric.MetricDescriptor {
+func (m *metricMapper) metricDescriptor(pm pdata.Metric) []*metricpb.MetricDescriptor {
 	if pm.DataType() == pdata.MetricDataTypeSummary {
 		return m.summaryMetricDescriptors(pm)
 	}
@@ -519,10 +518,10 @@ func (m *metricMapper) metricDescriptor(pm pdata.Metric) []*metric.MetricDescrip
 	metricType := m.metricNameToType(pm.Name())
 	labels := m.labelDescriptors(pm)
 	// Return nil for unsupported types.
-	if kind == metric.MetricDescriptor_METRIC_KIND_UNSPECIFIED {
+	if kind == metricpb.MetricDescriptor_METRIC_KIND_UNSPECIFIED {
 		return nil
 	}
-	return []*metric.MetricDescriptor{
+	return []*metricpb.MetricDescriptor{
 		{
 			Name:        pm.Name(),
 			DisplayName: m.metricTypeToDisplayName(metricType),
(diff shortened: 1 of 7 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -230,10 +316,24 @@ func (m *metricMapper) gaugePointToTimeSeries(
 	}
 }
 
+// Returns any configured prefix to add to unknown metric name.
+func (m *metricMapper) getMetricNamePrefix(name string) *string {
```
aabmass, line 320: Can you add unit test for this with some known_domains?

exporter/collector/metricsexporter.go
```diff
@@ -298,3 +398,197 @@ func mergeLabels(mergeInto labels, others ...labels) labels {
...
+func (m *metricMapper) labelDescriptors(pm pdata.Metric) []*label.LabelDescriptor {
+	// TODO - allow customization of label descriptions.
+	result := []*label.LabelDescriptor{}
+	switch pm.DataType() {
+	case pdata.MetricDataTypeGauge:
+		points := pm.Gauge().DataPoints()
+		for i := 0; i < points.Len(); i++ {
+			points.At(i).Attributes().Range(func(key string, _ pdata.AttributeValue) bool {
+				result = append(result, &label.LabelDescriptor{
+					Key: key,
```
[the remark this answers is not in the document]
jsuereth: If you don't include these, it deletes them from all timeseries when calling CMD on non-workload-metrics domains.

exporter/collector/metricsexporter.go
```diff
@@ -108,12 +158,48 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
+			// Not yet sent, now we sent it.
+			if !me.cfg.MetricConfig.SkipCreateMetricDescriptor && md != nil && mdCache[md.Type] == nil {
+				err := me.exportMetricDescriptor(ctx, md)
+				// TODO: Log-once on error, per metric descriptor?
+				if err != nil {
+					fmt.Printf("Unable to send metric descriptor: %s, %s", md, err)
+				}
+				mdCache[md.Type] = md
+			}
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
jsuereth: Cleaner, but different.   The question here is if we should try to export metric descriptors AFTER being closed (block the channel until all pending descriptors have been pulled).  IIUC, select can pick from first avialable so we can early-terminate.  Does that work with close + range?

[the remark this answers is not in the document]
jsuereth: Also, as an FYI - reading this https://go101.org/article/channel-closing.html makes my head spin.

1. If we assume a SINGLE SENDER thread per-exporter, range+close should be ok
2. I had hoped (at some point) to share MetricDescriptor cache *across* metric exporter instances, but realize that' wayyyy too insane.

I'll move to range, but posit that I think channels are rather complicated if you care about failure modes and early termination.

exporter/collector/metricsexporter.go
```diff
@@ -108,12 +158,48 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
 	return err
 }
 
+// Reads metric descriptors from the md channel, and reports them (once) to GCM.
+func (me *metricsExporter) exportMetricDescriptorRunner(ctx context.Context) {
+	mdCache := make(map[string]*metric.MetricDescriptor)
```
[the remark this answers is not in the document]
jsuereth: I do plan to use it (I think) added the TODO.

exporter/collector/metricsexporter.go
```diff
@@ -69,8 +113,12 @@ func newGoogleCloudMetricsExporter(
 		cfg:    cfg,
 		client: client,
 		mapper: metricMapper{cfg},
+		mds:    make(chan *metric.MetricDescriptor),
 	}
 
+	// Fire up the metric descriptor exporter.
+	go mExp.exportMetricDescriptorRunner(ctx)
```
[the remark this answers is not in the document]
jsuereth: Fixed to use channel closure.

exporter/collector/metricsexporter.go
```diff
@@ -230,10 +316,24 @@ func (m *metricMapper) gaugePointToTimeSeries(
...
+}
+
 // metricNameToType maps OTLP metric name to GCM metric type (aka name)
 func (m *metricMapper) metricNameToType(name string) string {
-	// TODO
-	return "workload.googleapis.com/" + name
+	prefix := m.getMetricNamePrefix(name)
+	if prefix != nil {
+		return path.Join(*prefix, name)
+	}
```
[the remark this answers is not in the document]
jsuereth: Copied from OC exporter, but also I think that behavior is "ok".  WDYT?

exporter/collector/metricsexporter.go
```diff
@@ -45,22 +52,59 @@ type metricMapper struct {
...
+
 type labels map[string]string
 
 func (me *metricsExporter) Shutdown(context.Context) error {
+	// TODO - will the context given to us at startup be cancelled already?
 	return me.client.Close()
 }
 
+// Updates config object to include all defaults for metric export.
+func (cfg *Config) SetMetricDefaults() {
```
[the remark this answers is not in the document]
jsuereth: So I wasn't sure how to go from `config.Exporter` back to `Config` PTAL

exporter/collector/metricsexporter.go
```diff
@@ -19,13 +19,18 @@ package collector
...
+	"path"
 	"strings"
 	"unicode"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 	"go.opentelemetry.io/collector/model/pdata"
+	"google.golang.org/genproto/googleapis/api/label"
+	"google.golang.org/genproto/googleapis/api/metric"
```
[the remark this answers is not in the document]
jsuereth: Fixed.

exporter/collector/metricsexporter.go
jsuereth: done.

exporter/collector/metricsexporter.go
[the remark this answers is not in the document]
aabmass: Haven't heard any complaints from OC around this

exporter/collector/metricsexporter.go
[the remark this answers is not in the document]
aabmass: I think what you did is fine.

The other option is to leave the createDefaultConfig returning a `config.Exporter` and type assert `metricMapper{cfg: createDefaultConfig().(*Config)}` in the tests.

exporter/collector/metricsexporter.go
```diff
@@ -69,8 +91,12 @@ func newGoogleCloudMetricsExporter(
 		cfg:    cfg,
 		client: client,
 		mapper: metricMapper{cfg},
+		mds:    make(chan *metricpb.MetricDescriptor),
```
aabmass, line 94: Do we want any buffer for this channel? As is, one slow CreateMetricDescriptor call will block `me.mds <- md`

exporter/collector/metricsexporter.go
```diff
@@ -108,12 +136,46 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
+
 func (me *metricsExporter) createTimeSeries(ctx context.Context, ts []*monitoringpb.TimeSeries) error {
-	return me.client.CreateServiceTimeSeries(
+	// TODO: me.client.CreateServiceTimeSeries(
+	return me.client.CreateTimeSeries(
 		ctx,
 		&monitoringpb.CreateTimeSeriesRequest{
 			TimeSeries: ts,
 			// TODO set Name field with project ID from config or ADC
+			Name: fmt.Sprintf("projects/%s", me.cfg.ProjectID),
```
aabmass, line 178: nit can share this code with L164

exporter/collector/metricsexporter.go
```diff
@@ -298,3 +369,197 @@ func mergeLabels(mergeInto labels, others ...labels) labels {
...
+		}
+	case pdata.MetricDataTypeSum:
+		points := pm.Sum().DataPoints()
+		for i := 0; i < points.Len(); i++ {
+			points.At(i).Attributes().Range(func(key string, _ pdata.AttributeValue) bool {
+				result = append(result, &label.LabelDescriptor{
+					Key: key,
+				})
+				return true
+			})
```
aabmass, line 407: nit could add a closure func which takes `pdata.AttributeMap` and adds the label descriptors to `result` to reduce some of this duplication. Don't think it's possible to get rid of the switch/for loops though :(

aabmass (approved):
nits but LGTM

exporter/collector/metricsexporter.go
jsuereth: In my mind CMD is optimistic.  I want to give it the minimal amount of resource and I don't care if any specific call fails.  The current state of it, we basically *have* to include logic around it, but it's dubious whether we want it in the long run.

I'm happy having it on its own "thread" churning away slowly.

aabmass:
@jsuereth can you update https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/col-exporter-rewrite/exporter/collector/breaking-changes.md if there are any breaking changes vs the old exporter?

exporter/collector/metricsexporter.go
jsuereth: This was, quite honestly, the most painful thing to write in Go so far, knowing that abstraction COULD have occurred.

That cleaned up the worst of it, still wish there was an abstract way to pull attributes from subpoints without knowing specific structures.

exporter/collector/metricsexporter.go
```diff
@@ -97,23 +123,82 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
+					me.mds <- md
+				}
 				timeSeries = append(timeSeries, me.mapper.metricToTimeSeries(monitoredResource, extraLabels, metric)...)
 			}
 		}
 	}
 
 	// TODO: self observability
+	// TODO: Figure out how to configure service time series calls.
+	if false {
```
jsuereth, line 136: @aabmass @dashpole I added this as a reminder that I don't think we have an open bug to support service timeseries calls in this rework.   Would one of you mind confirming/opening approrpiately? I think David has the most context on what we need here.

exporter/collector/metricsexporter.go
```diff
@@ -97,23 +123,82 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
 
+// Reads metric descriptors from the md channel, and reports them (once) to GCM.
+func (me *metricsExporter) exportMetricDescriptorRunner() {
+	mdCache := make(map[string]*metricpb.MetricDescriptor)
+	// We iterate over all metric descritpors until the channel is closed.
+	// Note: if we get terminated, this will still attempt to export all descriptors
+	// prior to shutdown.
+	for md := range me.mds {
+		// Not yet sent, now we sent it.
+		// TODO - check to see if this is a service/system metric and doesn't send descriptors.
```
jsuereth, line 154: FYI - OpenCensus had logic to NOT send metric descriptors on certain domains when we suspect they are "system" (or service) metrics.   I can add that in this CL or a follow on.  It's related to the design of how we want to handle CreateServiceTimeSeries vs. CreateTimeSeries calls.

cc @dashpole

exporter/collector/metricsexporter.go
dashpole: You can link to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/225.

exporter/collector/metricsexporter.go
dashpole: That makes sense.  It would be nice if CreateServiceTimeSeries == don't create MD so we can simplify the implementation

dashpole: I'd rather not implement that logic if we don't have to

exporter/collector/metricsexporter.go
```diff
@@ -58,9 +76,14 @@ func newGoogleCloudMetricsExporter(
 ) (component.MetricsExporter, error) {
 	setVersionInUserAgent(cfg, set.BuildInfo.Version)
 
-	// TODO: map cfg options into metric service client configuration with
+	// map cfg options into metric service client configuration with
```
dashpole, line 79: nit: this comment can probably be removed entirely.

exporter/collector/metricsexporter.go
dashpole: You can probably just copy https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/294/files#diff-9f6cddeabc26d57e3837fe66df90a6288f56e8a0e898cb6004d098da521583aeR656 and use it here.

exporter/collector/metricsexporter.go
dashpole: To rephrase aaron's point, the current implementation isn't really better than having the CMD call serialized with the CreateTimeSeries call.  Either way, a CreateMetricDescriptor call can block the CreateTimeSeries calls.  If you want CMD to be optimistic, you could have a buffered channel, but drop CMD calls if the buffer fills up.  Otherwise, you probably don't need to bother with the extra goroutine.

aabmass: That makes sense, but without a buffer (or non-blocking write), L126 in pushMetrics() will block until the background goroutine reads from the channel

exporter/collector/metricsexporter.go
aabmass: I assigned the bug to @jsuereth and pulled into the sprint

exporter/collector/metricsexporter.go
```diff
@@ -97,23 +123,82 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
+		// Not yet sent, now we sent it.
+		// TODO - check to see if this is a service/system metric and doesn't send descriptors.
+		if !me.cfg.MetricConfig.SkipCreateMetricDescriptor && md != nil && mdCache[md.Type] == nil {
+			err := me.exportMetricDescriptor(context.Background(), md)
+			// TODO: Log-once on error, per metric descriptor?
+			// TODO: Re-use passed-in logger to exporter.
+			if err != nil {
+				log.Printf("Unable to send metric descriptor: %s, %s", md, err)
+			}
+			mdCache[md.Type] = md
```
aabmass, line 162: If you intend to retry the CMD next time around if there was an error the first go, there is a bug here

```suggestion
			if err != nil {
				log.Printf("Unable to send metric descriptor: %s, %s", md, err)
			} else {
				mdCache[md.Type] = md
			}
```

exporter/collector/metricsexporter.go
```diff
@@ -97,23 +123,82 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
+func (me *metricsExporter) exportMetricDescriptorRunner() {
+	mdCache := make(map[string]*metricpb.MetricDescriptor)
+	// We iterate over all metric descritpors until the channel is closed.
+	// Note: if we get terminated, this will still attempt to export all descriptors
+	// prior to shutdown.
+	for md := range me.mds {
+		// Not yet sent, now we sent it.
+		// TODO - check to see if this is a service/system metric and doesn't send descriptors.
+		if !me.cfg.MetricConfig.SkipCreateMetricDescriptor && md != nil && mdCache[md.Type] == nil {
+			err := me.exportMetricDescriptor(context.Background(), md)
```
aabmass, line 156: You probably want a timeout for this ctx, or is this handled automatically by the client library?

jsuereth: I believe we should be able to configure that on gRPC clients and it'll attach to an existing timeout if one exists, or create one.   However, I'm rather weak at Go, as you know :)

exporter/collector/metricsexporter.go
jsuereth: Good catch!

exporter/collector/metricsexporter.go
jsuereth: sounds good, I'll work on that next.

exporter/collector/metricsexporter.go
jsuereth: Decided to go with a buffer of a few, and I do some pre-filtering of MDs before shoveling in the buffer.   I think what we have now matches requirements.

exporter/collector/metricsexporter.go
aabmass: I wasn't sure so looked around: https://pkg.go.dev/cloud.google.com/go#hdr-Timeouts_and_Cancellation TLDR; the client lib will set a default timeout if the context doesn't have one already.

@dashpole probably knows better, is the default reasonable for this or should we hardcode something?

exporter/collector/metricsexporter.go
aabmass: LGTM

exporter/collector/metricsexporter.go
dashpole: I think there is a timeout in the googlecloud exporter's config.  I'd expect that to apply to the CMD call as well.  Otherwise, i'd expect it to use the default timeout

dashpole: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ebea5bc82e5a146db11d4b496a7a62e2151fe89a/exporter/googlecloudexporter/googlecloud.go#L141

aabmass: Discussed in meeting, would be better to send the context in the channel from the other goroutine.

PR #248, opened 2021-12-20
Add gke metrics agent fixtures

dashpole:
This includes the gke metrics agent daemonset and control-plane metrics agent.  It fixes an issue in the metrics test server which caused generated fixtures to be empty.


jsuereth:
@dashpole Does the GKE-control-plane metric agent use the new CreateSystemTimeSeries call?  I think the code that captures test cases may need to hijack that call

dashpole:
Found the issue.  @jsuereth it does use CreateSystemTimeSeries, but there was an issue in the test server implementation for that call.  I fixed it, and it now generates the fixture correctly

dashpole:
the control plane metrics sets a custom resource mapping, so I had to change the opencensus.resourcetype from "k8s" to "host", which should make them sent to k8s_node instead of global.

aabmass (approved):

PR #251, opened 2021-12-21
Add support for exponential histograms and exemplars.

jsuereth:
- Adds new exponential histogram mapping and unit test.
- Adds mapping for Exemplars (with specific attachment protos) and unit tests.

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -185,7 +185,9 @@ func (m *metricMapper) exemplar(ex pdata.Exemplar) *distribution.Distribution_Ex
 		if err == nil {
 			attachments = append(attachments, sctx)
 		} else {
-			log.Printf("Failure to write span context: %v", err)
+			// This happens in the event of logic error (e.g. missing required fields).
+			// As such we complaining loudly to fail our unit tests.
+			log.Fatalf("Failure to write span context: %v", err)
 		}
 	}
 	if ex.FilteredAttributes().Len() > 0 {
@@ -195,7 +197,9 @@ func (m *metricMapper) exemplar(ex pdata.Exemplar) *distribution.Distribution_Ex
 		if err == nil {
 			attachments = append(attachments, attr)
 		} else {
-			log.Printf("Failure to write filtered_attributes: %v", err)
+			// This happens in the event of logic error (e.g. missing required fields).
+			// As such we complaining loudly to fail our unit tests.
+			log.Fatalf("Failure to write filtered_attributes: %v", err)
 		}
 	}
 	return &distribution.Distribution_Exemplar{
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -19,6 +19,9 @@ package collector
 
 import (
 	"context"
+	"fmt"
+	"log"
+	"math"
 	"strings"
 	"unicode"
 
@@ -26,9 +29,11 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 	"go.opentelemetry.io/collector/model/pdata"
+	"google.golang.org/genproto/googleapis/api/distribution"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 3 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
...
+	// TODO: Look into still sending exemplars with no span.
+	if !ex.TraceID().IsEmpty() && !ex.SpanID().IsEmpty() {
+		sctx, err := anypb.New(&monitoringpb.SpanContext{
+			// TODO - make sure project id is correct.
+			SpanName: fmt.Sprintf("projects/%s/traces/%s/spans/%s", m.cfg.ProjectID, ex.TraceID().HexString(), ex.SpanID().HexString()),
+		})
+		if err == nil {
+			attachments = append(attachments, sctx)
+		} else {
+			log.Printf("Failure to write span context: %v", err)
```
dashpole, line 188: If we are going to log within the exporter, we should use the `params.Logger` provided to us when the exporter as part of `newGoogleCloudMetricsExporter`.  But I think we should consider dropping the metric if we can't write the exemplar so the failures can be detected using the self-obs metrics (dropped/sent/failed send).

Same comment for error handling below.

exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
...
+	counts := []int64{int64(underflow)}
+	for _, v := range point.Positive().BucketCounts() {
+		counts = append(counts, int64(v))
+	}
+
+	// Overflow bucket is always empty
+	counts = append(counts, 0)
+
+	// TODO - error handling?
+	if len(counts) < 3 {
```
aabmass, line 229: Should we try to hack a distribution out of this? We could send an ExplicitBuckets distribution with `bounds = [0.0]` and `bucket_counts = [underflow, 0]`

exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
...
+func (m *metricMapper) exponentialHistogramToTimeSeries(
+	resource *monitoredrespb.MonitoredResource,
+	extraLabels labels,
+	metric pdata.Metric,
+	sum pdata.ExponentialHistogram,
+	point pdata.ExponentialHistogramDataPoint,
+) *monitoringpb.TimeSeries {
+	// We treat deltas as cumulatives w/ resets.
+	metricKind := metricpb.MetricDescriptor_CUMULATIVE
+	startTime := timestamppb.New(point.StartTimestamp().AsTime())
```
aabmass, line 268: nit move this to 277 for consistency with `EndTime`?

exporter/collector/metricsexporter.go
jsuereth: Two questions here that make me disagree with dropping the point:

1. If my exemplars are broken do I also want my histograms to stop showing up?
2. What are the likely failures where this `err` would be returned? (i.e. failure to locally serialize a proto into an Any).

I agree on logger.

exporter/collector/metricsexporter_test.go
```diff
@@ -93,6 +93,129 @@ func TestMergeLabels(t *testing.T) {
...
+	assert.Len(t, hdp.Exemplars, 1)
+	ex := hdp.Exemplars[0]
+	assert.Equal(t, float64(2), ex.Value)
+	assert.Equal(t, timestamppb.New(end), ex.Timestamp)
+	// We should see trace + dropped labels
+	assert.Len(t, ex.Attachments, 2)
+	spanctx := &monitoringpb.SpanContext{}
+	err := ex.Attachments[0].UnmarshalTo(spanctx)
+	assert.Nil(t, err)
+	assert.Equal(t, "projects/myproject/traces/00010203040506070809010203040506/spans/0001020304050607", spanctx.SpanName)
```
aabmass, line 157: I don't really remember, should these be zero-padded?

exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
 	return timeSeries
 }
 
+func (m *metricMapper) exemplar(ex pdata.Exemplar) *distribution.Distribution_Exemplar {
```
aabmass, line 177: cc @punya this should make your life easier for the explicit distribution

exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
...
+	growth := math.Exp2(math.Exp2(-float64(point.Scale())))
+	scale := math.Pow(growth, float64(point.Positive().Offset()))
+	// First calculate underflow bucket with all negatives + zeros.
+	underflow := point.ZeroCount()
+	for _, v := range point.Negative().BucketCounts() {
+		underflow += v
+	}
+
+	// Next, pull in remaining buckets.
+	counts := []int64{int64(underflow)}
```
aabmass, line 220: nit you can pre-allocate with `make([]int64, len[, cap])`. There should be `len(point.Positive().BucketCounts()) + 2` counts total

exporter/collector/metricsexporter.go
```diff
@@ -162,6 +174,121 @@ func (m *metricMapper) metricToTimeSeries(
...
+
+	// Overflow bucket is always empty
+	counts = append(counts, 0)
+
+	// TODO - error handling?
+	if len(counts) < 3 {
+		// Cannot report an exponential distribution with no buckets.
+		return nil
+	}
+	exemplars := []*distribution.Distribution_Exemplar{}
```
aabmass, line 233: nit can pre-alloc

aabmass (approved):
Looks great! Just some nits

exporter/collector/metricsexporter.go
dashpole: I'm not entirely sure about 2.  My hope would be "only if something went terribly wrong", but i'm just guessing.  1 is a good question, but I generally lean towards failing loudly.  Would this be a debug-level log, since it is per-request?  Or would we make it a warning?  If the former, i'm not sure it is enough of a signal for someone to notice if exemplars are broken

exporter/collector/metricsexporter_test.go
jsuereth: Yes, they should.  I *think* this should happen by calling hexstring.

exporter/collector/metricsexporter.go
jsuereth: It's a neat idea. Will implement this way for now, and I'd like toe xperiment w/ Monarch to see what happens with this!

exporter/collector/metricsexporter.go
jsuereth: For 2 - having done SOME inspection, it seem like there are two primary ways this fails:

- If you pass a `nil` message
- If you have a required filed that's missing a value.

I think both are very rare in practice.

For 1 - I'm fine if we fail loudly where it's a logic error.  If it's the case of "we don't know the shape of user data", I think here it's better to give the user as MUCH of a working solution as we can (while loudly complaining).  Failing to write the point is different than loudly complaining.  I think we can accomplish both of our goals here with a different technique (you're right about debug logs).  Let's noodle on this for next time where it's more likely an error could happen at runtime post-unit-tests.

jsuereth: Regarding using params.Logger, going to open a new bug about that one, need some wiring/funnelling.

exporter/collector/metricsexporter.go
jsuereth: fixed

exporter/collector/metricsexporter.go
jsuereth: Fixed

exporter/collector/metricsexporter.go
dashpole: sgtm

aabmass (approved):

PR #252, opened 2021-12-22
[Metrics Rewrite] implement monitored resource mapping

aabmass:
Implements monitored resource mapping. Added in a new file and added a bunch of tests. Integration tests will come in a separate PR.

There are a lot of breaking changes here from the previous OC mapping, I need to update the markdown still.

### DIFF
--- a/exporter/collector/monitoredresource.go
+++ b/exporter/collector/monitoredresource.go
@@ -42,55 +42,73 @@ const (
 	podName        = "pod_name"
 	region         = "region"
 	taskID         = "task_id"
+	zone           = "zone"
 )
 
 var (
-	// monitoredResourceMappings contains mappings of GCM resource label keys onto OTel
-	// resource keys for a given monitored resource type. For entries with multiple OTel
-	// resource keys, the keys' values will be coalesced in order until there is a non-empty
-	// value.
-	monitoredResourceMappings = map[string]map[string][]string{
+	// monitoredResourceMappings contains mappings of GCM resource label keys onto mapping config from OTel
+	// resource for a given monitored resource type.
+	monitoredResourceMappings = map[string]map[string]struct {
+		// OTel resource keys to try and populate the resource label from. For entries with
+		// multiple OTel resource keys, the keys' values will be coalesced in order until there
+		// is a non-empty value.
+		otelKeys []string
+		// If none of the otelKeys are present in the Resource, fallback to this literal value
+		fallbackLiteral string
+	}{
 		gceInstance: {
-			location:   {semconv.AttributeCloudAvailabilityZone},
-			instanceID: {semconv.AttributeHostID},
+			zone:       {otelKeys: []string{semconv.AttributeCloudAvailabilityZone}},
+			instanceID: {otelKeys: []string{semconv.AttributeHostID}},
 		},
 		k8sContainer: {
-			location:      {semconv.AttributeCloudAvailabilityZone},
-			clusterName:   {semconv.AttributeK8SClusterName},
-			namespaceName: {semconv.AttributeK8SNamespaceName},
-			podName:       {semconv.AttributeK8SPodName},
-			containerName: {semconv.AttributeK8SContainerName},
+			location:      {otelKeys: []string{semconv.AttributeCloudAvailabilityZone}},
+			clusterName:   {otelKeys: []string{semconv.AttributeK8SClusterName}},
+			namespaceName: {otelKeys: []string{semconv.AttributeK8SNamespaceName}},
+			podName:       {otelKeys: []string{semconv.AttributeK8SPodName}},
+			containerName: {otelKeys: []string{semconv.AttributeK8SContainerName}},
 		},
 		k8sPod: {
-			location:      {semconv.AttributeCloudAvailabilityZone},
-			clusterName:   {semconv.AttributeK8SClusterName},
-			namespaceName: {semconv.AttributeK8SNamespaceName},
-			podName:       {semconv.AttributeK8SPodName},
+			location:      {otelKeys: []string{semconv.AttributeCloudAvailabilityZone}},
+			clusterName:   {otelKeys: []string{semconv.AttributeK8SClusterName}},
+			namespaceName: {otelKeys: []string{semconv.AttributeK8SNamespaceName}},
+			podName:       {otelKeys: []string{semconv.AttributeK8SPodName}},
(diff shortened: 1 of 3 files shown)


tbarker25 (approved):

exporter/collector/monitoredresource.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+var (
+	// monitoredResourceMappings contains mappings of GCM resource label keys onto OTel
+	// resource keys for a given monitored resource type. For entries with multiple OTel
+	// resource keys, the keys' values will be coalesced in order until there is a non-empty
+	// value.
+	monitoredResourceMappings = map[string]map[string][]string{
+		gceInstance: {
+			location:   {semconv.AttributeCloudAvailabilityZone},
```
kjordy, line 54: gce_instance has a "zone" resource label, not a "location" label. I updated out spec.

exporter/collector/monitoredresource.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+func createMonitoredResource(
+	monitoredResourceType string,
+	resourceAttrs pdata.AttributeMap,
+) (*monitoredrespb.MonitoredResource, labels) {
+	mappings := monitoredResourceMappings[monitoredResourceType]
+	mrLabels := make(map[string]string, len(mappings))
+	// TODO handle extra labels
```
jsuereth, line 140: Is there a task/bug to handle this?

exporter/collector/monitoredresource.go
```diff
@@ -0,0 +1,166 @@
...
+	region         = "region"
+	taskID         = "task_id"
+)
+
+var (
+	// monitoredResourceMappings contains mappings of GCM resource label keys onto OTel
+	// resource keys for a given monitored resource type. For entries with multiple OTel
+	// resource keys, the keys' values will be coalesced in order until there is a non-empty
+	// value.
+	monitoredResourceMappings = map[string]map[string][]string{
```
jsuereth, line 52: How is this intended to be used with Configuration from the user?   Should we allow users to continue performing MR mapping via config as well, and if so is there a way for that to override this behavior?

Do we expect that to solely be via `map[string]map[string][]string` manipulation of defaults w/ config?

jsuereth (approved):

exporter/collector/monitoredresource.go
```diff
@@ -0,0 +1,166 @@
...
+		gceInstance: {
+			location:   {semconv.AttributeCloudAvailabilityZone},
+			instanceID: {semconv.AttributeHostID},
+		},
+		k8sContainer: {
+			location:      {semconv.AttributeCloudAvailabilityZone},
```
kjordy, line 58: There's a comment in the spec from @dashpole about the value of this field for regional clusters. I don't quite understand how the resource detection would work for regional clusters but this may need to have both zone and then region if zone isn't populated for the regional clusters. Though I'm good with starting with only zone for now until the resource detection piece is figured out.

exporter/collector/monitoredresource_test.go
```diff
@@ -0,0 +1,269 @@
...
+			expectMr: &monitoredrespb.MonitoredResource{
+				Type: "generic_node",
+				Labels: map[string]string{
+					"location":  "my-region",
+					"namespace": "myservicenamespace",
+					"node_id":   "myhostname",
+				},
+			},
+			expectExtraLabels: labels{},
+		},
```
kjordy, line 254: It would be good to add a test where the fall through to generic_node populates the labels with empty string.

exporter/collector/monitoredresource.go
```diff
@@ -0,0 +1,166 @@
...
+			awsAccount: {semconv.AttributeCloudAccountID},
+		},
+		genericTask: {
+			location:  {semconv.AttributeCloudAvailabilityZone, semconv.AttributeCloudRegion},
+			namespace: {semconv.AttributeServiceNamespace},
+			job:       {semconv.AttributeServiceName},
+			taskID:    {semconv.AttributeServiceInstanceID},
+		},
+		genericNode: {
+			location:  {semconv.AttributeCloudAvailabilityZone, semconv.AttributeCloudRegion},
```
kjordy, line 91: The spec says that on the last fall through, the location should be populated with "global" if semantic convention labels aren't populated. This is because writing with a location set to empty string will be an error. Do you think we need to handle that case?

exporter/collector/monitoredresource.go
aabmass: Created a bug for myself. I can do it another PR

exporter/collector/monitoredresource.go
jsuereth: I think we *can't* use global if we default to workload.googleapis.com right?  DO we need to update the spec?

exporter/collector/monitoredresource.go
aabmass: fixed

exporter/collector/monitoredresource.go
aabmass: Fixed for `generic_task/generic_node`. For the other resources it should really be present, so I'll leave those to empty to surface the error

kjordy: Setting the value of the "location" label to "global" is still allowed under workload.googleapis.com.

exporter/collector/monitoredresource_test.go
aabmass: Done

exporter/collector/monitoredresource.go
aabmass: > Though I'm good with starting with only zone for now until the resource detection piece is figured out.

Lets do this for now, but I'll leave this comment unresolved

kjordy (approved):

exporter/collector/monitoredresource.go
aabmass: > Do we expect that to solely be via map[string]map[string][]string manipulation of defaults w/ config?

That would work except we need an extra field for "discriminating" the incoming resource to one of the entries in the map. For the default logic here, this is `cloud.platform` but we have the special cases for differentiating `k8s_{container,pod,node,cluster}` and for the fallbacks.

We could provide a full escape-hatch function config that lets users (probably 1st-party) override this whole thing in a custom build if needed.

But for now can we punt this until we know the use cases better?

PR #253, opened 2021-12-22
Add instrumentationLibraryToLabels method to metrics exporter.

tbarker25:
BUG=https://b.corp.google.com/issues/210164355


exporter/collector/metricsexporter.go
```diff
@@ -129,15 +129,17 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
-					for _, md := range me.mapper.metricDescriptor(metric) {
-						if md != nil {
-							select {
-							case me.mds <- md:
-							default:
-								// Ignore drops, we'll catch descriptor next time around.
-							}
-						}
+				if me.cfg.MetricConfig.SkipCreateMetricDescriptor {
+					continue
```
jsuereth, line 133: This shouldn't be continuing here.  The timeseries still needs to be appended first.

You could move L145 above the if.

exporter/collector/metricsexporter.go
```diff
@@ -227,8 +229,10 @@ func (m *metricMapper) resourceMetricsToMonitoredResource(
 }
 
 func instrumentationLibraryToLabels(il pdata.InstrumentationLibrary) labels {
-	// TODO
-	return nil
+	return labels{
+		"instrumentation_source":  il.Name(),
```
jsuereth, line 233: Should this add source or version if the name is empty?   cc @kjordy / @aabmass

jsuereth:
Forgot to ask:

Could you also add a configuration parameter in `Config.MetricConfig`, something like `instrumentation_library_labels` that defaults to true that users can configure?

Then in the code that maps, only send them if this value is true.

exporter/collector/metricsexporter.go
```diff
@@ -226,9 +234,27 @@ func (m *metricMapper) resourceMetricsToMonitoredResource(
...
-func instrumentationLibraryToLabels(il pdata.InstrumentationLibrary) labels {
-	// TODO
-	return nil
+func (m *metricMapper) instrumentationLibraryToLabels(il pdata.InstrumentationLibrary) labels {
+	source := m.cfg.MetricConfig.InstrumentationSource
+	if source == "" {
+		source = il.Name()
+	}
+	if source == "" {
+		source = "fallback"
```
tbarker25, line 243: @aabmass @jsuereth @kjordy 

Should we have a fallback value for this? And what should it be?

exporter/collector/metricsexporter.go
tbarker25: Good catch. Done.

tbarker25:
> Forgot to ask:
> 
> Could you also add a configuration parameter in `Config.MetricConfig`, something like `instrumentation_library_labels` that defaults to true that users can configure?
> 
> Then in the code that maps, only send them if this value is true.

Done

dashpole:
@jsuereth why does it need to be configurable?  The spec says we MUST send these labels on metrics

exporter/collector/config.go
```diff
@@ -49,6 +49,9 @@ type MetricConfig struct {
 	SkipCreateMetricDescriptor bool   `mapstructure:"skip_create_descriptor"`
 	// If a metric belongs to one of these domains it does not get a prefix.
 	KnownDomains []string `mapstructure:"known_domains"`
+
+	// If true, set the instrumentation_source and instrumentation_version labels
+	InstrumentationLibraryLabels bool `mapstructure:"instrumentation_library_labels"`
```
aabmass, line 54: would be good to mention the default is true

aabmass (approved):

exporter/collector/metricsexporter.go
kjordy: I think empty string is ok here. If we're going to choose a fallback value it should be something clear like "unspecified" but I don't see much more value between that and just empty string.

jsuereth: I'd stick with empty string.  The specification is (now) loose here, there used to be a specififed default but no more (that I can see)

jsuereth:
@dashpole 

> @jsuereth why does it need to be configurable? The spec says we MUST send these labels on metrics

It does.  However I think two things are true:

1. The spec was mostly driven by Tracing exporters (where the cost-benefit analysis vs. tracing may be different)
2. I suspect we'll likely find that we ALWAYS want instrumentation source, but sometimes want to suppress/ignore instrumentation version.   There's a bit of discussion on this around prometheus metric exporter from weyert (can't find the link off hand)

exporter/collector/metricsexporter.go
tbarker25: Discussed. We will set to empty for now, since it is currently unspecified.

exporter/collector/config.go
tbarker25: Good idea. Done.

exporter/collector/metricsexporter.go
tbarker25: I've changed this to use an empty string

tbarker25:
Discussed offline:

We should only set instrumentation_library fields for custom metrics (as they can't be added to default metrics).
This means we have two options:
1. If InstrumentationLibraryLabels is true and metric domain is some custom domain, add those labels. Otherwise don't.
2. No special handling, just rely on InstrumentationLibraryLabels. GKE and ops agent will need to set it to false.

This CL goes with (1)

exporter/collector/metricsexporter.go
```diff
@@ -162,23 +162,33 @@ func (me *metricsExporter) pushMetrics(ctx context.Context, m pdata.Metrics) err
...
 
-			extraLabels := mergeLabels(instrumentationLibraryToLabels(ilm.InstrumentationLibrary()), extraResourceLabels)
+			instrumentationLibraryLabels := me.mapper.instrumentationLibraryToLabels(ilm.InstrumentationLibrary())
+
 			mes := ilm.Metrics()
 			for k := 0; k < mes.Len(); k++ {
 				metric := mes.At(k)
+				metricLabels := extraResourceLabels
+
+				if me.mapper.shouldAddInstrumentationLibraryLabels(metric) {
```
jsuereth, line 172: We made this same logic for resources, actually (@dashpole or @aabmass may have the most context on system-metrics being recorded with our APIs).

jsuereth (approved):

exporter/collector/config.go
```diff
@@ -50,6 +50,16 @@ type MetricConfig struct {
...
 
+	// If true, set the instrumentation_source and instrumentation_version
+	// labels. Defaults to true.
+	InstrumentationLibraryLabels bool `mapstructure:"instrumentation_library_labels"`
+
+	// Set of metric name prefixes to add instrumentation_source and
+	// instrumentation_version labels to. Defaults to custom metric domains:
+	// "custom.googleapis.com/", "external.googleapis.com/",
+	// "prometheus.googleapis.com/", "workload.googleapis.com/"
+	InstrumentationLibraryPrefixes []string `mapstructure:"instrumentation_library_domains"`
```
dashpole, line 61: Can this be hard-coded instead of configurable?  New custom domains are fairly rare.

jsuereth: For similar reasons to why we are allowing configurable metric domains in general, I'd like to have this exposed as an escape hatch.   This should help us avoid getting into a scenario in the future e.g. w/ open-census where we had lots of hard coded logic that become out-of-date and hard to adjust.  Specifically, usually changes to something like this come with a narrow time window.

I'd actually *change* this configuration, however, to be "custom_metric_domains".   Specifically, there's more than IL that needs to diverge its behavior based on whether or not something is in a known custom metric domain.

dashpole: I think `custom_metric_domains` is an improvement, but I still prefer not adding configuration for which we don't have a use-case.  If the list becomes outdated, the solution is to update the list, regardless of whether it is configurable or not.  Unnecessary config just makes using the exporter more complex for users.

> Specifically, there's more than IL that needs to diverge its behavior based on whether or not something is in a known custom metric domain.

What other behavior?  This is the first I know of, since CreateServiceTimeSeries is explicit, rather than based on domain.

tbarker25: discussed offline: we are going to remove this logic entirely and rely on the instrumentation_library_labels flag

dashpole (approved):

PR #259, opened 2022-01-05
Add config to call `CreateServiceTimeSeries`

jsuereth:
- Adds a binary flag where all metrics go to CreateServiceTimeSeries or not.
- Update metric descriptor logic so that *no* metric descriptors are sent if we're in `CreateServiceTimeSeries` mode.
- Tests using an integration test (as there are no unit tests for this portion of code, nor was I sure how to appropriately hijack).

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -19,6 +19,7 @@ package collector
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"math"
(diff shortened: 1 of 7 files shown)


dashpole:
cc @laoj2

dashpole (approved):

jsuereth:
@dashpole I found out we were missing the code to auto-detect ProjectID and update config when instantiating an exporter.   Added it, PTAL

exporter/collector/metricsexporter.go
```diff
@@ -79,6 +81,22 @@ func newGoogleCloudMetricsExporter(
 ) (component.MetricsExporter, error) {
 	setVersionInUserAgent(cfg, set.BuildInfo.Version)
 
+	// TODO - Share this lookup somewhere
+	if cfg.ProjectID == "" {
+		if ctx == nil {
```
dashpole, line 86: Checking for nil context is unusual in go.  Can we make sure the passed-in context is not nil everywhere and remove this check?

jsuereth: this was copy-pasta from opencensus exporter, I can just remove.

dashpole (approved):

PR #260, opened 2022-01-06
Re-enable ops-agent self-metric integration test.

jsuereth:


exporter/collector/internal/integrationtest/testcases.go
```diff
@@ -64,7 +64,12 @@ var (
 			Name:                 "Ops Agent Self-Reported metrics",
 			OTLPInputFixturePath: "testdata/fixtures/ops_agent_self_metrics.json",
 			ExpectFixturePath:    "testdata/fixtures/ops_agent_self_metrics_expect.json",
-			Skip:                 true,
+			Skip:                 false,
+			Configure: func(cfg *collector.Config) {
+				// Previous exporter did NOT export metric descriptors.
+				// TODO: Add a new test that also checks metric descriptors.
+				cfg.MetricConfig.SkipCreateMetricDescriptor = true
```
dashpole, line 71: Should it be using CreateServiceTimeSeries?

jsuereth: That's a separate question.    It was not, when this test was recorded :)

dashpole (approved):

PR #264, opened 2022-01-07
Enable ops agent host metric integration test.

jsuereth:
- There is a bug in upstream agent-metric-processor that sets incorrect units on usage metrics (https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/issues/72)
- We update the expectations for inculsion of units in CreateTimeSeries
- We disable metric descriptors (for now).  Given the bug in agent-metric-processor, liekly ops-agent will need upstream fix for this first.


dashpole (approved):

aabmass (approved):
🎉

PR #265, opened 2022-01-07
Add self observability metrics to integration test fixtures

aabmass:
Capture all of the OC instrumented self observability metrics during integration tests and add them to the test fixtures.

I found out you can reset OC's global state by unregistering and re-registering a view, which is what the [`obsreporttest`](https://github.com/open-telemetry/opentelemetry-collector/blob/964c85766e9290f3be08b2dbec9e228527a20b21/obsreport/obsreporttest/obsreporttest.go#L78) package does. Unfortunately, you have to know the views before hand which required some refactoring; OC doesn't give a mechanism to get all views.


exporter/collector/internal/integrationtest/fixtures.proto
```diff
@@ -22,4 +22,14 @@ message MetricExpectFixture {
...
+  repeated SelfObservabilityMetric self_observability_metrics = 4;
+}
+
+message SelfObservabilityMetric {
+  // metric name
+  string name = 1;
+  // stringified value
+  string val = 2;
+  // stringified labels
+  map<string, string> labels = 3;
```
aabmass, line 34: Is this enough information or do we also want to capture type, description, etc.. On the extreme side, I could hack the OCAgent exporter and capture the full proto, but if we ever migrate these to OTel metrics API, this minimal set should be compatible.

exporter/collector/factory.go
```diff
@@ -31,14 +30,10 @@ const (
 	defaultTimeout = 12 * time.Second // Consistent with Cloud Monitoring's timeout
 )
 
-var once sync.Once
-
 // NewFactory creates a factory for the googlecloud exporter
 func NewFactory() component.ExporterFactory {
-	// register view for self-observability
-	once.Do(func() {
```
aabmass, line 39: This `once` shouldn't be necessary https://github.com/census-instrumentation/opencensus-go/blob/bf52d9df8bb2d44cad934587ab946794456cf3c8/stats/view/worker.go#L333-L335

exporter/collector/internal/integrationtest/fixtures.proto
jsuereth: @sophieyfang for thoughts

jsuereth (approved):
LGTM, just one clarifying comment.

exporter/collector/metrics_test.go
```diff
@@ -28,6 +28,7 @@ func TestMetrics(t *testing.T) {
 	ctx := context.Background()
 	endTime := time.Now()
 	startTime := endTime.Add(-time.Second)
+	// view.SetReportingPeriod(time.Millisecond * 250)
```
dashpole, line 31: remove/uncomment code

dashpole (approved):

exporter/collector/testdata/fixtures/gke_control_plane_metrics_agent_metrics_expect.json
```diff
@@ -1062,5 +1062,49 @@
         }
       ]
     }
+  ],
+  "selfObservabilityMetrics": [
```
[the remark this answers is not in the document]
aabmass: > Any new OC metrics would cause this to fail

Yes and running the regeneration script would add any new metrics. Is there a different approach you'd suggest?

> It only sees OC metrics instantiated as part of our integration test setup

Yes which includes our instrumentation in observability.go and exporterhelper in the collector repo. There could be more.

> We'll need to rework the mechanism to pull self-observability metrics in the future, but we plan to always map it to this abstraction.

Which self-observability metrics do you mean here?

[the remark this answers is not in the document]
sophieyfang: > Yes and running the regeneration script would add any new metrics. Is there a different approach you'd suggest?

(Ugh.. i deleted my previous comment...😑)
Is there a public/internal doc that has those steps documented? Things like how was the script triggered? Is it possible to automate the script so that it can sync up with the latest status

sophieyfang (approved):

exporter/collector/testdata/fixtures/gke_control_plane_metrics_agent_metrics_expect.json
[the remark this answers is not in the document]
aabmass: Ya, it's documented here https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtests.md#adding-new-testshttps://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtests.md#adding-new-tests. The idea is to manually run the script when you make an intentional change or add a new test. We shouldn't need to automate since the tests should fail before it goes out of sync.

PR #267, opened 2022-01-07
Add a feature gate, which defaults to false, for using the re-written exporter

dashpole:
This allows the new pdata-based exporter to be enabled on the collector with `--feature-gates=exporter.googlecloud.OTLPDirect` when running the collector.

Code in exporter/collector/legacymetricsexporter.go (and the corresponding `_test.go` file) is taken from the [current upstream implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6a1c2471092b7f62ad1a37f62cbb7b0179676d01/exporter/googlecloudexporter/googlecloud.go#L126), which wraps the [opencensus stackdriver exporter](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver).

This means that changing the upstream implementation to use the rewritten implementation will be a no-op.


jsuereth (approved):
Would you mind documenting what this would look like in a readme/config/usage style for users of this expoerter?

dashpole:
Done

PR #270, opened 2022-01-10
Fix recordfixtures script to use featuregate

aabmass:
Follow up to #267. Feature gate needs to also be set in the recordfixtures test, or they will record using the old exporter. This unconditionally sets the feature gate before running the script.

- made feature gate string const public
- regenerated fixtures which produced some reformatted json with the same content.

Code re-use is tough here because of circular dependencies.

### DIFF
--- a/exporter/collector/testdata/fixtures/delta_counter_metrics_expect.json
+++ b/exporter/collector/testdata/fixtures/delta_counter_metrics_expect.json
@@ -18,19 +18,19 @@
             }
           },
           "metricKind": "CUMULATIVE",
-          "unit": "1",
           "valueType": "INT64",
           "points": [
             {
               "interval": {
-                "start_time": "1970-01-01T00:00:01Z",
-                "endTime": "1970-01-01T00:00:00Z"
+                "endTime": "1970-01-01T00:00:00Z",
+                "startTime": "1970-01-01T00:00:00Z"
               },
               "value": {
                 "int64Value": "253"
               }
             }
-          ]
+          ],
+          "unit": "1"
         }
       ]
     }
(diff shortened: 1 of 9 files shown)


exporter/collector/testdata/fixtures/delta_counter_metrics_expect.json
```diff
@@ -18,19 +18,19 @@
...
           "metricKind": "CUMULATIVE",
-          "unit": "1",
           "valueType": "INT64",
           "points": [
             {
               "interval": {
-                "start_time": "1970-01-01T00:00:01Z",
-                "endTime": "1970-01-01T00:00:00Z"
+                "endTime": "1970-01-01T00:00:00Z",
+                "startTime": "1970-01-01T00:00:00Z"
```
dashpole, line 26: Looks like the jitter in the start time was removed?

aabmass: I don't think so. `1970-01-01T00:00:00Z` is just used as a placeholder to determine if that timestamp field is present. When the recordfixtures script runs, it saves it with the normalized placeholder value.

aabmass: Basically the value here is ignored beyond that it needs to be a parseable datetime

dashpole: Maybe i'm confusing things.  The OC exporter will change the start time to be different from the end time if they are the same (adding jitter).  I thought maybe the difference between start/end was due to that behavior, and disappeared when switching to otel.  But if thats incorrect, then the rest of the change is fine

aabmass: I remember you mentioning that. I think we are actually missing that functionality, but because of timestamp modifications, the integration tests won't work for capturing that. Probably need a separate unit test.

I'll open a bug.

dashpole: sg

PR #275, opened 2022-01-11
Move logging to use zap-logger and set up self-observability to match…

jsuereth:
… collector expectations.

- Adds hook for us to create Meters/Instruments/Tracers from the passed-in collector TelemetrySettings
- Stores the zap logger for structured logging in the exporter.

Side question:  Why are we using OpenCensus for self-observability metrics instead of MeterProvider in TelemetrySettings?

### DIFF
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"math"
 	"net/url"
 	"path"
@@ -32,6 +31,7 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 	"go.opentelemetry.io/collector/model/pdata"
+	"go.uber.org/zap"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/genproto/googleapis/api/distribution"
 	"google.golang.org/genproto/googleapis/api/label"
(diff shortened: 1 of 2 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -42,10 +42,17 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// self-observability reporting meters/tracers/loggers.
+type selfObservability struct {
```
dashpole, line 46: why define a new type if component.TelemetrySettings is already meant to encapsulate initialized telemetry instruments?

aabmass:
> Side question: Why are we using OpenCensus for self-observability metrics instead of MeterProvider in TelemetrySettings?

Looks like it was added only 4 months back and no-one in opentelemetry-collector-contrib is using it yet. Another thing is Sophie is using the ocgrpc instrumentation and I'm not sure if there is a compatible OTel instrumentation.

Do we think OTel Go metrics is stable enough to migrate our metrics to it?

dashpole:
I would not recommend using otel metrics until its stable.  We have had a hard enough time with changes that it seems unnecessary to add another source.

jsuereth:
> I would not recommend using otel metrics until its stable. We have had a hard enough time with changes that it seems unnecessary to add another source.

Sounds good.  Let's see how much we can help prioritize stabilizing Go SDK metrics.

aabmass (approved):

dashpole (approved):

PR #269, opened 2022-01-10
Update otel dependencies to the latest version

dashpole:
MinMaxSumCount was removed (although we never supported it anyways).

export.ExportKind -> aggregation.Temporality.

And Descriptor moved to the sdkapi package.

Also, bump golang version to 1.16 everywhere to fix a build failure.

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/241

PR #276, opened 2022-01-11
merge main into col-exporter-rewrite

aabmass:
- fix trace initialization error message (#262)
- Update otel dependencies to the latest version (#269)
- Add self observability metrics to integration test fixtures (#265)
- Update Workload Metrics testcase to use workload.googleapis.com prefix (#273)


aabmass:
pushing manually

PR #278, opened 2022-01-11
Re-enable workload metrics integration tests

aabmass:
- Updated WLM test config with `skip_metric_descriptor` to true to pass tests
- Updated breaking-changes.md to reflect this is needed

The generated changes are from units being sent in CreateTimeSeries calls, which is functionally a no-op. The self-obs changes are because this was generated off main branch.

### DIFF
--- a/exporter/collector/integrationtest/testcases.go
+++ b/exporter/collector/integrationtest/testcases.go
@@ -83,9 +83,9 @@ var (
 			Name:                 "GKE Workload Metrics",
 			OTLPInputFixturePath: "testdata/fixtures/workload_metrics.json",
 			ExpectFixturePath:    "testdata/fixtures/workload_metrics_expect.json",
-			Skip:                 true,
 			Configure: func(cfg *collector.Config) {
 				cfg.MetricConfig.Prefix = "workload.googleapis.com/"
+				cfg.MetricConfig.SkipCreateMetricDescriptor = true
 			},
 		},
 		{
--- a/exporter/collector/integrationtest/testdata/fixtures/batching_metrics_expect.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/batching_metrics_expect.json
@@ -1,8329 +1,8388 @@
 {
-  "createTimeSeriesRequests":  [
+  "createTimeSeriesRequests": [
     {
-      "name":  "projects/fakeprojectid",
-      "timeSeries":  [
+      "timeSeries": [
         {
-          "metric":  {
-            "type":  "workload.googleapis.com/testsummary 1_summary_sum",
-            "labels":  {
-              "foo":  "bar"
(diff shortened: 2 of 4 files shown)


exporter/collector/testdata/fixtures/workload_metrics_expect.json
```diff
@@ -6452,7 +6467,7 @@
     },
     {
       "name": "googlecloudmonitoring/point_count",
-      "val": "200",
```
aabmass, line 6455: There are three points in this fixture, each with no percentiles. So this increased by three because each one contributes sum and count now.

exporter/collector/testdata/fixtures/workload_metrics_expect.json
```diff
@@ -6307,11 +6326,7 @@
               }
             }
           ]
-        }
-      ]
-    },
-    {
-      "timeSeries": [
```
aabmass, line 6314: This looks problematic, the OC exporter was splitting this into two request batches while we are just sending one..

dashpole: I thought I was using the batch processor when I generated these fixtures?  Is the batch processor not splitting correctly?

aabmass: Discussed with David. What's happening is the batchprocesor gives 200 OTLP points, but we split the summary points into multiple timeseries, which gives us more than the 200 metric limit. Possible fixes are:

1. Implement batching in the exporter, which is how the OC exporter got around this, or
2. Unpack the summary points into multiple timeseries in a new processor *before* batchprocessor

jsuereth: 1. also gives us the opportunity to deal with other oddities around sending metric data points (e.g. multiple points in the same timeseries).  I think this has my vote as an inevitability here.   Given we can't really control how users set up processors, I think we should aim for this exporter to be as resilient as possible to things users will shove at it.

dashpole: Just to see where it went, I opened https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7134.  But implementing batching in the exporter (could we use the batch processor as a library?) is definitely the most expedient solution.

aabmass: Discussed at standup. We are implement some batching in our exporter for expediency and maybe revisit this in the future depending what happens with David's issue. Not clear yet if we want to try and re-use the batchprocessor's batching code or do something bespoke based on GCM's unique batching requirements.

We will probably stop recommending users use `batchprocessor` in their pipeline though.

dashpole (approved):

PR #279, opened 2022-01-12
Remove dependency on go.opentelemetry.io/collector

dashpole:
Based on https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7132#issuecomment-1011304861, we should try and reduce our dependencies on collector core to the bare minimum. After this change, this package only references `go.opentelemetry.io/collector/model/pdata`, and `go.opentelemetry.io/collector/model/semconv/v1.8.0`, which are all within the `go.opentelemetry.io/collector/model` go module.

This PR moves the `go.opentelemetry.io/collector` dependency, which includes the collector component interfaces, as well as exporterhelper packages.

It creates a new module for `exporter/collector/internal/integrationtest` so that tests can still use collector helpers to verify that config parsing, etc. still works.

The way this package would work is that the NewFactory() method would exist in collector-contrib, but would call into this package's new public methods:

* DefaultConfig()
* NewGoogleCloudTracesExporter()
* NewGoogleCloudMetricsExporter()
* MetricExporter.PushMetrics()
* MetricExporter.Shutdown()
* TraceExporter.PushTraces()
* TraceExporter.Shutdown()

The collector.Config struct would be embedded into the Config struct with the `mapstructure:",squash"` tag to preserve backwards-compatibility, see config_test.go for an example of how it could be embedded within a config.

Removing the collector dependency required the following additional changes:

* Remove the feature gate.  This will live upstream.
* Remove tests for exporterhelper-produced metrics.  These will be replaced by upstream unit (not integration) tests.


jsuereth (approved):
read through all the back discussion and this LGTM as a path forward.

exporter/collector/config_test.go
```diff
@@ -1,100 +0,0 @@
...
-)
-
-func TestLoadConfig(t *testing.T) {
-	defer SetPdataFeatureGateForTest(true)()
-	factories, err := componenttest.NopFactories()
-	assert.Nil(t, err)
-
-	factory := NewFactory()
-	factories.Exporters[typeStr] = factory
-	cfg, err := configtest.LoadConfigAndValidate(path.Join(".", "testdata", "config.yaml"), factories)
```
aabmass, line 37: One concern here, this test is not trivial because it validates the mapstructure tags on fields of certain types (I actually ran into this a few PRs back). So as an example,

- we add a new public string struct member without a tag and cut a release
- when we go to upgrade the version of this mod in collector-contrib, there would be a test failure.
- we have to fix it, rinse and repeat.

Would it make sense to move these smoke tests into the unreleased testing module along with the integration tests? That way the actual exporter doesn't depend on it, but we can still catch these issues in CI.

exporter/collector/internal/integrationtest/metric_test_server.go
```diff
@@ -151,19 +150,19 @@ func (m *MetricsTestServer) NewExporter(
...
 	cfg collector.Config,
-) component.MetricsExporter {
-	factory := collector.NewFactory()
+) *collector.MetricsExporter {
 	cfg.Endpoint = m.Endpoint
 	cfg.UseInsecure = true
 	cfg.ProjectID = "fakeprojectid"
 
-	exporter, err := factory.CreateMetricsExporter(
+	exporter, err := collector.NewGoogleCloudMetricsExporter(
```
aabmass, line 158: > Remove tests for exporterhelper-produced metrics. These will live upstream.

I'm a little confused how these would look upstream, would we move the whole integration testing setup there?

IIRC you were going to duplicate the factory + exporterhelper code in a separate integration test module so that the tests are as realistic as possible.

exporter/collector/config_test.go
dashpole: That makes sense to me.  Would you like that done in this PR?

exporter/collector/internal/integrationtest/metric_test_server.go
dashpole: It wouldn't be full end-to-end tests if they are upstream.  It would have to use something like a mock to check and see if metrics are recorded and incremented correctly.  Alternatively, the unit testing in the exporterhelper library itself may be sufficient.  Maybe we just need to check to see if the metrics are registered or something else super-simple.

exporter/collector/config_test.go
aabmass: Up to you whatever is easiest!

aabmass (approved):

exporter/collector/config_test.go
dashpole: I'll do it in this PR (in a separate commit so you don't have to re-review everything), since i'd like this to merge after https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/253 so it doesn't disrupt that effort.

exporter/collector/internal/integrationtest/metric_test_server.go
dashpole: updated the PR description to capture this.

exporter/collector/config_test.go
dashpole: Done, and rebased on #253.  I had to make a wrapper struct for the config, and I decided to drop queued retry and other structs not in the config.  But otherwise, I was able to keep the config parsing test intact.

aabmass (approved):
🚢

PR #280, opened 2022-01-18
Add ocgrpc metrics to exporter's self-obs metrics

dashpole:
This ensures we continue to get metrics added in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6945 when users switch to the new exporter.  I'm adding the ocgrpc metrics here (rather than upstream) so that we can integration test them.

Install the views in both the trace and metric exporters, since duplicate registration will just be a no-op.  That ensures users of either the trace or metrics exporter get opencensus grpc metrics.

Also, fix instructions for generating fixtures, which should have been updated when the testdata folder was moved.

### DIFF
--- a/exporter/collector/googlecloud.go
+++ b/exporter/collector/googlecloud.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"go.opencensus.io/plugin/ocgrpc"
+	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/collector/model/pdata"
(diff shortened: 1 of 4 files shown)


dashpole:
@aabmass any tips on dealing with self-obs metrics that change each test run?  Some of the things (latency, bytes sent, etc.) seem to change

aabmass (approved):
> @aabmass any tips on dealing with self-obs metrics that change each test run? Some of the things (latency, bytes sent, etc.) seem to change

Discussed offline, will do in a follow up PR.

PR #282, opened 2022-01-20
Use OC stackdriver exporter to capture self observability metrics as GCM protos

aabmass:
Collect the OC self observability metrics using the same `MetricTestServer` (GCM mock server) + stackdriver exporter instead of a bespoke proto format. They now get serialized as they would be exported to GCM.


exporter/collector/internal/integrationtest/config/config_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package integrationtest
+package config
```
aabmass, line 15: I had to move this into a separate package because it imports `go.opentelemetry.io/collector/exporter/exporterhelper`, which has an init function creating some instruments/metric producers: https://github.com/open-telemetry/opentelemetry-collector/blob/a8ff2dd98c254538b97348a10c7f270e94991caf/exporter/exporterhelper/obsreport.go#L36-L37. This causes metric descriptors for these to show in the integration tests, but not in the `recordfixtures` script (which doesn't import exporterhelper or run that init).

The fix works because tests in separate packages produce different binaries by default so this stops the side effect for the integration tests.

exporter/collector/internal/integrationtest/fixtures.proto
```diff
@@ -22,14 +22,10 @@ message MetricExpectFixture {
   repeated google.monitoring.v3.CreateTimeSeriesRequest create_time_series_requests = 1;
   repeated google.monitoring.v3.CreateMetricDescriptorRequest create_metric_descriptor_requests = 2;
   repeated google.monitoring.v3.CreateTimeSeriesRequest create_service_time_series_requests = 3;
-  repeated SelfObservabilityMetric self_observability_metrics = 4;
+  SelfObservabilityMetric self_observability_metrics = 4;
```
aabmass, line 25: The fields in SelfObservabilityMetric could be flattened into this message, but I feel like this nesting makes the JSON easier to read than a long name.

exporter/collector/internal/integrationtest/config/config_test.go
dashpole: Did you consider importing the exporterhelper package in the record fixtures script?  Or are those metrics we aren't interested in testing?

exporter/collector/internal/integrationtest/inmemoryocexporter.go
```diff
@@ -16,85 +16,108 @@ package integrationtest
...
 }
 
-func (i *InMemoryOCExporter) ExportMetrics(ctx context.Context, data []*metricdata.Metric) error {
-	i.c <- data
-	return nil
+func getViews() []*view.View {
+	views := []*view.View{}
+	views = append(views, collector.MetricViews()...)
+	// TODO: enable this
+	// views = append(views, ocgrpc.DefaultClientViews...)
```
dashpole, line 42: Should this TODO be fixed in this PR?

dashpole (commented):
Why does using the OpenCensus exporter help us?  Won't the metrics still have inconsistent values?

exporter/collector/internal/integrationtest/inmemoryocexporter.go
aabmass: I'm gonna do in the next PR along with the normalization of the values for the problematic ocgrpc metrics

exporter/collector/internal/integrationtest/config/config_test.go
aabmass: That worked as well to make it pass. Those metrics end up having no points (because no measurements made) but the SD exporter creates metric descriptors. So it's kinda just noise, wdyt?

dashpole: sg

aabmass:
> Why does using the OpenCensus exporter help us? Won't the metrics still have inconsistent values?

Yes it doesn't fix that problem, but i was going down a rabbit hole of trying to add metric type and histogram bucket options to the `SelfObservabilityMetric` and figured it's easier to just use the GCM proto.

This is also better if/when we switch these self obs metrics to OTel instrumentation, we can just use our OTel exporter to capture the same metrics rather than write conversion to a bespoke protobuf format.

dashpole:
> it's easier to just use the GCM proto.

I think i've almost connected the dots.  How does using opencensus relate to using the GCM proto?

aabmass:
> How does using opencensus relate to using the GCM proto?

Do you mean using the OpenCensus stackdriver exporter? It's just the way to serialize OC metrics to GCM protobufs. I may be misunderstanding your question

dashpole:
Ah, ok.  I got it now.  We are essentially changing from opencensus to GCM protos.

dashpole (approved):

aabmass:
> Ah, ok. I got it now. We are essentially changing from opencensus to GCM protos.

Yup. Probably should have explained that better, sorry

PR #283, opened 2022-01-20
Capture ocgrpc self observability metrics

aabmass:
The captured ocgrpc time series and metric descriptors need to be normalized so the tests are reproducible. In addition to the normalization we already do (reset timestamps):
- Clear values of metrics like latency/byte size which vary across runs. For int/float metrics just unset the value. For distributions, we clear the bucket_counts field, but preserve the bucket options.
- Sort metrics into a reproducible order

### DIFF
--- a/exporter/collector/internal/integrationtest/diff.go
+++ b/exporter/collector/internal/integrationtest/diff.go
@@ -15,6 +15,8 @@
 package integrationtest
 
 import (
+	"testing"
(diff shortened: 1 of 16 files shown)


dashpole (approved):

dashpole:
looks great!

aabmass:
thanks!

PR #285, opened 2022-01-21
Make integrationtest public

dashpole:
This is so that users of this exporter can easily write similar GCM integration tests to the ones we have here.


aabmass (approved):

punya:
I have no objection to this PR, but I thought we also discussed adding an option to dump out RPC requests. Are we still planning to do that in a separate PR?

punya (approved):

dashpole:
I think we can do that at a later point if we still want to.  Other users of the exporter can actually start using the integration tests directly, even while using the opencensus-based exporter.

PR #286, opened 2022-01-21
Add batching support to metrics-exporter.

tbarker25:
Add batching support to GCM exporter.

BUG=214237990


exporter/collector/metricsexporter.go
```diff
@@ -126,22 +158,28 @@ func NewGoogleCloudMetricsExporter(
...
-		// MetricDescritpors are asycnhronously sent and optimistic.
+		// MetricDescritpors are asychronously sent and optimistic.
 		// We only get Unit/Description/Display name from them, so it's ok
 		// to drop / conserve resources for sending timeseries.
-		mds:     make(chan *metricpb.MetricDescriptor, cfg.MetricConfig.CreateMetricDescriptorBufferSize),
-		mdsDone: make(chan struct{}),
+		metricDescriptorC: make(chan *metricpb.MetricDescriptor, cfg.MetricConfig.CreateMetricDescriptorBufferSize),
+		mdCache:           make(map[string]*metricpb.MetricDescriptor),
+		timeSeriesC:       make(chan *monitoringpb.TimeSeries),
+		shutdownC:         make(chan struct{}, 1),
```
dashpole, line 167: nit: I don't think this needs to be a buffered channel

exporter/collector/metricsexporter.go
```diff
@@ -156,57 +194,95 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pdata.Metrics) err
...
+		st = statusCodeToString(s)
+	} else {
+		st = "UNKNOWN"
 	}
-	err := me.createTimeSeries(ctx, timeSeries)
-	recordPointCount(ctx, len(timeSeries), m.DataPointCount()-len(timeSeries), err)
-	return err
+
+	recordPointCountDataPoint(ctx, len(ts), st)
+	me.obs.log.Error("could not export time series to GCM", zap.Error(err))
```
dashpole, line 255: do we only want to do this if err != nil?

exporter/collector/metricsexporter.go
```diff
@@ -925,3 +1008,52 @@ func mapMetricPointKind(m pdata.Metric) (metricpb.MetricDescriptor_MetricKind, m
...
+			}
+			// Return and continue graceful shutdown.
+			return
+		case item := <-me.timeSeriesC:
+			me.processItem(item)
+		case <-me.batchTimeoutTimer.C:
+			for len(me.pendingTimeSerieses) > 0 {
+				me.exportPendingTimeSerieses()
+			}
+			me.batchTimeoutTimer.Reset(batchTimeout)
```
dashpole, line 1040: Should we reset the timer before exporting pending time series so that each batch only contains 10s of metrics, and rather than 10s + createTimeSeries latency?

dashpole (commented):
for testing, we might want to generate a new fixture from the gke metrics agent, but without the batch processor already batching things for us.

dashpole:
> Should we buffer the send-timeseries channel? If we do throughput will likely increase but back-pressure might be less apparent

I would prefer not buffering so we still get back-pressure.

exporter/collector/metricsexporter.go
```diff
@@ -156,57 +194,95 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pdata.Metrics) err
...
+		select {
+		case <-me.shutdownC:
+			for {
+				// We are shutting down. Publish all the pending
+				// items on the channel before we stop.
+				select {
+				case md := <-me.metricDescriptorC:
+					me.exportMetricDescriptor(md)
+				default:
+					goto DONE
```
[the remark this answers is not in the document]
tbarker25: Yep. Done.

exporter/collector/metricsexporter.go
```diff
@@ -156,57 +194,95 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pdata.Metrics) err
...
-		return err
+		err = me.createServiceTimeSeries(ctx, ts)
+	} else {
+		err = me.createTimeSeries(ctx, ts)
+	}
+
+	var st string
+	if err == nil {
+		st = "OK"
+	} else if s, ok := status.FromError(err); ok {
```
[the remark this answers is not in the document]
tbarker25: Done. When I was looking at the implementations I forgot that methods could exist for nil

exporter/collector/metricsexporter.go
tbarker25: Done

exporter/collector/metricsexporter.go
tbarker25: Done.

exporter/collector/metricsexporter.go
tbarker25: Done

exporter/collector/metricsexporter.go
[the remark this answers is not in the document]
tbarker25: Done

exporter/collector/metricsexporter.go
```diff
@@ -156,57 +194,95 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pdata.Metrics) err
...
 		}
 	}
 
 	// TODO: self observability
+
+	return nil
+}
+
+func (me *MetricsExporter) exportPendingTimeSerieses() {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
tbarker25: Nope. We call `defer me.goroutines.Done()` when we are spinning up a new goroutine. This runs in an existing one.

exporter/collector/metricsexporter.go
```diff
@@ -215,14 +286,21 @@ func (me *MetricsExporter) projectName() string {
 }
 
 // Helper method to send metric descriptors to GCM.
-func (me *MetricsExporter) exportMetricDescriptor(ctx context.Context, md *metricpb.MetricDescriptor) error {
-	// export
+func (me *MetricsExporter) exportMetricDescriptor(md *metricpb.MetricDescriptor) {
+	if _, exists := me.mdCache[md.Type]; exists {
+		return
+	}
+	me.mdCache[md.Type] = md
```
dashpole, line 293: should we only set this if we were successful in creating the metric descriptor?

dashpole (approved):

exporter/collector/metricsexporter.go
tbarker25: Hmm. I guess that depends on whether we believe the error to be ephemeral or not.

dashpole (approved):

Issue #221, opened 2021-10-21
`service.name` is truncated from the trace attributes?


tyrone-anz:
I noticed `service.name` and other attributes are mapped to different label `g.co/gae/app/module` before being exported to googlecloud.

Is it possible to keep a copy with `service.name` attribute as well? 

Also for the other attributes listed below?

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/44a596863f1ef0ccc374f086b2f2502de0f78fe5/exporter/trace/trace_proto.go#L234

PR #287, opened 2022-01-24
Convert OTel Resource into metric labels

aabmass:
Copy OTel Resource attributes `service.name`, `service.namespace` and `service.instance.id` into metric labels by default. This avoids duplicate timeseries when multiple instances of a service are running on a single monitored resource, for example running multiple service processes on a single GCE VM. Added a config option `metric.include_service_resource_attributes` to control this.

Also added a `metric.resource_filters` config option which which allows copying certain resource attributes onto metrics. This has been asked for in our trace exporters (https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/347, https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/145, https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/221 maybe). It may be necessary for users to configure this also when the `service.name,service.namespace,service.instance.id` triplet is not present to distinguish duplicate time series.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -66,9 +66,13 @@ type MetricConfig struct {
 	// resource. Default is true.
 	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
 
-	// If provided, resources attributes with keys matching this regex will be copied into
-	// metric labels.
-	ResourceFilter string `mapstructure:"resource_filter"`
+	// If provided, resources attributes matching any filter will be included in metric labels.
+	ResourceFilters []ResourceFilter `mapstructure:"resource_filter"`
+}
+
+type ResourceFilter struct {
+	// Match resource keys by prefix
+	Prefix string `mapstructure:"prefix"`
 }
 
 // ResourceMapping defines mapping of resources from source (OpenCensus) to target (Google Cloud).
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -59,6 +59,16 @@ type MetricConfig struct {
 	// Buffer size for the channel which asynchronously calls CreateMetricDescriptor. Default
 	// is 10.
 	CreateMetricDescriptorBufferSize int `mapstructure:"create_metric_descriptor_buffer_size"`
+
+	// If true, the exporter will copy OTel's service.name, service.namespace, and
+	// service.instance.id resource attributes into the GCM timeseries metric labels. This
+	// option is recommended to avoid writing duplicate timeseries against the same monitored
+	// resource. Default is true.
+	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
+
+	// If provided, resources attributes with keys matching this regex will be copied into
+	// metric labels.
+	ResourceFilter string `mapstructure:"resource_filter"`
 }
 
 // ResourceMapping defines mapping of resources from source (OpenCensus) to target (Google Cloud).
@@ -89,6 +99,7 @@ func DefaultConfig() Config {
 			Prefix:                           "workload.googleapis.com",
 			CreateMetricDescriptorBufferSize: 10,
 			InstrumentationLibraryLabels:     true,
+			IncludeServiceResourceAttributes: true,
 		},
 	}
 }
--- a/exporter/collector/integrationtest/testdata/fixtures/with_resource_filter_metrics_expect.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/with_resource_filter_metrics_expect.json
@@ -1,0 +1,685 @@
+{
+  "createTimeSeriesRequests": [
+    {
+      "timeSeries": [
+        {
+          "metric": {
+            "type": "workload.googleapis.com/testresourcefilter",
+            "labels": {
+              "foo": "bar",
+              "service_instance_id": "abc123",
+              "service_name": "foo-service",
+              "service_namespace": "foo",
+              "telemetry_sdk_language": "java",
+              "telemetry_sdk_version": "1.1.1"
+            }
+          },
+          "resource": {
+            "type": "generic_task",
+            "labels": {
+              "job": "foo-service",
+              "location": "global",
+              "namespace": "foo",
+              "task_id": "abc123"
+            }
+          },
+          "metricKind": "CUMULATIVE",
+          "valueType": "INT64",
+          "points": [
+            {
+              "interval": {
+                "endTime": "1970-01-01T00:00:00Z",
+                "startTime": "1970-01-01T00:00:00Z"
+              },
+              "value": {
+                "int64Value": "253"
+              }
+            }
+          ],
+          "unit": "1"
+        }
+      ]
+    }
+  ],
+  "createMetricDescriptorRequests": [
+    {
+      "metricDescriptor": {
+        "type": "workload.googleapis.com/testresourcefilter",
+        "labels": [
+          {
+            "key": "foo"
+          }
+        ],
+        "metricKind": "CUMULATIVE",
+        "valueType": "INT64",
+        "unit": "1",
+        "description": "This is a test counter",
+        "displayName": "testresourcefilter"
+      }
+    }
+  ],
+  "selfObservabilityMetrics": {
+    "createTimeSeriesRequests": [
+      {
+        "timeSeries": [
+          {
+            "metric": {
+              "type": "custom.googleapis.com/opencensus/googlecloudmonitoring/point_count",
+              "labels": {
+                "status": "OK"
+              }
+            },
+            "resource": {
+              "type": "global"
+            },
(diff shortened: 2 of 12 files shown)


exporter/collector/integrationtest/testdata/fixtures/with_resource_filter_metrics_expect.json
```diff
@@ -0,0 +1,685 @@
...
+      ]
+    }
+  ],
+  "createMetricDescriptorRequests": [
+    {
+      "metricDescriptor": {
+        "type": "workload.googleapis.com/testresourcefilter",
+        "labels": [
+          {
+            "key": "foo"
```
aabmass, line 50: Note that the labels from the monitored resource are NOT in the CMD call as written. I think this should probably be updated.

aabmass (commented):
I have some concerns with this approach mostly around churning metric descriptors since they are global regardless of who is writing them or against what monitored resource. If the `service.*` resource attributes are not present, the exporter will send a time series/CMD without them. The same issue could happen with the `resource_filter` option is set and some metrics get written with different resource filters.

I think this should all be fine for workload.googleapis.com metrics, but is kinda dangerous for non-additive domains. Thoughts?

exporter/collector/config.go
```diff
@@ -59,6 +59,16 @@ type MetricConfig struct {
 	// Buffer size for the channel which asynchronously calls CreateMetricDescriptor. Default
 	// is 10.
 	CreateMetricDescriptorBufferSize int `mapstructure:"create_metric_descriptor_buffer_size"`
+
+	// If true, the exporter will copy OTel's service.name, service.namespace, and
+	// service.instance.id resource attributes into the GCM timeseries metric labels. This
+	// option is recommended to avoid writing duplicate timeseries against the same monitored
+	// resource. Default is true.
+	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
```
dashpole, line 67: I don't think we need two config options to gate this behavior... `include_service_resource_attributes=true` seems to be equivalent to `resource_filter=service.*`.  Also, this should probably be a list?  What if I want my service.* and my host.*, but not others?

exporter/collector/config.go
```diff
@@ -59,6 +59,16 @@ type MetricConfig struct {
 	// Buffer size for the channel which asynchronously calls CreateMetricDescriptor. Default
 	// is 10.
 	CreateMetricDescriptorBufferSize int `mapstructure:"create_metric_descriptor_buffer_size"`
+
+	// If true, the exporter will copy OTel's service.name, service.namespace, and
+	// service.instance.id resource attributes into the GCM timeseries metric labels. This
+	// option is recommended to avoid writing duplicate timeseries against the same monitored
+	// resource. Default is true.
```
dashpole, line 66: I feel like this behavior should be off by default.  It may make sense for this to be enabled by default for some ways of running the collector (i.e. maybe the ops-agent should enable this by default), but for OSS users, this behavior would be surprising to me.

Also, if we someday want to make the behavior better (e.g. confluence for resource attributes), writing these as metric labels by default might be confusing (it would show up as both a metric label and a resource attribute).

exporter/collector/config.go
```diff
@@ -59,6 +59,16 @@ type MetricConfig struct {
...
+	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
+
+	// If provided, resources attributes with keys matching this regex will be copied into
```
dashpole, line 69: As an alternative to regex, we could consider prefixes only.  It makes this a bit less of a foot-gun, in that users can't unintentionally match multiple prefixes (e.g. service.namespace and k8s.namespace.name with .*namespace.*), and also makes it more readable (no .* regex notation anywhere).  It may not be as powerful as regex, though.  Prometheus relabel configs have proven very powerful with regex matching.

exporter/collector/config.go
```diff
@@ -59,6 +59,16 @@ type MetricConfig struct {
...
+	// metric labels.
+	ResourceFilter string `mapstructure:"resource_filter"`
```
dashpole, line 71: I think this should be a list.

exporter/collector/config.go
aabmass: Relevant part of the spec on these:

> [[3]: MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique). The ID helps to distinguish instances of the same service that exist at the same time (e.g. instances of a horizontally scaled service).](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.8.0/specification/resource/semantic_conventions#:~:text=%5B3%5D%3A%20MUST%20be%20unique%20for%20each%20instance%20of%20the%20same%20service.namespace%2Cservice.name%20pair%20(in%20other%20words%20service.namespace%2Cservice.name%2Cservice.instance.id%20triplet%20MUST%20be%20globally%20unique).)


So other service attributes (just `service.version` atm) are not needed for uniqueness. You could do `service.name|service.instance.id|service.namespace` but it would be easy to accidentally override this behavior when you specify a `resource_filter`. That was my thinking for separating it.

exporter/collector/config.go
aabmass: See comment above about uniqueness. For OSS users, I think this should be on by default so metrics just work: 

- On `gce_instance`, you need some labels to differentiate different processes or you'll get the duplicate time series error
- In Python, Ruby, and other GIL'ed languages, it's common to use multiple processes on a single host for parallelism instead of threads. Even in a single container, it's common to do this

exporter/collector/config.go
aabmass: Talked offline, decided we are gonna go with a more structured approach with just prefix matching to being and we can add regex later if requested.

exporter/collector/config.go
aabmass: @dashpole I think we are on the same page with this now. If so, can you resolve?

exporter/collector/config.go
aabmass: @dashpole I think we are on the same page with this now. If so, can you resolve?

exporter/collector/integrationtest/testdata/fixtures/with_resource_filter_metrics_expect.json
aabmass: Done

exporter/collector/config.go
```diff
@@ -59,6 +59,20 @@ type MetricConfig struct {
 	// Buffer size for the channel which asynchronously calls CreateMetricDescriptor. Default
 	// is 10.
 	CreateMetricDescriptorBufferSize int `mapstructure:"create_metric_descriptor_buffer_size"`
+
+	// If true, the exporter will copy OTel's service.name, service.namespace, and
+	// service.instance.id resource attributes into the GCM timeseries metric labels. This
+	// option is recommended to avoid writing duplicate timeseries against the same monitored
+	// resource. Default is true.
+	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
```
dashpole, line 67: Can we unify the naming of this and instrumentation library labels?  E.g.
`instrumentation_library_labels` and `service_resource_labels`
or
`include_instrumentation_library_attributes` and `include_service_resource_attributes`

exporter/collector/config.go
```diff
@@ -59,6 +59,20 @@ type MetricConfig struct {
...
+
+	// If provided, resources attributes matching any filter will be included in metric labels.
+	ResourceFilters []ResourceFilter `mapstructure:"resource_filter"`
```
dashpole, line 70: yaml comment should be plural here: `resource_filters`

exporter/collector/config.go
```diff
@@ -59,6 +59,20 @@ type MetricConfig struct {
...
 	// is 10.
 	CreateMetricDescriptorBufferSize int `mapstructure:"create_metric_descriptor_buffer_size"`
+
+	// If true, the exporter will copy OTel's service.name, service.namespace, and
+	// service.instance.id resource attributes into the GCM timeseries metric labels. This
+	// option is recommended to avoid writing duplicate timeseries against the same monitored
+	// resource. Default is true.
+	IncludeServiceResourceAttributes bool `mapstructure:"include_service_resource_attributes"`
+
+	// If provided, resources attributes matching any filter will be included in metric labels.
```
dashpole, line 69: Can you expand the comment to say what the default value is and how it interacts with include service resource attributes? (e.g. does setting resource filter to an empty list remove the service attributes above?)

exporter/collector/config.go
aabmass: Do you have a preference between those two?

dashpole: I'd prefer `instrumentation_library_labels` and `service_resource_labels`

PR #297, opened 2022-01-31
Drop data points with start and end time too close together

dashpole:
See breaking-changes.md for the rationale for this change.

### DIFF
--- a/exporter/collector/breaking-changes.md
+++ b/exporter/collector/breaking-changes.md
@@ -52,9 +52,9 @@ behavior, please open an issue.
 
 ## Time staggering
 
-The [OTel datamodel specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resets-and-gaps) an equal start and end time on a cumulative metric indicate a reset timeseries:
+According to the [OTel datamodel specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resets-and-gaps), an equal start and end time on a cumulative metric indicate a reset timeseries:
 
-```When StartTimeUnixNano equals TimeUnixNano, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.```
+> When StartTimeUnixNano equals TimeUnixNano, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.
 
 The OpenCensus-based exporter took points with identical start and end times, and added a millisecond to the end time: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/287.  However, the resulting point makes little sense: It has a very small time period, and a normal cumulative value for an interval.  The result is that graphs display a very large rate over that millisecond, which is misleading.
 
--- a/exporter/collector/breaking-changes.md
+++ b/exporter/collector/breaking-changes.md
@@ -50,6 +50,16 @@ For now, it is not possible to customizate the mapping algorithm, beyond using t
 in the collector pipeline before the exporter. If you have a use case for customizing the
 behavior, please open an issue.
 
+## Time staggering
+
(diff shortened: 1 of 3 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -350,7 +350,11 @@ func (m *metricMapper) metricToTimeSeries(
 		sum := metric.Sum()
 		points := sum.DataPoints()
 		for i := 0; i < points.Len(); i++ {
-			ts := m.sumPointToTimeSeries(resource, extraLabels, metric, sum, points.At(i))
+			point := points.At(i)
+			if !intervalIsValid(point.StartTimestamp().AsTime(), point.Timestamp().AsTime()) {
+				continue
+			}
```
aabmass, line 356: Would it make more sense to move this logic into the corresponding `xPointToTimeSeries()` methods?

exporter/collector/breaking-changes.md
```diff
@@ -50,6 +50,16 @@ For now, it is not possible to customizate the mapping algorithm, beyond using t
...
+
+```When StartTimeUnixNano equals TimeUnixNano, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.```
+
+The OpenCensus-based exporter took points with identical start and end times, and added a millisecond to the end time: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/287.  However, the resulting point makes little sense: It has a very small time period, and a normal cumulative value for an interval.  The result is that graphs display a very large rate over that millisecond, which is misleading.
+
+This exporter instead drops points with identical start and end times.  Resets are already indicated by a change in the start timestamp for the subsequent point.
```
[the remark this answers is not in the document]
dashpole: I will do more experimentation to ensure this does what we want, or propose an alternative.

exporter/collector/metricsexporter.go
dashpole: I thought about that, but then we would need to check the return value of xPointToTimeSeries() anyways.

aabmass: Ya, another option is having them return `[]*TimeSeries` so `append()` will do the right thing. Up to you, just calling it out

exporter/collector/breaking-changes.md
```diff
@@ -50,6 +50,16 @@ For now, it is not possible to customizate the mapping algorithm, beyond using t
 in the collector pipeline before the exporter. If you have a use case for customizing the
 behavior, please open an issue.
 
+## Time staggering
+
+The [OTel datamodel specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resets-and-gaps) an equal start and end time on a cumulative metric indicate a reset timeseries:
```
punya, line 55:
```suggestion
According to the [OTel datamodel specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resets-and-gaps), an equal start and end time on a cumulative metric indicate a reset timeseries:
```

exporter/collector/breaking-changes.md
```diff
@@ -50,6 +50,16 @@ For now, it is not possible to customizate the mapping algorithm, beyond using t
...
+
+```When StartTimeUnixNano equals TimeUnixNano, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.```
```
punya, line 57:
```suggestion
> When StartTimeUnixNano equals TimeUnixNano, a new unbroken sequence of observations begins with a reset at an unknown start time. The initial observed value is recorded to indicate that an unbroken sequence of observations resumes. These points have zero duration, and indicate that nothing is known about previously-reported points and that data may have been lost.
```

punya (approved):

exporter/collector/metricsexporter.go
dashpole: done

exporter/collector/breaking-changes.md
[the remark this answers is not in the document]
jsuereth: `normalize_sums_processor` which takes the first point, remembers its value and timestamp, then just reports cumulative diffs from that point.

I think we may want to add that feature directly into the exporter (we can discuss in chat).

PR #298, opened 2022-01-31
Split client configuration per-signal

dashpole:
Endpoints for cloud monitoring are different per-signal.  For example, I would want to use `monitoring.googleapis.com` for metrics, and `tracing.googlapis.com` for traces.  Overriding both signals with the same endpoint makes little sense, since it would only work for one signal.

Also, resource mappings should be nested inside monitoring, since it only applies to metrics.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -28,23 +28,33 @@ const (
 type Config struct {
 	ProjectID string `mapstructure:"project"`
 	UserAgent string `mapstructure:"user_agent"`
-	Endpoint  string `mapstructure:"endpoint"`
+
+	TraceConfig TraceConfig `mapstructure:"trace"`
+
+	MetricConfig MetricConfig `mapstructure:"metric"`
+}
+
+type ClientConfig struct {
+	Endpoint string `mapstructure:"endpoint"`
 	// Only has effect if Endpoint is not ""
 	UseInsecure bool `mapstructure:"use_insecure"`
 
-	ResourceMappings []ResourceMapping `mapstructure:"resource_mappings"`
 	// GetClientOptions returns additional options to be passed
 	// to the underlying Google Cloud API client.
 	// Must be set programmatically (no support via declarative config).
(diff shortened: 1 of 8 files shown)


exporter/collector/config.go
```diff
@@ -28,23 +28,33 @@ const (
 type Config struct {
 	ProjectID string `mapstructure:"project"`
 	UserAgent string `mapstructure:"user_agent"`
-	Endpoint  string `mapstructure:"endpoint"`
```
aabmass, line 31: Just calling out, I think this is the first breaking change to trace portion from our exporter rewrite. Should we keep this for backward compat and mark it deprecated or are we ok making this?

exporter/collector/config.go
```diff
@@ -28,23 +28,33 @@ const (
 type Config struct {
 	ProjectID string `mapstructure:"project"`
 	UserAgent string `mapstructure:"user_agent"`
```
aabmass, line 30: Should we move this into the ClientConfig as well?

dashpole: Good question.  I initially moved it into different ones for trace/metrics, since it is part of the config for clients.  I haven't found a concrete definition for what should be put in user agent though... It seems to usually be defined as the "software" that is making the request, and I consider the exporter to be the same piece of "software", but we could definitely split it as well.

exporter/collector/config.go
dashpole: That is correct.  I've thought about how to handle that, but didn't write it here:

In the upstream collector, we should leave items that we removed here in the configuration.  We should still pass them to the opencensus exporter.  But when the feature gate is enabled, we should thrown an error if a user tries to set them.  Once the feature gate is GA, we can safely rip out the old exporter and old config options.

jsuereth:
> Also, resource mappings should be nested inside monitoring, since it only applies to metrics.

We should support logging resource in the (very near) future.

PR #299, opened 2022-02-01
Reenable disabled metrics test

dashpole:
This fixes a TODO.

There is one change in behavior that I want to double-check is intentional:  The MetricDescriptor.Name used to be the format:  `projects/<project>/metricDescriptors/<prefix>/<metric name>`.  Now, the MetricDescriptor.Name appears to be just <metric name>.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq ($(UNAME_S),Darwin)
 	endif
 endif
 
-GOTEST_MIN = go test -v -timeout 60s
+GOTEST_MIN = go test -v -timeout 70s
 GOTEST = $(GOTEST_MIN) -race
 GOTEST_WITH_COVERAGE = $(GOTEST) -coverprofile=coverage.txt -covermode=atomic
 
--- a/exporter/collector/googlecloud_test.go
+++ b/exporter/collector/googlecloud_test.go
@@ -148,8 +148,6 @@ func (ms *mockMetricServer) CreateTimeSeries(ctx context.Context, req *cloudmoni
 }
 
 func TestGoogleCloudMetricExport(t *testing.T) {
-	// TODO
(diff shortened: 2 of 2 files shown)


aabmass:
> The MetricDescriptor.Name used to be the format: projects/<project>/metricDescriptors/<prefix>/<metric name>. Now, the MetricDescriptor.Name appears to be just .

The integration tests are passing against GCM, so I guess it is automatically set. I just tried manually testing against the API with the same result.

However, I'm not sure if this change was intentional. @jsuereth wdyt?

aabmass (approved):

aabmass:
It actually seems to completely ignore it:

request:
```json
{
  "displayName": "My metric",
  "name": "ehlllwdl",
  "type": "custom.googleapis.com/bar/foo",
  "labels": [
    {
      "key": "bax",
      "description": "my bax"
    }
  ],
  "metricKind": "CUMULATIVE",
  "valueType": "INT64"
}
```

response:

```json
{
  "name": "projects/otel-starter-project/metricDescriptors/custom.googleapis.com/bar/foo",
  "labels": [
    {
      "key": "bax",
      "description": "my bax"
    }
  ],
  "metricKind": "CUMULATIVE",
  "valueType": "INT64",
  "displayName": "My metric",
  "type": "custom.googleapis.com/bar/foo"
}
```

punya:
According to https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#:~:text=The%20resource%20name%20of%20the%20metric%20descriptor., the `name` field is supposed to mean the "resource name" (in fact, http://shortn/_HH9psvkVBe says it's the project name). I'm guessing it's ignored and defaulted if it looks nonsensical.

punya (approved):

PR #300, opened 2022-02-01
Remove Known Domains

dashpole:
`known_domains` was originally added in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/247 because we wanted to support attaching a prefix only to some metrics.  However, on other topics (create service timeseries, create metric descriptor, etc), we decided to make behavior explicitly configurable instead of implicit based on metric domain.

I think we should consider making metric prefixing also an explicitly-configurable behavior, which would mean there is no domain-specific behavior in the collector exporter.  Prefix = "" would disable attaching prefixes.

The only question I have is: Are there any use-cases we know of where we need to be able to mix metrics with/without a domain?

Marking as draft until we have had a chance to discuss.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -43,10 +43,11 @@ type Config struct {
 }
 
 type MetricConfig struct {
+	// Prefix is attached to all metrics sent by this exporter.
+	// Disable prefixes by setting it to "".
+	// Defaults to "workload.googleapis.com".
 	Prefix                     string `mapstructure:"prefix"`
 	SkipCreateMetricDescriptor bool   `mapstructure:"skip_create_descriptor"`
-	// If a metric belongs to one of these domains it does not get a prefix.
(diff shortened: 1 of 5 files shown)


jsuereth:
> The only question I have is: Are there any use-cases we know of where we need to be able to mix metrics with/without a domain?

Yeah, I think we shouldn't break this use-case.  I believe Ops-Agent relies (or relied) on it.

For the full scenario:

1. We have Ops agent reporting custom metrics via an OTLP ingestion endpoint.  If users are not sending specific metric names, we should be appending workload.googleapis.com/ to them.
2. We have Ops Agent reporting its own agent metrics (agent.googleapis.com)
3. We have Ops Agent reporting 3P integration metrics (these may already be fully namespaced).


While we *could* force everyone to namespace via a processor, you know my stance on forcing processor usage for our exporter to work correctly.

dashpole:
Ok, happy to leave this as-is then.

PR #301, opened 2022-02-01
Make summaries more prometheus-friendly

dashpole:
This PR changes the default behavior for summaries in the googlecloud exporter to match prometheus metric naming and semantics.  Summary metrics are now split into metrics with suffixes `_sum`, `_count`, ``, with the quantile metric having a `quantile` label with values in 0 < 1.

This adds a new config option: `summary_percentiles`, which reverts behavior to the previous naming for backwards-compatibility.  To switch to the previous behavior:

```yaml
googlecloud:
  metric:
    summary_percentiles: true
```

### DIFF
--- a/exporter/collector/breaking-changes.md
+++ b/exporter/collector/breaking-changes.md
@@ -26,6 +26,16 @@ domain name of the metric type.  E.g. if a metric called
 `workload.googleapis.com/nginx/latency` is created, the display name will
 be `nginx/latency` instead of `workload.googleapis.com/nginx/latency`.
 
+## Summaries
+
+The previous exporter would split summary metrics into `_summary_count`, `_summary_sum`, `_summary_percentile`, and the percentiles metric had the label `percentile` with values in 0 < 100. The new exporter splits summary metrics into `_count`, `_sum`, no suffix. The metric without a suffix has a `quantile`, with values in 0 < 1. This change was done to match the semantics of prometheus summary metrics, which users are more familiar with. The `summary_percentiles` option reverts to the previous behavior:
+
+```yaml
+googlecloud:
+  metric:
+    summary_percentiles: true
+```
+
 ## Monitored Resources
 
 Mapping from OTel Resource to GCM monitored resource has been completely changed. The OC based
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -83,9 +83,11 @@ type metricMapper struct {
 
 // Constants we use when translating summary metrics into GCP.
 const (
-	SummaryCountPrefix      = "_summary_count"
-	SummarySumSuffix        = "_summary_sum"
-	SummaryPercentileSuffix = "_summary_percentile"
+	SummaryCountPrefix            = "_count"
+	SummarySumSuffix              = "_sum"
+	LegacySummaryCountPrefix      = "_summary_count"
+	LegacySummarySumSuffix        = "_summary_sum"
+	LegacySummaryPercentileSuffix = "_summary_percentile"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


exporter/collector/breaking-changes.md
```diff
@@ -26,6 +26,16 @@ domain name of the metric type.  E.g. if a metric called
...
 be `nginx/latency` instead of `workload.googleapis.com/nginx/latency`.
 
+## Summaries
+
+The previous exporter would split summary metrics into `_summary_count`, `_summary_sum`, `_summary_percentile`, and the percentiles metric had the label `percentile` with values in 0 < 100. The new exporter splits summary metrics into `_count`, `_sum`, no suffix. The metric without a suffix has a `quantile`, with values in 0 < 1. This change was done to match the semantics of prometheus summary metrics, which users are more familiar with. The `summary_percentiles` option reverts to the previous behavior:
+
+```yaml
+googlecloud:
+  metric:
+    summary_percentiles: true
```
aabmass, line 36: Do we have first party use cases for this option, or is it just for external users who wish to keep the old behavior?

exporter/collector/metricsexporter.go
```diff
@@ -448,9 +450,17 @@ func (m *metricMapper) summaryPointToTimeSeries(
...
+		if m.cfg.MetricConfig.SummaryPercentiles {
+			pLabel = labels{
+				// Convert to percentile.
+				"percentile": fmt.Sprintf("%f", quantile.Quantile()*100),
+			}
+		} else {
+			pLabel = labels{
+				// Convert to quantile.
+				"quantile": fmt.Sprintf("%f", quantile.Quantile()),
+			}
```
punya, line 463: Nit: This function is pretty long already, would you consider extracting this logic out into a small utility?

exporter/collector/metricsexporter.go
```diff
@@ -878,16 +888,31 @@ func (m *metricMapper) labelDescriptors(pm pdata.Metric) []*label.LabelDescripto
 }
 
 // Returns (sum, count, percentile) metric names for a summary metric.
-func summaryMetricNames(name string) (string, string, string) {
+func (m *metricMapper) summaryMetricNames(name string) (string, string, string) {
```
punya, line 891: Nit: I worry that returning three strings from such a function makes it easy to mix up the order. I'd prefer a small struct even though it's a bit more verbose.

(I see it was already this way before the current PR, so feel free to fix this in a separate PR if you agree with the feedback.)

exporter/collector/metricsexporter.go
```diff
@@ -878,16 +888,31 @@ func (m *metricMapper) labelDescriptors(pm pdata.Metric) []*label.LabelDescripto
...
 func (m *metricMapper) summaryMetricDescriptors(pm pdata.Metric) []*metricpb.MetricDescriptor {
-	sumName, countName, percentileName := summaryMetricNames(pm.Name())
+	sumName, countName, percentileName := m.summaryMetricNames(pm.Name())
 	labels := m.labelDescriptors(pm)
+	quantileDescriptor := &label.LabelDescriptor{
+		Key:         "quantile",
+		Description: "the value at a given quantile of a distribution",
+	}
+	if m.cfg.MetricConfig.SummaryPercentiles {
+		quantileDescriptor = &label.LabelDescriptor{
```
punya, line 911: Nit: while it's clear enough in the context of this PR, I think a future reader would be confused that `quantileDescriptor` is actually about the percentile metric.

punya (approved):
I added some nitpicky comments, but I'm approving regardless of whether you choose to address them.

exporter/collector/breaking-changes.md
dashpole: This is just for external users.  There is one anthos metric with this naming (cc @catherinef-dev), but they seem to want the new behavior.

exporter/collector/metricsexporter.go
dashpole: renamed

exporter/collector/metricsexporter.go
dashpole: added a TODO

exporter/collector/metricsexporter.go
dashpole: Done

exporter/collector/breaking-changes.md
dashpole: Since this breaks naming of their metrics, I was thinking we would need to support the old names longer-term.  Do you think we should just change naming without a config option?

jsuereth: Given other breaking changes (and that we're putting this behind a feature gate), we can break the naming here.

jsuereth (approved):
I'm fine with this as-is *or* with removing the old `_summary_*` naming convention.

PR #303, opened 2022-02-02
Batching docs

tbarker25:
Update collector docs to remove recommendation of using the batch processor
    
    BUG=216620312

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -44,7 +44,6 @@ These instructions are to get you up and running quickly with the GCP exporter i
     processors:
       memory_limiter:
       batch:
-        # Google Cloud Monitoring limits batches to 200 metric points.
         send_batch_max_size: 200
         send_batch_size: 200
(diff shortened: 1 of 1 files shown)


exporter/collector/README.md
```diff
@@ -43,19 +43,15 @@ These instructions are to get you up and running quickly with the GCP exporter i
...
       memory_limiter:
-      batch:
-        # Google Cloud Monitoring limits batches to 200 metric points.
-        send_batch_max_size: 200
-        send_batch_size: 200
     service:
       pipelines:
         traces:
           receivers: [otlp]
-          processors: [memory_limiter, batch]
```
dashpole, line 54: We only do batching for metrics, right?  Should we still recommend it for traces?

tbarker25: Good point. I've changed this to only recommending it for traces.

dashpole (approved):

PR #307, opened 2022-02-02
Support ResourceMapping in the new collector exporter

dashpole:
Third attempt at #302.


exporter/collector/monitoredresource.go
```diff
@@ -123,37 +130,37 @@ func (m *metricMapper) resourceMetricsToMonitoredResource(
...
 		_, hasServiceName := attrs.Get(semconv.AttributeServiceName)
 		_, hasServiceInstanceID := attrs.Get(semconv.AttributeServiceInstanceID)
 		if hasServiceName && hasServiceInstanceID {
-			return createMonitoredResource(genericTask, attrs)
+			return m.createMonitoredResource(genericTask, attrs)
 		}
 
 		// If not possible, fallback to generic_node
-		return createMonitoredResource(genericNode, attrs)
+		return m.createMonitoredResource(genericNode, attrs)
```
aabmass, line 155: Config changes LGTM. It will allow config to override the source labels, but there is still no way for them to override the monitored resource type selection in the first place (hence the changes to GKE control plane fixture didn't work).

Any plans for how to do this?

aabmass: I could be wrong about the fixture. But I think there still needs to be a way to override the MR type selection

exporter/collector/monitoredresource.go
```diff
@@ -123,37 +130,37 @@ func (m *metricMapper) resourceMetricsToMonitoredResource(
 
 	switch cloudPlatform {
```
jsuereth, line 131: I think @aabmass talked to you about this in standup.

I think you'll want resource-mapping config to allow altering this logic (the swtich on cloudPlatform and other related switches).

exporter/collector/config.go
```diff
@@ -72,8 +72,8 @@ type MetricConfig struct {
 }
 
 // ResourceMapping defines mapping of resources from source (OpenCensus) to target (Google Cloud).
+// In order to use these resource mappings, the cloud.platform attribute must be set to gcp_custom.
```
dashpole, line 75: This now requires the cloud.platform to be gcp_custom.  An alternative would be to add a cloud_platform field in ResourceMapping that allows telling the exporter which cloud platform to apply it to.

exporter/collector/monitoredresource.go
```diff
@@ -137,28 +146,56 @@ func (m *metricMapper) resourceMetricsToMonitoredResource(
...
+		mrLabels, foundAll := mrLabelsForMapping(mapping, resourceAttrs)
+		if foundAll {
+			// TODO handle extra labels
+			return &monitoredrespb.MonitoredResource{
+				Type:   monitoredResourceType,
+				Labels: mrLabels,
+			}, labels{}
+		}
 	}
+	return fallbackMonitoredResource(resourceAttrs)
```
dashpole, line 177: curious what we should do if it doesn't match any of the custom mappings... I've written it to fall back to generic_task/generic_node.

exporter/collector/monitoredresource.go
dashpole: I fixed it.  We can discuss design below.

exporter/collector/monitoredresource.go
dashpole: Yep, thats what I ended up needing.  If you have feedback on the design, you can add it below.

exporter/collector/monitoredresource.go
aabmass: It might be better to put empty strings for the missing MR labels and let any errors surface from the CreateTimeSeries call. That's what we do for the others.

exporter/collector/config.go
aabmass: I prefer the second option so we don't need a magic string

exporter/collector/monitoredresource.go
dashpole: What if someone adds multiple MR for one platform?  We can't really add the detection/fallback mechanisms like we do with k8s_container.  Which MR do we default to if it doesn't match either?

dashpole:
This will need to be rebased on https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/317.  I'll mark this as a draft until we've settled on a design.

dashpole:
Closing for now.

PR #309, opened 2022-02-02
Split client configuration per-endpoint

dashpole:
Second attempt at https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/298

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -28,23 +28,33 @@ const (
 type Config struct {
 	ProjectID string `mapstructure:"project"`
 	UserAgent string `mapstructure:"user_agent"`
-	Endpoint  string `mapstructure:"endpoint"`
+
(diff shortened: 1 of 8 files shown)


exporter/collector/config.go
```diff
@@ -28,23 +28,33 @@ const (
 type Config struct {
 	ProjectID string `mapstructure:"project"`
 	UserAgent string `mapstructure:"user_agent"`
-	Endpoint  string `mapstructure:"endpoint"`
```
aabmass, line 31: Saw your reply on the older PR. IIUC, if we are using the new exporter it will throw an error if this option is supplied. If we are using the old one, it will use the old config struct so should be fine. 

Is that right?

aabmass (approved):

exporter/collector/config.go
dashpole: Yes

PR #311, opened 2022-02-02
Drop data points with start and end time too close together

dashpole:
Second attempt at #297.

We still need to normalize cumulative points based on the first observed value, but that can be done as a follow-up.


dashpole:
We are going to drop these points during normalization

PR #314, opened 2022-02-03
Address a bunch of small TODOs

dashpole:
They are either trivial to fix, or already done elsewhere.  See commit message for reason for removing each TODO

### DIFF
--- a/exporter/collector/googlecloud.go
+++ b/exporter/collector/googlecloud.go
@@ -18,14 +18,17 @@ package collector
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
 
+	traceapi "cloud.google.com/go/trace/apiv2"
 	"go.opencensus.io/plugin/ocgrpc"
 	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/collector/model/pdata"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"golang.org/x/oauth2/google"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
@@ -44,6 +47,20 @@ func (te *TraceExporter) Shutdown(ctx context.Context) error {
 
 func setVersionInUserAgent(cfg *Config, version string) {
 	cfg.UserAgent = strings.ReplaceAll(cfg.UserAgent, "{{version}}", version)
+}
+
+func setProjectFromADC(ctx context.Context, cfg *Config, scopes []string) error {
+	if cfg.ProjectID == "" {
+		creds, err := google.FindDefaultCredentials(ctx, scopes...)
+		if err != nil {
+			return fmt.Errorf("error finding default application credentials: %v", err)
+		}
+		if creds.ProjectID == "" {
+			return errors.New("no project found with application default credentials")
+		}
+		cfg.ProjectID = creds.ProjectID
+	}
+	return nil
 }
 
 func generateClientOptions(cfg *ClientConfig, userAgent string) ([]option.ClientOption, error) {
@@ -78,9 +95,10 @@ func generateClientOptions(cfg *ClientConfig, userAgent string) ([]option.Client
 	return copts, nil
 }
 
-func NewGoogleCloudTracesExporter(cfg Config, version string, timeout time.Duration) (*TraceExporter, error) {
+func NewGoogleCloudTracesExporter(ctx context.Context, cfg Config, version string, timeout time.Duration) (*TraceExporter, error) {
 	view.Register(ocgrpc.DefaultClientViews...)
 	setVersionInUserAgent(&cfg, version)
+	setProjectFromADC(ctx, &cfg, traceapi.DefaultAuthScopes())
 
 	topts := []cloudtrace.Option{
 		cloudtrace.WithProjectID(cfg.ProjectID),
--- a/exporter/collector/metricsexporter.go
+++ b/exporter/collector/metricsexporter.go
@@ -19,7 +19,6 @@ package collector
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math"
 	"net/url"
@@ -34,7 +33,6 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/collector/model/pdata"
 	"go.uber.org/zap"
-	"golang.org/x/oauth2/google"
 	"google.golang.org/genproto/googleapis/api/distribution"
 	"google.golang.org/genproto/googleapis/api/label"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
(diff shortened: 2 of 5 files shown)


exporter/collector/googlecloud.go
```diff
@@ -45,6 +48,20 @@ func setVersionInUserAgent(cfg *Config, version string) {
 	cfg.UserAgent = strings.ReplaceAll(cfg.UserAgent, "{{version}}", version)
 }
 
+func setProjectFromADC(ctx context.Context, cfg *Config, scopes []string) error {
+	if cfg.ProjectID == "" {
+		creds, err := google.FindDefaultCredentials(ctx, scopes...)
+		if err != nil {
+			return fmt.Errorf("error finding default application credentials: %v", err)
```
aabmass, line 55:
```suggestion
			return fmt.Errorf("error finding application default credentials: %v", err)
```

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
```diff
@@ -3150,6 +3150,486 @@
       ]
     }
   ],
+  "createMetricDescriptorRequests": [
```
aabmass, line 3153: @jsuereth does this look OK for ops agent to send these descriptors?

exporter/collector/metricsexporter.go
```diff
@@ -381,7 +363,6 @@ func (m *metricMapper) metricToTimeSeries(
 			ts := m.exponentialHistogramToTimeSeries(resource, extraLabels, metric, eh, points.At(i))
 			timeSeries = append(timeSeries, ts)
 		}
-	// TODO: add cases for other metric data types
 	default:
 		// TODO: log unsupported metric
```
aabmass, line 367: This should be a small fix, or should we open a bug for this one?

aabmass (approved):

exporter/collector/metricsexporter.go
dashpole: There aren't any other data types, except DataTypeNone (which isn't a real data-type, just the default value for datatype)

aabmass: But more could be added to the protocol ([this oneof](https://github.com/open-telemetry/opentelemetry-proto/blob/c444081593837fe770c825a83c864de736a768b4/opentelemetry/proto/metrics/v1/metrics.proto#L176)) and old collectors should at least warn that they can't be handled. Or does the collector strip these out before they reach the exporter?

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
dashpole: Ops agent can still disable CMD.  Maybe i'm misinterpreting the TODO, but I thought we just wanted to test with CMD enabled?

exporter/collector/metricsexporter.go
dashpole: That would be handled by the default case below, right?

aabmass: Ya, shouldn't we log something though rather than silently dropping?

exporter/collector/googlecloud.go
```diff
@@ -45,6 +48,20 @@ func setVersionInUserAgent(cfg *Config, version string) {
 	cfg.UserAgent = strings.ReplaceAll(cfg.UserAgent, "{{version}}", version)
 }
 
+func setProjectFromADC(ctx context.Context, cfg *Config, scopes []string) error {
```
jsuereth, line 51: How many places do we have this code now :)

exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics_expect.json
```diff
@@ -83,6 +108,29 @@
               }
             ]
           },
+          {
+            "metric": {
+              "type": "custom.googleapis.com/opencensus/grpc.io/client/completed_rpcs",
```
jsuereth, line 113: I realize this is just the nature of how we did our self observabiltiy metrics, but it's highly confusing seeing the opencensus metrics here when Ops Agent defines its own self-observability metrics, that this test is measuring if we can write those self-observability metrics.

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
jsuereth: Yes we want a test with it enabled.

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
```diff
@@ -3150,6 +3150,486 @@
...
+        "unit": "By",
+        "description": "Disk bytes transferred.",
+        "displayName": "processes/disk/write_bytes_count"
+      }
+    },
+    {
+      "metricDescriptor": {
+        "type": "agent.googleapis.com/swap/percent_used",
+        "metricKind": "GAUGE",
+        "unit": "By",
```
jsuereth, line 3320: FYI - this unit is wrong.  It's also wrong in the underlying data.  We may want to re-capture ops agent host metrics and see if we can pull the version where this unit is fixed before we turn on CMD for ops-agent.

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
jsuereth: Oh, actually, I'm dumb.

`agent.googleapis.com` should never send metric descriptors (they're defined like platform metrics IIRC).

We want CMD turned on for any workload metrics.....  Maybe check with Ling or Sophie.

jsuereth (changes requested):

exporter/collector/googlecloud.go
dashpole: Just one in the collector exporter, which is now here.

exporter/collector/integrationtest/testdata/fixtures/ops_agent_host_metrics_expect.json
dashpole: Removed the TODO to update agent.googleapis.com tests, and added a TODO to add new integration tests for workload.googleapis.com from ops-agent

exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics_expect.json
aabmass: Ya since this uses OpenCensus instrumentation we sort of have to use the OC exporter to capture them. Should we:
1. Tweak the metrics prefix config for the OC exporter or
2. Strip out the whole `custom.googleapis.com/opencensus/` prefix when saving the fixture

dashpole:
@jsuereth when you get the chance, can you re-review?

jsuereth (approved):

PR #318, opened 2022-02-07
Use google CEL for custom resource mappings

aabmass:


ItsLastDay:
This looks like a great feature! Why wasn't it implemented in the end?

aabmass:
This was intended to be a PoC and we decided we wanted to hear more use cases for this kind flexibility before including it. You can still pass a Go function to configure resource mappings ([example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6f0cc7a7be2b22d419d4c6dd1bf5a059df978a8c/exporter/googlemanagedprometheusexporter/config.go#L62)) but you'll have to build your own collector (we figure 1st party use cases are already doing this).

Another option would be Starlark which would probably be less clunky. IIRC, this CEL approach was also quite slow in benchmarks.

PR #324, opened 2022-02-14
Add attribute mapping function

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/221

This would allow users to set the WithAttributeKeyMapping option to configure how keys are written to cloud trace.  We currently do some translation (such as converting `service.name` to `g.co/gae/app/module`, which is used as "Service" in the cloud trace UI.  This does not change the default behavior.


jsuereth:
Should there be a yaml-config option here for the collector exporter?

dashpole:
Hmm... Good point.  This is just for the go SDK exporter, but i'd assume users would want the same in the collector.

dashpole:
Added config in the collector to set the attribute mapping.

dashpole:
@tyrone-anz does this meet your needs?  This lets you choose how to map each attribute key, but doesn't let you double-write an attribute (you can't set both representations of service.name, for example)

dashpole:
You could definitely use that workaround.  Eventually, we hope service.name can be displayed prominently in the UI (like g.co/gae/app/module is), so you shouldn't need the duplicate.

exporter/collector/config.go
```diff
@@ -48,6 +48,17 @@ type ClientConfig struct {
 
 type TraceConfig struct {
 	ClientConfig ClientConfig `mapstructure:",squash"`
+	// AttributeKeyMappings determines how to map from OpenTelemetry attribute keys to
+	// Google Cloud Trace keys.  By default, it changes keys so that they appear more
+	// prominently in the UI.
+	AttributeKeyMappings []AttributeKeyMapping `mapstructure:"attribute_mappings"`
```
punya, line 54: Nit: can the mapstructure name be consistent with the type and field names? E.g., `AttributeMappings`, `AttributeMapping`, `attribute_mappings`.

exporter/collector/traces.go
```diff
@@ -119,6 +123,22 @@ func NewGoogleCloudTracesExporter(ctx context.Context, cfg Config, version strin
 	return &TraceExporter{texporter: exp}, nil
 }
 
+func mappingFuncFromAKM(akm []AttributeKeyMapping) func(attribute.Key) attribute.Key {
+	// convert list to map for easy lookups
+	mapFromConfig := map[string]string{}
```
punya, line 128:
```suggestion
	mapFromConfig := make(map[string]string, len(akm))
```

exporter/collector/traces.go
```diff
@@ -119,6 +123,22 @@ func NewGoogleCloudTracesExporter(ctx context.Context, cfg Config, version strin
...
+	for _, mapping := range akm {
+		mapFromConfig[mapping.Key] = mapping.Replacement
+	}
+	return func(input attribute.Key) attribute.Key {
+		// if a replacement was specified in the config, use it.
+		if replacement, ok := mapFromConfig[string(input)]; ok {
+			return attribute.Key(replacement)
+		}
+		// otherwise, leave the attribute as-is
+		return attribute.Key(input)
```
punya, line 138:
```suggestion
		return input
```

exporter/collector/traces.go
```diff
@@ -119,6 +123,22 @@ func NewGoogleCloudTracesExporter(ctx context.Context, cfg Config, version strin
 	return &TraceExporter{texporter: exp}, nil
 }
 
+func mappingFuncFromAKM(akm []AttributeKeyMapping) func(attribute.Key) attribute.Key {
```
punya, line 126:
```suggestion
func (akm []AttributeKeyMapping) mapKey(attribute.Key) attribute.Key {
```

It's a bit more readable here as well as at initialization time.

exporter/collector/config.go
punya, line 54: Why do we use a list of pairs representation, when a map would be more concise and document as well as ensure that keys are distinct?

dashpole: We could switch to maps.  The current way this would look in config is:

```yaml
attribute_mappings:
- key: service.name
  replacement: different.service.key
- key: foo
  replacement: bar
```

If we used map[string]string it would look like:

```yaml
attribute_mappings:
  service.name: different.service.key
  foo: bar
```

But that doesn't allow the API to evolve, since you can't add additional fields to the value.

We could allow evolution with a sub-object if we used map[string]struct{replacement: string}, and it would look like:

```yaml
attribute_mappings:
  service.name:
    replacement: different.service.key
  foo:
    replacement: bar
```

Of the three, I found the first to be the easiest to read (but I can see your case for the second).  You are right that I should be validating the uniqueness of the keys.  Right now the behavior is that the last key takes precedence, but returning an error to the user would be better.

dashpole: renamed

punya: Supporting evolution is a great reason to rule out `map[string]string`, thanks for calling that out. I'm agnostic between list-of-pairs and map-with-struct-values; though I'd request some validation of keys if we go with the list representation.

punya (approved):

exporter/collector/config.go
dashpole: added validation

PR #325, opened 2022-02-17
Fix formatting of summary quantile labels to match prometheus

dashpole:
This trims trailing zeroes.  Thanks @quentinmit for pointing this out.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/batching_metrics_expect.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/batching_metrics_expect.json
@@ -67,7 +67,7 @@
             "type": "workload.googleapis.com/testsummary 1",
             "labels": {
               "foo": "bar",
-              "quantile": "0.500000"
+              "quantile": "0.5"
             }
           },
           "resource": {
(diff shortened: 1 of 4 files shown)


exporter/collector/metricsexporter.go
```diff
@@ -430,7 +431,7 @@ func (m *metricMapper) summaryPointToTimeSeries(
 	for i := 0; i < quantiles.Len(); i++ {
 		quantile := quantiles.At(i)
 		pLabel := labels{
-			"quantile": fmt.Sprintf("%f", quantile.Quantile()),
+			"quantile": strconv.FormatFloat(quantile.Quantile(), 'f', -1, 64),
```
quentinmit, line 434: Are there any conventions from Prometheus land that we should make sure we're following here? (e.g. should 0 be reported as `0` or `0.0` or something else? Ditto for 1) I think as-is this will report 0 as `0` and 1 as `1`.

dashpole: It reports them a `0` and `1`.  See these test-cases: https://github.com/prometheus/prometheus/blob/4c56a193c518ae6f56008b0a4c850a9c3f1477c6/model/textparse/promtestdata.txt#L3

dashpole: So I think we are good on that front

PR #333, opened 2022-03-24
Initial setup for logs collector exporter

damemi:
This sets up the basic functions and interface implementation for the logging exporter

Part of b/221852803

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -31,7 +31,7 @@ func NewGoogleCloudLogsExporter(ctx context.Context, cfg Config) (*LogsExporter,
 	if err != nil {
 		return nil, err
 	}
-	logger := client.Logger("my-log") // TODO(@damemi): detect log name
+	logger := client.Logger("my-log") // TODO(@damemi) detect log name
 	return &LogsExporter{
 		client: client,
 		logger: logger,
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -1,0 +1,62 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"context"
+
+	"cloud.google.com/go/logging"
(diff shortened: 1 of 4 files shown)


exporter/collector/logs.go
```diff
@@ -0,0 +1,50 @@
...
+	"go.opentelemetry.io/collector/model/pdata"
+)
+
+type LogsExporter struct {
+	client *logging.Client
+	logger *logging.Logger
+}
+
+func NewGoogleCloudLogsExporter(ctx context.Context, cfg Config) (*LogsExporter, error) {
+	client, err := logging.NewClient(ctx, "mikedame-joonix")
```
dashpole, line 17: It looks like this should be of the form `projects/<project id>`: https://github.com/googleapis/google-cloud-go/blob/main/logging/logging.go#L126.  Can you add a TODO to detect the project id as well?

exporter/collector/logs.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return nil, err
+	}
+	logger := client.Logger("my-log")
```
dashpole, line 21: Add a TODO to update the name used.

exporter/collector/logs.go
```diff
@@ -0,0 +1,50 @@
...
+			for k := 0; k < ill.LogRecords().Len(); k++ {
+				log := ill.LogRecords().At(k)
+				l.logger.Log(logging.Entry{
+					Resource: mr,
+					Payload:  log.Body().AsString(),
+				})
+			}
+		}
+	}
+	fmt.Printf("Pdata: %+v\n\n", ld)
```
dashpole, line 48: remove this.

exporter/collector/logs.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (l *LogsExporter) Shutdown(ctx context.Context) error {
+	return l.client.Close()
+}
+
+func (l *LogsExporter) PushLogs(ctx context.Context, ld pdata.Logs) error {
+	for i := 0; i < ld.ResourceLogs().Len(); i++ {
+		rl := ld.ResourceLogs().At(i)
+		mapper := &metricMapper{} // Refactor metricMapper to map MRs for logging?
```
punya, line 48: Can we instantiate this just once outside the loop?

punya (commented):
I'm assuming that we'll flesh out the mapping to `logging.Entry` in subsequent PRs.

Will unit tests for this code also come in later PRs?

exporter/collector/logs.go
damemi: ideally this will be a field of `logsExporter` (how it's used for metrics right now) and won't need to be instantianted in `PushLogs` at all

damemi:
> I'm assuming that we'll flesh out the mapping to `logging.Entry` in subsequent PRs.
> 
> Will unit tests for this code also come in later PRs?

Yes to both :) This is just to establish a branch that we can start working off of

punya (approved):

PR #336, opened 2022-03-28
Do not use IsZero on timestamps obtained from pdata

dashpole:
Context: https://github.com/open-telemetry/opentelemetry-collector/issues/5102.  It isn't possible to round-trip timestamps before 1970 in the pdata timestamp package.  An unset timestamp would be Timestamp(0), which returns false for AsTime().IsZero()

### DIFF
--- a/exporter/collector/metrics_test.go
+++ b/exporter/collector/metrics_test.go
@@ -35,7 +35,7 @@ import (
 )
 
 var (
-	start = time.Unix(0, 0)
+	start = time.Unix(1000, 1000)
 )
 
 func newTestMetricMapper() (metricMapper, func()) {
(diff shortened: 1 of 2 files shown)


exporter/collector/metrics_test.go
```diff
@@ -586,6 +586,7 @@ func TestSummaryPointToTimeSeries(t *testing.T) {
 	quantile.SetQuantile(1.0)
 	quantile.SetValue(1.0)
 	end := start.Add(time.Hour)
+	point.SetStartTimestamp(pdata.NewTimestampFromTime(start))
```
damemi, line 589: should we add a test for the 0 value conversion?

dashpole: Zero is the default value, so it was being implicitly used in one of the tests.  I made that usage explicit in the most recent commit.

damemi: Ah, sounds good to me. Think I get it now

damemi (approved):

PR #339, opened 2022-03-31
Drop data points with no value set

dashpole:
For now, I think this is only used for prometheus staleness markers.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/workload_metrics_expect.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/workload_metrics_expect.json
@@ -368,36 +368,6 @@
         },
         {
           "metric": {
-            "type": "workload.googleapis.com/rabbitmq_channel_acks_uncommitted",
-            "labels": {
-              "job": "default/rabbitmq/0"
-            }
-          },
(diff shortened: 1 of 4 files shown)


exporter/collector/integrationtest/testdata/fixtures/workload_metrics_expect.json
```diff
@@ -6454,7 +6424,7 @@
                   "startTime": "1970-01-01T00:00:00Z"
                 },
                 "value": {
-                  "int64Value": "203"
```
dashpole, line 6457: Note: this is a change in the self-obs metrics, since we now send 202 metrics instead of 203 metrics.

exporter/collector/integrationtest/testdata/fixtures/workload_metrics_expect.json
```diff
@@ -6326,11 +6296,7 @@
               }
             }
           ]
-        }
-      ]
-    },
-    {
```
dashpole, line 6332: note: because we dropped one of the metrics above, the batch cutoff is now one point later.

damemi (approved):

PR #340, opened 2022-03-31
Handle NaN sums in histograms and exponential histograms

dashpole:
Make sure we leave the mean at 0 for Nan sums in histograms.  This can happen when using the Prometheus receiver.  This mirrors behavior in google [prometheus-engine](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/17c5e8c475e0c224bf54c7a15531597f18885949/pkg/export/transform.go#L260).  Also, avoid divide by zero in exponential histograms.

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -523,7 +523,7 @@ func (m *metricMapper) histogramPoint(point pdata.HistogramDataPoint) *monitorin
 	}
 
 	mean := float64(0)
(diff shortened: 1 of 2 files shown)


damemi (approved):
lgtm

small side note for a first-time reader of this the divide-by-zero checks are a little confusing, because I expected the if statements to do something different in the case of div/0 but they are actually doing the normal flow in case of !div/0. not a big deal though

PR #341, opened 2022-04-01
Add sum of squared deviation estimate

dashpole:
This is based on how google [prometheus-engine](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/17c5e8c475e0c224bf54c7a15531597f18885949/pkg/export/transform.go#L264) calculates the sum of squared deviations.  I need to double-check the implications of sending an estimate without having an exact SOSD.

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -500,25 +500,43 @@ func (m *metricMapper) exemplars(exs pdata.ExemplarSlice) []*distribution.Distri
 // histogramPoint maps a histogram data point into a GCM point.
 func (m *metricMapper) histogramPoint(point pdata.HistogramDataPoint) *monitoringpb.TypedValue {
 	counts := make([]int64, len(point.BucketCounts()))
+	var mean, deviation, prevBound float64
+
 	for i, v := range point.BucketCounts() {
 		counts[i] = int64(v)
 	}
 
-	mean := float64(0)
 	if !math.IsNaN(point.Sum()) && point.Count() > 0 { // Avoid divide-by-zero
 		mean = float64(point.Sum() / float64(point.Count()))
+	}
+
+	bounds := point.ExplicitBounds()
+	// Calculate the sum of squared deviation.
+	for i, bound := range bounds {
+		// Assume all points in the bucket occur at the middle of the bucket range
+		middleOfBucket := (prevBound + bound) / 2
+		deviation += float64(counts[i]) * (middleOfBucket - mean) * (middleOfBucket - mean)
+		prevBound = bound
+	}
+	// The infinity bucket is an implicit +Inf bound after the list of explicit bounds.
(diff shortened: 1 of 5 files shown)


dashpole:
This is ready for review

exporter/collector/metrics.go
```diff
@@ -518,25 +518,43 @@ func (m *metricMapper) exemplars(exs pdata.ExemplarSlice) []*distribution.Distri
...
+	// Calculate the sum of squared deviation.
+	for i, bound := range bounds {
+		// Assume all points in the bucket occur at the middle of the bucket range
+		middleOfBucket := (prevBound + bound) / 2
+		deviation += float64(counts[i]) * (middleOfBucket - mean) * (middleOfBucket - mean)
+		prevBound = bound
+	}
+	// The infinity bucket is an implicit +Inf bound after the list of explicit bounds.
+	// Make sure we have a count for that bucket, just in-case.
+	if len(counts) == len(bounds)+1 {
```
aabmass, line 541: Are there actually cases where the overflow bucket is not present? It doesn't appear to be optional in the [proto code](https://github.com/open-telemetry/opentelemetry-proto/blob/cef5447c443d558a4056a93a35d48b8a72e0fad6/opentelemetry/proto/metrics/v1/metrics.proto#L473). I believe Prometheus also requires +Inf to be present.

aabmass:
The assumptions this makes seem unreasonable. I'm thinking for our regular GCM exporter, we should just not send SSD at all, what do you think?

dashpole:
The best case I can make for generally setting SOSD is that it isn't an optional field.  We are essentially always setting it to 0.  The estimate here _seems_ better than setting it to 0, although could possibly be confusing and misleading if it is too far off...

Alternatively, I can make this optional and off by default.  But users may want to set this if they want to graph it for some reason.  I do find it somewhat odd that GMP sets it given that it isn't part of a prom histogram...

Let me know what direction you think is best, and we can do that.

exporter/collector/metrics.go
dashpole: Prom does require it.  I can remove it.  I just found it surprising, and didn't trust it.

dashpole:
Example from the integration test run.  The higher lines (~6) are with the estimated SOSD. The lower line is with a zero SOSD.  The higher line is generally closer to accurate, although both are technically not correct.  SOSD does not appear to impact any percentiles or the mean or count aggregations.  It only matters when graphing the standard deviation.

Screen Shot 2022-04-19 at 4 31 37 PM

dashpole:
This doesn't enable it by default anymore.

dashpole:
Updated to not set SOSD by default.  The enable config option is only useable by other exporters (e.g. GMP), and is not settable by users.

aabmass (approved):

PR #342, opened 2022-04-01
Initial gmp exporter

dashpole:
The new exporter adds type suffixes, and maps to the the prometheus_target monitored resource.  It wraps the configuration exposed by the standard googlecloud exporter, and only exposes a subset of that configuration.  The remaining fields are hard-coded to ensure a GMP-compatible export.

Customizing the reset logic will be done after https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/360 merges in a follow-up change.


dashpole:
This is ready for review

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+
+	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
+)
+
+const (
+	// The value of "type" key in configuration.
+	typeStr        = "googlemanagedprometheus"
```
punya, line 30: I think we're required to say "Google Managed Service for Prometheus." Please confirm with Lee before resolving this comment.

dashpole: I double-checked.  "googlemanagedprometheus" is fine to use for now.

exporter/collector/googlemanagedprometheus.go
```diff
@@ -0,0 +1,61 @@
...
+
+package collector
+
+import (
+	"go.opentelemetry.io/collector/model/pdata"
+	semconv "go.opentelemetry.io/collector/model/semconv/v1.8.0"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+func MapToPrometheusTarget(res pdata.Resource) *monitoredrespb.MonitoredResource {
```
aabmass, line 23: Should this file go in the googlemanagedprometheus module folder?

exporter/collector/config.go
```diff
@@ -33,6 +35,12 @@ type Config struct {
 	TraceConfig TraceConfig `mapstructure:"trace"`
 
 	MetricConfig MetricConfig `mapstructure:"metric"`
+
+	// MapMonitoredResource is not exposed as an option in the configuration, but
+	// can be used by other exporters to extend the functionality of this
+	// exporter. It allows overriding the function used to map otel resource to
+	// monitored resource.
+	MapMonitoredResource func(pdata.Resource) *monitoredrespb.MonitoredResource
```
aabmass, line 43: I wonder if putting this in the each of the MetricConfig and TraceConfig would be better. Plus, logging has some inconsistencies with GCM monitored resource

aabmass (approved):

exporter/collector/googlemanagedprometheus_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/model/pdata"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+func TestMapToPrometheusTarget(t *testing.T) {
```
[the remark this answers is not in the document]
dashpole: I had it there originally, but it meant the collector integration tests adds a dependency on collector interfaces, which I didn't like.   But I do like the idea of keeping all of the GMP code together, and the integration test's dependencies probably don't matter.

exporter/collector/googlemanagedprometheus/config.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+
+	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
+)
+
+// Config defines configuration for Google Cloud Managed Service for Prometheus exporter.
+type Config struct {
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/googlemanagedprometheus.go
dashpole: moved.

exporter/collector/config.go
dashpole: moved to metric config, since logging config doesn't exist yet, and trace doesn't do MR yet

dashpole:
I confirmed that metrics from the GMP integration test are queryable via the "Managed Service for Prometheus" UI in the GCP console.

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -0,0 +1,49 @@
...
+	serviceNamespace := getStringOrEmpty(attrs, semconv.AttributeServiceNamespace)
+	if serviceNamespace != "" {
+		job = serviceNamespace + "/" + job
+	}
+	return &monitoredrespb.MonitoredResource{
+		Type: "prometheus_target",
+		Labels: map[string]string{
+			"location":  getStringOrEmpty(attrs, semconv.AttributeCloudAvailabilityZone),
+			"cluster":   getStringOrEmpty(attrs, semconv.AttributeK8SClusterName),
+			"namespace": getStringOrEmpty(attrs, semconv.AttributeK8SNamespaceName),
```
xichen2020, line 37: For Managed GMP, the namespace is renamed from `__meta_kubernetes_namespace`. 
(https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/apis/monitoring/v1/types.go#L381).

For unmanaged GMP, IIRC we rely users to set the `namespace` label via global labels or relabeling rules in their Prometheus configuration.

Does that match the behavior here?

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -0,0 +1,49 @@
...
+			"job":       job,
+			"instance":  getStringOrEmpty(attrs, semconv.AttributeServiceInstanceID),
```
xichen2020, line 39: For Managed GMP, `instance` label is set to `<pod_name>:<port number>` or `<pod_name>:<port name>`. https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/apis/monitoring/v1/types.go#L508

Self-managed GMP sets the `instance` label to the Prometheus `instance` label (https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/setup/setup.go#L70).

What would it look like for a process using GMP exporter on GKE or GCE?

exporter/collector/googlemanagedprometheus/monitoredresource_test.go
```diff
@@ -0,0 +1,115 @@
+// Copyright 2022 Google LLC
```
xichen2020, line 1: Unit testing is useful, but it may be valuable to run GMP collector and GMP exporter side-by-side in the same testing environment and compare the exported metrics to see whether the configured resource labels for GMP exporter match expectations (and resemble the GMP collector where it makes sense)

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -0,0 +1,39 @@
...
+	switch metric.DataType() {
+	case pmetric.MetricDataTypeSum:
+		return "/counter"
+	case pmetric.MetricDataTypeGauge:
+		return "/gauge"
+	case pmetric.MetricDataTypeSummary:
+		return "/summary"
+	case pmetric.MetricDataTypeHistogram:
+		return "/histogram"
+	default:
```
xichen2020, line 36: Should this return empty string or an error? What other scenarios can fall through to here?

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -0,0 +1,39 @@
...
+	return baseName + gmpMetricSuffix(metric)
+}
+
+func gmpMetricSuffix(metric pdata.Metric) string {
+	switch metric.DataType() {
+	case pmetric.MetricDataTypeSum:
+		return "/counter"
+	case pmetric.MetricDataTypeGauge:
+		return "/gauge"
+	case pmetric.MetricDataTypeSummary:
```
xichen2020, line 32: This doesn't quite match the logic in here: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/series_cache.go#L428

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"go.opentelemetry.io/collector/model/pdata"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+)
+
+func GetMetricName(baseName string, metric pdata.Metric) string {
+	return baseName + gmpMetricSuffix(metric)
+}
+
+func gmpMetricSuffix(metric pdata.Metric) string {
```
xichen2020, line 26: What do we decide to do with `unknown` typed metrics?

exporter/collector/integrationtest/testdata/gmp_config.yaml
```diff
@@ -0,0 +1,20 @@
...
+  googlemanagedprometheus:
+  googlemanagedprometheus/customname:
+    project: my-project
+    user_agent: opentelemetry-collector-contrib {{version}}
+    endpoint: test-metric-endpoint
+    use_insecure: true
+
+service:
+  pipelines:
+    traces:
```
xichen2020, line 17: Should this be `metrics`?

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+
+// createDefaultConfig creates the default configuration for exporter.
+func createDefaultConfig() config.Exporter {
+	return &Config{
+		ExporterSettings: config.NewExporterSettings(config.NewComponentID(typeStr)),
+		TimeoutSettings:  exporterhelper.TimeoutSettings{Timeout: defaultTimeout},
+		RetrySettings:    exporterhelper.NewDefaultRetrySettings(),
+		QueueSettings:    exporterhelper.NewDefaultQueueSettings(),
+		GMPConfig: GMPConfig{
+			UserAgent: "opentelemetry-collector-contrib {{version}}",
```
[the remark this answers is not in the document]
jsuereth: +1 to a consistent layout, however also +1 to making sure we know "opentelemetry" is involved somewhere in the UA (for diagnosing bugs).

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	mExp, err := collector.NewGoogleCloudMetricsExporter(ctx, eCfg.GMPConfig.toCollectorConfig(), params.TelemetrySettings.Logger, params.BuildInfo.Version, eCfg.Timeout)
+	if err != nil {
+		return nil, err
+	}
+	return exporterhelper.NewMetricsExporter(
+		cfg,
+		params,
+		mExp.PushMetrics,
+		exporterhelper.WithShutdown(mExp.Shutdown),
+		// Disable exporterhelper Timeout, since we are using a custom mechanism
```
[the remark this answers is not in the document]
jsuereth: Yeah.  The exporterhelper is designed to work with "naive" clients that just fire bits at a connection.   You can use an "external" mechanism to do timeouts there, or disable it if your client DOES have sophistication.

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (c *GMPConfig) toCollectorConfig() collector.Config {
+	// start with whatever the default collector config is.
+	cfg := collector.DefaultConfig()
+	// hard-code some config options to make it work with GMP
+	cfg.MetricConfig.Prefix = "prometheus.googleapis.com"
+	cfg.MetricConfig.SkipCreateMetricDescriptor = true
+	cfg.MetricConfig.InstrumentationLibraryLabels = false
+	cfg.MetricConfig.ServiceResourceLabels = false
+	cfg.MetricConfig.GetMetricName = GetMetricName
```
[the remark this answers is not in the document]
dashpole: This overrides the normal exporter's normal naming function (just leaving the name as-is) with the one defined in this package, which adds type suffixes.

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	cfg.MetricConfig.MapMonitoredResource = MapToPrometheusTarget
```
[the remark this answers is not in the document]
dashpole: This overrides the monitored resource mapping function with one that maps metrics to the prometheus_target resource.

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+		exporterhelper.WithTimeout(exporterhelper.TimeoutSettings{Timeout: 0}),
+		exporterhelper.WithQueue(eCfg.QueueSettings),
+		exporterhelper.WithRetry(eCfg.RetrySettings))
+}
+
+func (c *GMPConfig) toCollectorConfig() collector.Config {
+	// start with whatever the default collector config is.
+	cfg := collector.DefaultConfig()
+	// hard-code some config options to make it work with GMP
+	cfg.MetricConfig.Prefix = "prometheus.googleapis.com"
```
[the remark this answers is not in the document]
dashpole: ack

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -0,0 +1,49 @@
...
+	"go.opentelemetry.io/collector/model/pdata"
+	semconv "go.opentelemetry.io/collector/model/semconv/v1.8.0"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+func MapToPrometheusTarget(res pdata.Resource) *monitoredrespb.MonitoredResource {
+	attrs := res.Attributes()
+	// Prepend namespace if it exists to match what is specified in
+	// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resource-attributes-1
+	job := getStringOrEmpty(attrs, semconv.AttributeServiceName)
```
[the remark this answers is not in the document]
dashpole: `service.name` is set to the job name if using the prometheus receiver, so the same mechanism (relabel rules) used by GMP to set the `job` label will work here.

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -0,0 +1,49 @@
...
+	serviceNamespace := getStringOrEmpty(attrs, semconv.AttributeServiceNamespace)
+	if serviceNamespace != "" {
+		job = serviceNamespace + "/" + job
+	}
+	return &monitoredrespb.MonitoredResource{
+		Type: "prometheus_target",
+		Labels: map[string]string{
+			"location":  getStringOrEmpty(attrs, semconv.AttributeCloudAvailabilityZone),
```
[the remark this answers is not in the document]
dashpole: Ack. This mirrors the logic used in the collector's [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor).

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: Soon ([open-telemetry/opentelemetry-collector-contrib#9416](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9416)), the prometheus receiver will automatically grab `__meta_kuberentes_namespace` from scrape SD labels without requiring any relabel rules.

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: Right now, this sets `instance` to the prometheus `instance` label.  The prometheus receiver converts `instance` to `service.instance.id`, which is the OTel equivalent.

If you want, I can change this to prefer using `<pod_name>:<port number>`, or `<pod_name>:<port name>`if those attributes are available.  Just let me know which should take precedence, and i'll add it.

exporter/collector/googlemanagedprometheus/monitoredresource_test.go
dashpole: Sounds good.  We can definitely do that.  If there are particular setups you would like me to test, let me know.

exporter/collector/googlemanagedprometheus/naming.go
dashpole: Otel currently does not have a notion of unknown typed metrics.  They are currently changed to gauges in the receiver.  Once otel adds support for them, we can handle them correctly here.  Supporting unknown-typed metrics is tracked here: https://github.com/open-telemetry/wg-prometheus/issues/69

exporter/collector/googlemanagedprometheus/naming.go
dashpole: The only case that this handled was exponential histograms.  I've added a specific case for it.  What suffix should we put on exponential histograms?

exporter/collector/googlemanagedprometheus/config.go
```diff
@@ -0,0 +1,52 @@
...
+	GMPConfig               `mapstructure:",squash"`
+
+	// Timeout for all API calls. If not set, defaults to 12 seconds.
+	exporterhelper.TimeoutSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	exporterhelper.QueueSettings   `mapstructure:"sending_queue"`
+	exporterhelper.RetrySettings   `mapstructure:"retry_on_failure"`
+}
+
+// GMPConfig is a subset of the collector config.
+type GMPConfig struct {
```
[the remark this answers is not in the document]
dashpole: I added a README which answers most of these questions.

exporter/collector/googlemanagedprometheus/factory.go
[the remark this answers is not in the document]
dashpole: changed to opentelemetry-collector-contrib/{version} to match formatting.

exporter/collector/googlemanagedprometheus/factory.go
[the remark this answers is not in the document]
dashpole: added a comment

exporter/collector/googlemanagedprometheus/factory.go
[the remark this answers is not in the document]
dashpole: added a comment

exporter/collector/googlemanagedprometheus/naming.go
dashpole: fixed

exporter/collector/integrationtest/testdata/gmp_config.yaml
dashpole: fixed

exporter/collector/googlemanagedprometheus/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+// createMetricsExporter creates a metrics exporter based on this config.
+func createMetricsExporter(
+	ctx context.Context,
+	params component.ExporterCreateSettings,
+	cfg config.Exporter) (component.MetricsExporter, error) {
+	eCfg := cfg.(*Config)
+	mExp, err := collector.NewGoogleCloudMetricsExporter(ctx, eCfg.GMPConfig.toCollectorConfig(), params.TelemetrySettings.Logger, params.BuildInfo.Version, eCfg.Timeout)
```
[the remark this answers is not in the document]
dashpole: Fixed in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/369.

dashpole:
Confirmed that summary metrics appear correctly in the Managed Prometheus promql window.

exporter/collector/googlemanagedprometheus/factory.go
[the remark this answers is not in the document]
xichen2020: Do you also want to structure it so it can be extended to include the source in the future? e.g., it may be useful to know whether the metrics are coming from Ops Agent vs otherwise.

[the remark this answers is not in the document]
dashpole: It is settable via config options.  The ops-agent could set it to: `user_agent: ops-agent-gmp/{{version}}` if they wanted, for example.

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -15,25 +15,34 @@
...
 	case pmetric.MetricDataTypeGauge:
-		return "/gauge"
+		return baseName + "/gauge"
 	case pmetric.MetricDataTypeSummary:
-		return "/summary"
+		// summaries are sent as the following series:
+		// * Sum: prometheus.googleapis.com/<baseName>_sum/summary:counter
+		// * Count: prometheus.googleapis.com/<baseName>_count/summary
+		// * Quantiles: prometheus.googleapis.com/<baseName>/summary
+		if strings.HasSuffix(baseName, "_sum") {
```
xichen2020, line 35: Just to confirm, `_sum` and `_count` series are sent as cumulatives and quantile series are sent as gauges?

exporter/collector/googlemanagedprometheus/monitoredresource.go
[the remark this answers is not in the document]
xichen2020: Thanks. I'd like to call attention to this comment block: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/setup/setup.go#L77. Specifically on GKE we may not be able to always use the underlying node's zone as the location as we may have both a zonal cluster and a regional cluster in the same project with the same name, so if we use the node's zone location we'll end up creating conflicts.

exporter/collector/googlemanagedprometheus/monitoredresource.go
xichen2020: Thanks. If users do configure a relabel rule to set the `namespace` label, which takes precedence?

exporter/collector/googlemanagedprometheus/monitoredresource.go
xichen2020: I generally see the OT setup w/ Prometheus receiver and GMP exporter resembles the self-deployed GMP experience as opposed to the managed experience. If so, using whatever `instance` label is set by user's Prometheus configuration makes sense as is done here, but we should note this creates a different experience from managed GMP. cc/ @jsuereth

exporter/collector/googlemanagedprometheus/naming.go
xichen2020: Can exponential histograms be modeled the same way as standard Prometheus histograms? GMP doesn't specifically know about exponential histograms (since it's not something Prometheus knows about either) so if it requires any special handling, it won't be supported and we should return an error.

exporter/collector/googlemanagedprometheus/naming.go
xichen2020: Note fixing this to handle unknown metrics properly is backward-incompatible and would require a major version bump. We should probably try to get it addressed before marking exporter support for Prometheus stable.

exporter/collector/googlemanagedprometheus/naming.go
dashpole: When I implement normalization, i'll drop exponential histograms.  Then it won't be possible for those to fall through to here.  Should I just remove that case then?

exporter/collector/googlemanagedprometheus/naming.go
dashpole: Agreed.  The same goes for info + stateset types if those are added in the future (they become gauges right now).

exporter/collector/googlemanagedprometheus/naming.go
dashpole: Yes, that is correct.  In the gmp integration test, here is the [_sum series for a summary](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/342/files#diff-92007daaaf28c12906bcd7ab2014ab132aaca919b99bbcfe6b1cca41c0510afcR227), here is the [_count series](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/342/files#diff-92007daaaf28c12906bcd7ab2014ab132aaca919b99bbcfe6b1cca41c0510afcR258), and here is a [quantile series](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/342/files#diff-92007daaaf28c12906bcd7ab2014ab132aaca919b99bbcfe6b1cca41c0510afcR290).

The logic is the same for the standard (non-gmp) exporter.

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: Relabel rules which set `namespace` won't have any impact here.  It would add a label named "namespace" to the metric.  To set the namespace attribute (e.g. outside of k8s, or for a static target), someone would have to use the resource processor:
```yaml
resource:
  attributes:
  - key: k8s.namespace.name
    value: my-custom-namespace
    action: upsert
```

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: The managed GMP link above just points to relabel rules added by the managed GMP operator.  You **can** still use relabel rules for `instance` in the otel Prometheus receiver (e.g. the way managed GMP does), and it will correctly change the instance label on the exporter.  So it should work in both cases.

exporter/collector/googlemanagedprometheus/monitoredresource.go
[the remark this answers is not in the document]
dashpole: I've updated the GMP exporter to fall-back to region if zone is not found.  I've opened https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9682 in collector-contrib to fix resource detection to work correctly in that casel

exporter/collector/googlemanagedprometheus/naming.go
xichen2020: Yes, let's remove the case. I'd still suggest returning an error in the default case to be cautious.

dashpole: Removed and it now returns an error.

dashpole:
I think this should be all ready to go.  @xichen2020 can you take a final pass?

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -0,0 +1,54 @@
...
+	attrs := res.Attributes()
+	// Prepend namespace if it exists to match what is specified in
+	// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#resource-attributes-1
+	job := getStringOrEmpty(attrs, semconv.AttributeServiceName)
+	serviceNamespace := getStringOrEmpty(attrs, semconv.AttributeServiceNamespace)
+	if serviceNamespace != "" {
+		job = serviceNamespace + "/" + job
+	}
+	location := getStringOrEmpty(attrs, semconv.AttributeCloudAvailabilityZone)
+	if location == "" {
```
xichen2020, line 33: Can you add a TODO and link to the issue mentioning the resource detection logic fix? Note that's also a breaking change for the exporter once merged here.

exporter/collector/metrics.go
```diff
@@ -949,40 +983,54 @@ func (m *metricMapper) labelDescriptors(
 	return result
 }
 
-// Returns (sum, count) metric names for a summary metric.
-func summaryMetricNames(name string) (string, string) {
-	sumName := fmt.Sprintf("%s%s", name, SummarySumSuffix)
-	countName := fmt.Sprintf("%s%s", name, SummaryCountPrefix)
-	return sumName, countName
+// Returns (sum, count, quantile) metric names for a summary metric.
+func (m *metricMapper) summaryMetricNames(pm pdata.Metric) (string, string, string, error) {
```
xichen2020, line 987: nit: `summaryMetricTypes` is a better name.

xichen2020 (approved):

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: Nothing needs to be done in the exporter, which already has the correct behavior.  Once the fix is in the resource detector, it will work correctly without any additional changes here.

xichen2020: Does that mean the GKE resource detector will use `cluster-location` instance attribute, determine whether the location is a zone or region (how feasible is this?), and set the zone/region attribute?

dashpole: Yes, that is exactly what it will do.  I have a [WIP detector](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/383/files#diff-67c35f4fb2417f61d45777e33b607496b52caba05000bc350d4b22b3a9b3ebc5R40-R50) that does this for SDK detection if you are curious.  It should be feasible to do.

exporter/collector/metrics.go
```diff
@@ -633,6 +638,11 @@ func (m *metricMapper) histogramToTimeSeries(
 		// Drop points without a value or without a sum
 		return nil
 	}
+	name, err := m.metricNameToType(metric.Name(), metric)
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/metrics.go
dashpole: done

exporter/collector/metrics.go
```diff
@@ -949,40 +983,54 @@ func (m *metricMapper) labelDescriptors(
...
 }
 
-// Returns (sum, count) metric names for a summary metric.
-func summaryMetricNames(name string) (string, string) {
-	sumName := fmt.Sprintf("%s%s", name, SummarySumSuffix)
-	countName := fmt.Sprintf("%s%s", name, SummaryCountPrefix)
-	return sumName, countName
+// Returns (sum, count, quantile) metric names for a summary metric.
+func (m *metricMapper) summaryMetricNames(pm pdata.Metric) (string, string, string, error) {
+	sumName, err := m.metricNameToType(pm.Name()+SummarySumSuffix, pm)
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/googlemanagedprometheus/monitoredresource.go
xichen2020: I see. Would it be more robust to add a new `location` resource attribute and use that? That said I don't know how difficult it is to do so.

dashpole: Resource attributes are defined by the otel semantic conventions: https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/cloud.yaml#L45.  Right now, it defines zone + region.  I suspect it would be difficult to convince them to add a third convention, "location", which is the zone or region.

xichen2020: I see. It seems possible to add this as a GKE-specific attribute, but maybe that's an overkill.

PR #343, opened 2022-04-01
feat(propagator): create a new one-way propagator

### LINKED ISSUE #268: propagator not mentioned or used
The `propagator` package correctly implements a necessary component in hooking user spans into GCP-generated traces. However, it is not mentioned in the README and not used in the examples. In fact, no code in the repo refers to it.

Please modify both the README and the example to use it.

muncus:
CloudTraceOneWayPropagator moves trace info from X-cloud-trace-context
into 'traceparent'.

Fixes #268

### DIFF
--- a/propagator/README.md
+++ b/propagator/README.md
@@ -1,0 +1,51 @@
+# OpenTelemetry Google Cloud Trace Propagators
+
+[![Docs](https://godoc.org/github.com/GoogleCloudPlatform/opentelemetry-operations-go/propagator?status.svg)](https://pkg.go.dev/github.com/GoogleCloudPlatform/opentelemetry-operations-go/propagator)
+
+This package contains Trace Context Propagators for use with [Google Cloud
+Trace](https://cloud.google.com/trace) that make it compatible with
+[OpenTelemetry](http://opentelemetry.io). 
+
+There are two available propagators in this package:
+
+### `CloudTraceOneWayPropagator`
+
+The `CloudTraceOneWayPropagator` reads the `X-Cloud-Trace-Context` header for
+trace and span IDs, and writes this data into the `traceparent` header.
+
+### `CloudTraceFormatPropagator`
+
+The `CloudTraceFormatPropagator` reads and writes the `X-Cloud-Trace-Context` header only.
+
+## Differences between Google Cloud Trace and W3C Trace Context
+
+Google Cloud Trace encodes trace information in the `X-Cloud-Trace-Context` HTTP
+header, using the format described in the [Trace documentation](https://cloud.google.com/trace/docs/setup#force-trace).
+
+OpenTelemetry uses the newer, W3C standard
+[`traceparent` header](https://www.w3.org/TR/trace-context/#traceparent-header)
+
+There is an important semantic difference between Cloud Trace's
+`TRACE_TRUE` flag, and W3C's `sampled` flag.
+
+As outlined in the [Trace
+documentation](https://cloud.google.com/trace/docs/setup#force-trace), setting
+the `TRACE_TRUE` flag will cause trace information to be collected.
+
+This differs from the W3C behavior, where the [`sampled`
+flag](https://www.w3.org/TR/trace-context/#sampled-flag) indicates that the
+caller *may* have recorded trace information, but does not necessarily impact
+the sampling done by other services.
+
+To preserve the Cloud-Trace behavior when using `traceparent`, you can use the
+[`ParentBased`
+sampler](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased) like
+so:
+
+```go
+import sdktrace go.opentelemetry.io/otel/sdk/trace
+sampler := sdktrace.ParentBased(
+    sdktrace.NeverSample(),
+    sdktrace.WithRemoteParentSampled(sdktrace.AlwaysSample()))
+)
+```
--- a/propagator/README.md
+++ b/propagator/README.md
@@ -10,8 +10,12 @@ There are two available propagators in this package:
 
 ### `CloudTraceOneWayPropagator`
 
-The `CloudTraceOneWayPropagator` reads the `X-Cloud-Trace-Context` header for
-trace and span IDs, and writes this data into the `traceparent` header.
+The `CloudTraceOneWayPropagator` works like the standard `TraceContext`
+propagator if a `traceparent` header is present. If `traceparent` is not found,
+it reads the `X-Cloud-Trace-Context` header for trace and span IDs, and creates
+the corresponding `traceparent` header.
+
+This is useful for ensuring spans created in your code are attached to the traces that some Google Cloud services [automatically trace](https://cloud.google.com/trace/docs/overview#configurations_with_automatic_tracing).
 
 ### `CloudTraceFormatPropagator`
 
--- a/propagator/propagator.go
+++ b/propagator/propagator.go
@@ -54,14 +54,24 @@ func (e errInvalidHeader) Error() string {
 	return fmt.Sprintf("invalid header %s", e.header)
 }
 
-// CloudTraceOneWayPropagator will extract Trace IDs from the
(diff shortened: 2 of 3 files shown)


propagator/propagator.go
```diff
@@ -38,9 +38,13 @@ const traceContextHeaderFormat = "^(?P<trace_id>[0-9a-f]{32})/(?P<span_id>[0-9]{
...
 
-// traceContextHeaders is the list of headers that are propagated. Cloud Trace only requires
+// cloudTraceContextHeaders is the list of headers that are propagated. Cloud Trace only requires
 // one element in the list.
-var traceContextHeaders = []string{TraceContextHeaderName}
+var cloudTraceContextHeaders = []string{TraceContextHeaderName}
+
+// TraceparentHeaderName is the HTTP header field for w3c standard trace information
+// https://www.w3.org/TR/trace-context/
+const TraceparentHeaderName = "traceparent"
```
dashpole, line 47:
```suggestion
// traceparentHeaderName is the HTTP header field for w3c standard trace information
// https://www.w3.org/TR/trace-context/
const traceparentHeaderName = "traceparent"
```
Lets keep this unexported unless neccessary.

propagator/propagator.go
```diff
@@ -50,6 +54,32 @@ func (e errInvalidHeader) Error() string {
...
+	sc, err := spanContextFromXCTCHeader(header)
+	if err != nil {
+		log.Printf("CloudTraceFormatPropagator: %v", err)
+		return ctx
+	}
+	return trace.ContextWithRemoteSpanContext(ctx, sc)
+}
+
+func (p CloudTraceOneWayPropagator) Fields() []string {
+	return []string{TraceContextHeaderName, TraceparentHeaderName}
```
dashpole, line 78:
```suggestion
	return []string{traceparentHeaderName}
```
Based on [the docs](https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/propagation.go#L108), fields should only include those set during Inject().  Also, make it unexported.

propagator/propagator_test.go
```diff
@@ -114,6 +114,114 @@ func TestValidTraceContextHeaderFormats(t *testing.T) {
...
+	}{
+		{
+			"valid TraceID and SpanID with sampled flag",
+			trace.SpanContextConfig{
+				TraceID:    validTraceID,
+				SpanID:     validSpanID,
+				TraceFlags: trace.FlagsSampled,
+			},
+			map[string]string{
+				TraceparentHeaderName: fmt.Sprintf("00-%s-%s-01", validTraceID.String(), validSpanID.String()),
```
dashpole, line 191:
```suggestion
				traceparentHeaderName: fmt.Sprintf("00-%s-%s-01", validTraceID.String(), validSpanID.String()),
```

propagator/propagator_test.go
```diff
@@ -114,6 +114,114 @@ func TestValidTraceContextHeaderFormats(t *testing.T) {
...
+			},
+		},
+		{
+			"valid TraceID and SpanID without sampled flag",
+			trace.SpanContextConfig{
+				TraceID: validTraceID,
+				SpanID:  validSpanID,
+			},
+			map[string]string{
+				TraceparentHeaderName: fmt.Sprintf("00-%s-%s-00", validTraceID.String(), validSpanID.String()),
```
dashpole, line 201:
```suggestion
				traceparentHeaderName: fmt.Sprintf("00-%s-%s-00", validTraceID.String(), validSpanID.String()),
```

propagator/propagator.go
```diff
@@ -50,6 +54,32 @@ func (e errInvalidHeader) Error() string {
...
+}
+
+func (p CloudTraceOneWayPropagator) Extract(ctx context.Context, carrier propagation.TextMapCarrier) context.Context {
+	header := carrier.Get(TraceContextHeaderName)
+	if header == "" {
+		return ctx
+	}
+	sc, err := spanContextFromXCTCHeader(header)
+	if err != nil {
+		log.Printf("CloudTraceFormatPropagator: %v", err)
```
dashpole, line 71:
```suggestion
		log.Printf("CloudTraceOneWayPropagator: %v", err)
```
Typo

dashpole:
cc @aabmass

propagator/README.md
```diff
@@ -0,0 +1,51 @@
...
+Trace](https://cloud.google.com/trace) that make it compatible with
+[OpenTelemetry](http://opentelemetry.io). 
+
+There are two available propagators in this package:
+
+### `CloudTraceOneWayPropagator`
+
+The `CloudTraceOneWayPropagator` reads the `X-Cloud-Trace-Context` header for
+trace and span IDs, and writes this data into the `traceparent` header.
+
```
dashpole, line 15: Can we add a recommended usage golang snippet?  Something like:

```golang
propagation.NewCompositeTextMapPropagator(propagator.CloudTraceOneWayPropagator{}, propagation.TraceContext{}, propagation.Baggage{})
```

We want to make sure traceparent takes precedence over X-Cloud-Trace-Context when reading.

propagator/propagator.go
```diff
@@ -50,6 +54,32 @@ func (e errInvalidHeader) Error() string {
 	return fmt.Sprintf("invalid header %s", e.header)
 }
 
+// CloudTraceOneWayPropagator will extract Trace IDs from the
+// x-cloud-trace-context header, and inject the w3c-standard 'traceparent'
+// header.
+type CloudTraceOneWayPropagator struct {
+	propagation.TraceContext
```
dashpole, line 61: Do we need to embed TraceContext here?  If we use this the way I recommended in the readme above, we would end up calling `propagation.TraceContext.Inject()` twice.  Are there other uses-cases where we don't want to make a composite propagator that it is needed for?

propagator/README.md
muncus: Do we want to force users to create a composite propagator for Trace Context?
Multiple propagators using the same key in the TextMap can lead to silent failures and, IMO, should be avoided where possible.

From a DevEx perspective, a single Propagator would let us have explicit precedence (and tests to prove it!), in addition to being less prone to user error (e.g. by getting the order wrong). (this approach requires an embedded TraceContext propagator). Usage:
`propagation.NewCompositeTextMapPropagator(gcpprop.CloudTraceOneWayPropagator{}, propagation.Baggage{})` (if we go this way, i'd suggest we rename to `CloudTraceToTraceContextPropagator` or similar)

If we want to take the opposite approach, and make a single-purpose propagator, perhaps we should avoid having `Inject()` do anything, and instead rely on the user specifying their own propagator. DevEx on this would be more error-prone, since the order of propagators matters. Usage:
`propagation.NewCompositeTextMapPropagator(gcpprop.CloudTraceOneWayPropagator{}, propagation.TraceContext{}, propagation.Baggage{})`

Either way, the code needs to change, but we should be explicit about the choice. 😄 What do you think?

propagator/propagator.go
muncus: I left a longer response to this on your comment in the README. Lets continue the discussion there.

propagator/README.md
dashpole: For the first case, we would likely also want to call `propagation.TraceContext.Extract()` from as part of Extract as well.

The order of propagators only matters if both are present, which should be rare. There could be a case in which the user wants to prefer XCT (e.g. if they want to be a child of a GCLB span in GAE).  It would be more flexible to have it separate as well if someone wanted to use it with a different "base" propagator (e.g. jaeger) for some reason.

@aabmass any thoughts?

muncus: I've chosen to go for the simpler DevEx, and made the Propagator read both headers, but prioritize `traceparent`.
This is a tradeoff with flexibility, as you mention, but helps users converge on the w3c standard, which seems like an acceptable compromise.
I also refactored the tests so that the priority of trace headers is clear.

dashpole (approved):

PR #344, opened 2022-04-04
collector: added log conversion to entries

braydonk:
Added functionality to convert pdata Logs to Cloud Logging entries. Also
added necessary configuration and self observability.


exporter/collector/logs.go
```diff
@@ -43,20 +62,149 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+	httpRequestMap, ok := unmarshalledMessage[httpRequestKey]
+	if !ok {
+		return nil, message, fmt.Errorf("message has no key %s", httpRequestKey)
+	}
+	httpRequestBytes, err := json.Marshal(httpRequestMap)
+	if err != nil {
+		return nil, message, err
+	}
+	var httpRequest *httpRequestLog
+	if err := json.Unmarshal(httpRequestBytes, &httpRequest); err != nil {
```
damemi, line 197: instead of marshalling and unmarshalling, can you just cast the map directly like `httpRequest := httpRequestMap.(*httpRequestLog)`?

exporter/collector/logs.go
```diff
@@ -43,20 +62,149 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+func (l *LogsExporter) logToEntry(
+	log pdata.LogRecord,
+	attributes map[string]interface{},
+	mr *monitoredres.MonitoredResource,
+) (logging.Entry, error) {
+	entry := logging.Entry{
+		Timestamp: log.Timestamp().AsTime(),
+		Severity:  logging.Severity(log.SeverityNumber()),
+		Trace:     log.TraceID().HexString(),
+		SpanID:    log.SpanID().HexString(),
```
damemi, line 98: I couldn't get logs to send through to the cloud logging dashboard with these 4 fields. Maybe we disable them for now. We will have to do more processing on Severity and Timestamp anyway

exporter/collector/logs.go
```diff
@@ -43,20 +62,149 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+		}
+	}
+
+	if len(logBody) > 0 {
+		unmarshalledLogBody := make(map[string]interface{})
+		err := json.Unmarshal(logBody, &unmarshalledLogBody)
+		if err != nil {
+			entry.Payload = unmarshalledLogBody
+		}
+	}
```
damemi, line 120: Not sure what the issue is, but log bodies are coming through empty when I run this (see below). is there a simpler way we can pass this for now or just continue using text?
Screen Shot 2022-04-05 at 12 35 17 PM

damemi (commented):
couple comments. I had some trouble getting this to send logs through. @braydonk how are you building/testing? could you share your config?

exporter/collector/logs.go
```diff
@@ -43,20 +62,149 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+
+func (l *LogsExporter) logToEntry(
+	log pdata.LogRecord,
+	attributes map[string]interface{},
+	mr *monitoredres.MonitoredResource,
+) (logging.Entry, error) {
+	entry := logging.Entry{
+		Timestamp: log.Timestamp().AsTime(),
+		Severity:  logging.Severity(log.SeverityNumber()),
+		Trace:     log.TraceID().HexString(),
```
fiarabbit, line 97: I think Cloud Logging `LogEntry.trace` should be prefixed by `projects/{projectId}/traces` (https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry); otherwise, `trace` value will be different between GCP infrastructure log (such as AppEngine request log, etc.) and application log.

fiarabbit (commented):
I'm really looking forward to this feature.

exporter/collector/logs.go
braydonk: You're probably right. If it's okay with you @damemi I'll probably punt Trace off of this PR as well since you and your team are probably better equipped to do this part right.

damemi: @braydonk sounds good, we're trying to iterate on this rather than get it perfect all at once :)

exporter/collector/logs.go
braydonk: Fixed in [72b3690](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/344/commits/72b3690d4a49970b7eaffdd10072dab376d4c72e)

exporter/collector/logs.go
braydonk: Fixed in [72b3690](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/344/commits/72b3690d4a49970b7eaffdd10072dab376d4c72e)

exporter/collector/logs.go
```diff
@@ -43,20 +62,149 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+		if err == nil {
+			httpRequest.Latency = latency
+		}
+	}
+
+	return httpRequest, strippedMessage, nil
+}
+
+func extractHttpRequestFromLog(message []byte) (*httpRequestLog, []byte, error) {
+	httpRequestKey := "httpRequest" // TODO(@braydonk) Should this come from the config?
```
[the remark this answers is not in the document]
braydonk: Fixed in [7903140](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/344/commits/79031407118a05e991071d25953f06de42894659)

exporter/collector/logs.go
braydonk: Obsolete with [7903140](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/344/commits/79031407118a05e991071d25953f06de42894659)

exporter/collector/logs.go
braydonk: I'll mark this conversation resolved, we'll keep a note of it as something to look into next.

exporter/collector/logs.go
```diff
@@ -43,20 +74,107 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+	}
+
+	logBody := log.Body().BytesVal()
+	if len(logBody) > 0 {
+		entry.Payload = json.RawMessage(logBody)
+	}
+
+	httpRequestAttr, ok := log.Attributes().Get(HTTPRequestAttributeKey)
+	if ok {
+		httpRequest, err := l.parseHTTPRequest(httpRequestAttr.BytesVal())
```
braydonk, line 117: I actually don't know if this is going to be bytes or string in a typical use case. @damemi if you happen to know offhand let me know, otherwise I'll manually test it tomorrow.

exporter/collector/logs.go
```diff
@@ -43,20 +74,107 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+	Referer                        string `json:"referer"`
+	Latency                        string `json:"latency"`
+	CacheLookup                    bool   `json:"cacheLookup"`
+	CacheHit                       bool   `json:"cacheHit"`
+	CacheValidatedWithOriginServer bool   `json:"cacheValidatedWithOriginServer"`
+	CacheFillBytes                 int64  `json:"cacheFillBytes,string"`
+	Protocol                       string `json:"protocol"`
+}
+
+func (l logMapper) parseHTTPRequest(httpRequestAttr []byte) (*logging.HTTPRequest, error) {
```
damemi, line 147: What if you pass in `httpRequestAttr.MapVal()` (instead of `BytesVal`), so you can access all the keys directly using `Get()` instead of needing to marshal to a new `httpRequestLog` struct?

braydonk: I'm trying to do this and it should work, but it leads to a ton of extra stuff casting and checking types. Might be better to use `mapstructure`, i.e.
- Get the `com.google.httpRequest` attribute 
- convert the map value to `map[string]interface{}`
- Unmarshal to `httpRequestLog` with `mapstructure`
That way we still have the nicety of a struct enforcing the types of the fields while not needing to do a json Unmarshal from bytes.

damemi: Ah sorry, I didn't realize the level of encapsulation `MapVal()` had with its return value. I was hoping it would just be a more accessible map. We can mark this as a TODO for optimization then, the alternative doesn't seem to give much of a difference

braydonk: Added a TODO comment in [c0d53f7](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/344/commits/c0d53f791521bae806c4f308ab7c2cb4263fdc48)

damemi (approved):
LGTM
as we add more conversions to fill out the spec requirements, we might find some things that can be refactored from this. But I think this is a great spot to build from. Thanks @braydonk!

PR #346, opened 2022-04-06
Update fixtures based on upstream changes to the prometheus receiver

dashpole:
The resource labels set by the upstream prometheus receiver are changed to better align with OpenTelemetry semantic conventions: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/8266

This changes the input fixtures to reflect those changes.  It does not change any of the expectations, since those labels aren't used.  The GMP exporter will on these new values, though.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics.json
@@ -10,31 +10,25 @@
                         }
                     },
                     {
-                        "key": "host.name",
+                        "key": "net.host.name",
                         "value": {
                             "stringValue": "ops-agent-metrics.c.otel-opsagenttester.internal"
                         }
                     },
                     {
-                        "key": "job",
-                        "value": {
-                            "stringValue": "otel-collector"
-                        }
-                    },
-                    {
-                        "key": "instance",
+                        "key": "service.instance.id",
                         "value": {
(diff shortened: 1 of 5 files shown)


exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics.json
```diff
@@ -10,31 +10,25 @@
...
                     },
                     {
-                        "key": "host.name",
+                        "key": "net.host.name",
                         "value": {
                             "stringValue": "ops-agent-metrics.c.otel-opsagenttester.internal"
                         }
                     },
                     {
-                        "key": "job",
```
damemi, line 19: @dashpole I'm assuming you are intentionally removing all the `job` keys rather than switch to `service.name` (trying to gleam context from https://github.com/open-telemetry/opentelemetry-specification/pull/2381). Are these obsolete now or will be added back later? You do have a new `service.name` entry in the workload_metrics fixture below so just curious

damemi (approved):

exporter/collector/integrationtest/testdata/fixtures/ops_agent_self_metrics.json
dashpole: Good catch.  I'm trying to remove the job label everywhere.  That one below should've been removed, rather than changed.

PR #353, opened 2022-04-12
Prepare for releasing v1.4.0 and v0.28.0

dashpole:
For stable minor releases, increment minor version by 1, from v1.3.0 to v1.4.0.

For unstable minor releases, increment to the minor version of the opentelemetry-go unstable dependency (v0.28.0).


punya (commented):
Wait, why is v0.28.0 and not v0.27.0?

dashpole:
I added make commands to make the release process easier

PR #355, opened 2022-04-12
Parse basic attributes for cloud logging exporter

damemi:
Parse out some basic fields from OTel LogRecord to GCP Entry:
* Instrumentation scope/version
* Timestamp (need to figure out how to read `observed_time_unix_nano`)
* SourceLocation
* Trace/Span ID

And changes how `log.Body()` is passed to Payload depending on the body's value type

xref https://b.corp.google.com/issues/228845177

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -79,12 +79,20 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld pdata.Logs) error {
 	for i := 0; i < ld.ResourceLogs().Len(); i++ {
 		rl := ld.ResourceLogs().At(i)
 		mr, _ := mapper.resourceToMonitoredResource(rl.Resource())
+
 		for j := 0; j < rl.ScopeLogs().Len(); j++ {
 			sl := rl.ScopeLogs().At(j)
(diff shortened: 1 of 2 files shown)


damemi:
Working on tests

exporter/collector/logs.go
```diff
@@ -102,14 +112,47 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld pdata.Logs) error {
...
 	entry := logging.Entry{
 		Resource: mr,
 	}
 
-	logBody := log.Body().BytesVal()
-	if len(logBody) > 0 {
-		entry.Payload = json.RawMessage(logBody)
+	entry.Labels = make(map[string]string)
+	if len(instrumentationSource) > 0 {
+		entry.Labels["instrumentation_source"] = instrumentationSource
```
dashpole, line 124: For metrics, we control whether or not these are added with a config option: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/collector_logging/exporter/collector/config.go#L75.  We might want to bring that option to the top-level of the config and use it for all types of telemetry (including here).  Maybe just add a TODO for now.

dashpole (approved):

PR #356, opened 2022-04-13
fix: trace http server example for probability sampler

bweston92:
There is no `ProbabilitySampler` (I believe that opencensus did have this though...), the probability functionality is available through the `TraceIDRatioBased`.


dashpole (approved):
Thanks.  ProbabilitySampler was renamed to TraceIDRatioBased sampler a while back: https://github.com/open-telemetry/opentelemetry-go/blob/57a248d37518dde41f2fc11de09114280f7515f5/CHANGELOG.md#changed-20

PR #358, opened 2022-04-14
Send timeseries synchronously

dashpole:
This makes self-observability metrics work again, and makes integration tests fail when sending time-series fails.

It updates the documentation to again recommend using the batch processor.  The `telemetry_scrape_duration_seconds` metric has an INT64 count for the summary metric's count instead of a DOUBLE value.  Even after deleting the metric descriptor, it won't allow me to change the type from INT64 to DOUBLE.  To fix the test, I just renamed the metric.

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -54,7 +54,7 @@ These instructions are to get you up and running quickly with the GCP exporter i
           exporters: [googlecloud, logging]
         metrics:
           receivers: [otlp]
-          processors: [memory_limiter]
+          processors: [memory_limiter, batch]
           exporters: [googlecloud, logging]
     ```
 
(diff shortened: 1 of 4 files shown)


exporter/collector/README.md
```diff
@@ -54,7 +54,7 @@ These instructions are to get you up and running quickly with the GCP exporter i
           exporters: [googlecloud, logging]
         metrics:
           receivers: [otlp]
-          processors: [memory_limiter]
+          processors: [memory_limiter, batch]
```
tbarker25, line 57: Why are we adding `batch` here? I thought that we were doing batching internally so we didn't need this.

tbarker25 (approved):

exporter/collector/README.md
dashpole: We are still splitting large batches in the exporter, but aren't grouping smaller batches together anymore after this change.  We still need the processor to do that for us.

tbarker25: Gotcha. It does feel a little bit icky to have two batching steps, but correctness > simplicity so I guess that's fine.

dashpole: Its what we do in the opencensus exporter as well, so at least it won't be a regression for anyone.

PR #359, opened 2022-04-14
Drop histograms without a sum

dashpole:
With the newest release of the collector, pdata now supports histograms with an unset sum.  In our case, we don't want to send those histograms to cloud monitoring.

The first commit updates the collector version, and the second commit handles dropping points without a sum.


exporter/collector/metrics.go
```diff
@@ -647,7 +648,7 @@ func (m *metricMapper) exponentialHistogramToTimeSeries(
 	_ pdata.ExponentialHistogram,
 	point pdata.ExponentialHistogramDataPoint,
 ) []*monitoringpb.TimeSeries {
-	if point.Flags().HasFlag(pdata.MetricDataPointFlagNoRecordedValue) {
+	if point.Flags().HasFlag(pmetric.MetricDataPointFlagNoRecordedValue) {
```
damemi, line 651: just checking, this doesn't apply to exponential histograms?

dashpole: At least right now, exponential histograms' sum is [not optional](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L541), whereas fixed bucket histograms' sum [is optional](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L462).

damemi: Ah thanks

damemi (approved):

PR #360, opened 2022-04-18
Normalize histograms, summaries, and exponential histograms

dashpole:
## Background

Normalization is done for cumulative streams of metrics.  For example, if we have a counter that counts the number of seconds since 1970, we wouldn't want to report that number as having happened since we started reading it, since that would produce an enormous rate  (>1 billion over 10 seconds).  Instead, the first time we see a metric, we record its current value.  Then, each time we see the metric again, we "subtract" the original value from the new one.  That is trivial for a counter, and we already did that in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/323.  It is harder for other data types (how do you subtract a histogram from another histogram?).

## Changes

This PR extends tracking for the first time we saw a metric by extending the data point cache for other types.  It then copies logic for setting the start time for intervals from the sum implementation.  Finally, it implements "subtraction" for each of the remaining types: summaries, histograms, and exponential histograms.

For histograms, just subtract the previous sum, previous count, and subtract each bucket from the previous point.

For summaries, leave the quantiles as they are (treat them like gauges passing through), and just subtract previous sum and previous count.

For exponential histograms, subtract the previous sum, previous count.  Also subtract each bucket from the previous point, taking the offset into account.

### DIFF
--- a/exporter/collector/internal/datapointstorage/datapointcache.go
+++ b/exporter/collector/internal/datapointstorage/datapointcache.go
@@ -25,16 +25,41 @@ import (
 
 const gcInterval = 20 * time.Minute
 
-type Cache map[string]usedPoint
+type Cache struct {
+	numberCache               map[string]usedNumberPoint
+	summaryCache              map[string]usedSummaryPoint
+	histogramCache            map[string]usedHistogramPoint
+	exponentialHistogramCache map[string]usedExponentialHistogramPoint
+}
 
-type usedPoint struct {
+type usedNumberPoint struct {
 	point *pdata.NumberDataPoint
 	used  bool
 }
 
-// New instantiates a cache and starts background processes
+type usedSummaryPoint struct {
+	point *pdata.SummaryDataPoint
+	used  bool
+}
+
+type usedHistogramPoint struct {
+	point *pdata.HistogramDataPoint
+	used  bool
+}
+
+type usedExponentialHistogramPoint struct {
+	point *pdata.ExponentialHistogramDataPoint
+	used  bool
+}
+
+// NewCache instantiates a cache and starts background processes
 func NewCache(shutdown <-chan struct{}) Cache {
-	c := make(Cache)
+	c := Cache{
+		numberCache:               make(map[string]usedNumberPoint),
+		summaryCache:              make(map[string]usedSummaryPoint),
+		histogramCache:            make(map[string]usedHistogramPoint),
+		exponentialHistogramCache: make(map[string]usedExponentialHistogramPoint),
+	}
 	go func() {
 		ticker := time.NewTicker(gcInterval)
 		for c.gc(shutdown, ticker.C) {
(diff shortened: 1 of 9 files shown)


exporter/collector/metrics_test.go
```diff
@@ -495,7 +631,7 @@ func TestExponentialHistogramPointToTimeSeries(t *testing.T) {
 	// Add a second point with no value
 	hist.DataPoints().AppendEmpty().SetFlags(pmetric.MetricDataPointFlags(pmetric.MetricDataPointFlagNoRecordedValue))
 
-	tsl := mapper.exponentialHistogramToTimeSeries(mr, labels{}, metric, hist, point)
+	tsl := mapper.metricToTimeSeries(mr, labels{}, metric)
```
dashpole, line 634: Note to reviewers: this test was broken, in that it wasn't checking the second point with the "NoRecordedValue" flag at all.  This change fixes it.

exporter/collector/metrics.go
```diff
@@ -595,6 +602,14 @@ func (m *metricMapper) histogramToTimeSeries(
 		// Drop points without a value or without a sum
 		return nil
 	}
+	// Normalize the histogram point.
+	metricIdentifier := datapointstorage.Identifier(resource, extraLabels, metric, point.Attributes())
+	normalizedPoint := m.normalizeHistogramDataPoint(point, metricIdentifier)
```
dashpole, line 607: TODO: only normalize if it is cumulative

dashpole: done

quentinmit:
Commenting on the logic from your description, without reading the code:

- Histograms and summaries should only be normalized as described if they represent a cumulative "aggregation temporality". (Side note: I don't see a gauge aggregation temporality - that would also of course not require any normalization.) Subtracting the original sum/count/bucket counts seems correct, but make sure you are subtracting the correct buckets - if the bucket boundaries change you need to either reset or check if they're a superset of the previous point.
- For summaries I don't see any way to normalize quantiles (what does a quantile-over-cumulative even mean in this context?)
- Exponential histograms seem the same as regular histograms, except you can more reliably line up corresponding buckets

exporter/collector/metrics.go
```diff
@@ -746,12 +774,166 @@ func (m *metricMapper) normalizeNumberDataPoint(point pdata.NumberDataPoint, ide
...
+			// We found a cached start timestamp that wouldn't produce a valid point.
+			// Drop it and log.
+			m.obs.log.Info(
+				"data point being processed older than last recorded reset, will not be emitted",
+				zap.String("lastRecordedReset", start.Timestamp().String()),
+				zap.String("dataPoint", point.Timestamp().String()),
+			)
+			return nil
+		}
+		// Make a copy so we don't mutate underlying data
```
dashpole, line 896: Note: the scale can change, and the math below currently assumes the scale is the same.  We either need to treat the scale as a reset, or handle it correctly.

exporter/collector/metrics.go
```diff
@@ -746,12 +774,166 @@ func (m *metricMapper) normalizeNumberDataPoint(point pdata.NumberDataPoint, ide
...
+		point.CopyTo(newPoint)
+		// Use the start timestamp from the normalization point
+		newPoint.SetStartTimestamp(start.Timestamp())
+		// Adjust the value based on the start point's value
+		newPoint.SetCount(point.Count() - start.Count())
+		// We drop points without a sum, so no need to check here.
+		newPoint.SetSum(point.Sum() - start.Sum())
+		pointBuckets := point.BucketCounts()
+		startBuckets := start.BucketCounts()
+		if len(pointBuckets) != len(startBuckets) {
```
dashpole, line 850: TODO: we also need to treat it as a reset when bucket boundaries change, even if the number of buckets stays the same.

exporter/collector/metrics.go
dashpole: added a TODO, and reset when scale changes.

exporter/collector/metrics.go
dashpole: done.

exporter/collector/metrics.go
```diff
@@ -746,12 +774,180 @@ func (m *metricMapper) normalizeNumberDataPoint(point pdata.NumberDataPoint, ide
...
+			// We found a cached start timestamp that wouldn't produce a valid point.
+			// Drop it and log.
+			m.obs.log.Info(
+				"data point being processed older than last recorded reset, will not be emitted",
+				zap.String("lastRecordedReset", start.Timestamp().String()),
+				zap.String("dataPoint", point.Timestamp().String()),
+			)
+			return nil
+		}
+		// Make a copy so we don't mutate underlying data
```
jsuereth, line 800: Maybe explicitly call out that `Percentile`/`Quantile` values are copied here, as they are over unknown time windows, and can't be normalized.

exporter/collector/metrics.go
```diff
@@ -746,12 +774,180 @@ func (m *metricMapper) normalizeNumberDataPoint(point pdata.NumberDataPoint, ide
...
+			// Drop it and log.
+			m.obs.log.Info(
+				"data point being processed older than last recorded reset, will not be emitted",
+				zap.String("lastRecordedReset", start.Timestamp().String()),
+				zap.String("dataPoint", point.Timestamp().String()),
+			)
+			return nil
+		}
+		if point.Scale() != start.Scale() {
+			// TODO: It is possible, but difficult to compare exponential
```
jsuereth, line 904: This is "ok" for now.

I'd like to open abug that does the following:


1. If the "start" histogram scale is "lower resolution", then rescale the newest point to the start histogram scale and normalize.
2. If the "start" histogram scale is "higher resolution" then rescale the start point to the new resolution and REPLACE the cache with this rescaled version, using this version from now forward.
3. Always ensure scale value is the "lowest resolution" of start, current.

If OTEL hasn't provided a "rescale" method for ExponentialHistogram, it should.

Side note:  The "offset" for buckets shouldn't be as important, but MAY matter if we need to lower resolution even further.  The crux of the algorithm should remain the same.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -25,16 +25,41 @@ import (
 
 const gcInterval = 20 * time.Minute
 
-type Cache map[string]usedPoint
+type Cache struct {
```
jsuereth, line 28: Given the (possibly large) memory overhead of this feature, is it disableable?

exporter/collector/metrics.go
dashpole: Opened https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/366

exporter/collector/metrics.go
dashpole: done

exporter/collector/internal/datapointstorage/datapointcache.go
jsuereth: This is great!

jsuereth (approved):

PR #361, opened 2022-04-18
Parse log name from config or attribute

damemi:
This switches to using the generated clients and parses a log name either from the specified attribute or falls back to the default log name as configured (or returns an error if none is set)

xref b/228842726


damemi:
Been having some trouble compiling and running this manually, still figuring that out

exporter/collector/config.go
```diff
@@ -100,6 +102,10 @@ type ResourceFilter struct {
 	Prefix string `mapstructure:"prefix"`
 }
 
+type LogConfig struct {
+	DefaultLogName string `mapstructure:"defaultLogName"`
```
dashpole, line 106:
```suggestion
	DefaultLogName string `mapstructure:"default_log_name"`
```
config fields are separated by underscores instead of camelcase.

exporter/collector/logs.go
```diff
@@ -141,6 +161,17 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld pdata.Logs) error {
...
 
+func (l logMapper) getLogName(log pdata.LogRecord) (string, error) {
+	logNameAttr, exists := log.Attributes().Get("com.google.logName")
+	if exists {
+		return logNameAttr.AsString(), nil
+	}
+	if len(l.cfg.LogConfig.DefaultLogName) > 0 {
+		return l.cfg.LogConfig.DefaultLogName, nil
+	}
+	return "", fmt.Errorf("no default log name set")
```
dashpole, line 172:
```suggestion
	return "", fmt.Errorf("no log name provided.  Set the 'default_log_name' option, or add the 'com.google.logName' resource attribute to set a log name.")
```
Maybe provide users a bit more info here.

dashpole (approved):
lgtm after nits.

exporter/collector/config.go
damemi: 👍

exporter/collector/logs.go
damemi: updated, except changed to just `attribute` (I spent time trying to get the `resourceAttributes` processor to work until I realized they are different)

exporter/collector/logs.go
```diff
@@ -141,6 +161,17 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld pdata.Logs) error {
...
 
+func (l logMapper) getLogName(log pdata.LogRecord) (string, error) {
+	logNameAttr, exists := log.Attributes().Get("com.google.logName")
+	if exists {
+		return logNameAttr.AsString(), nil
+	}
+	if len(l.cfg.LogConfig.DefaultLogName) > 0 {
+		return l.cfg.LogConfig.DefaultLogName, nil
+	}
+	return "", fmt.Errorf("No log name provided.  Set the 'default_log_name' option, or add the 'com.google.logName' attribute to set a log name.")
```
dashpole, line 172:
```suggestion
	return "", fmt.Errorf("no log name provided.  Set the 'default_log_name' option, or add the 'com.google.logName' attribute to set a log name.")
```
Error strings should start with lowercase.

exporter/collector/logs.go
dashpole: sgtm

exporter/collector/logs.go
damemi: That's what I thought, but I thought it failed CI linting for having a lowercase (or maybe I misread the failure)

PR #363, opened 2022-04-21
Batch multiple entries into a maximum request size

damemi:
This adds batching for an entire request, grouping log records into requests below a configurable size

Follow up will add request sizing for an individual entry

xref b/228844902

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -104,13 +104,6 @@ type ResourceFilter struct {
 
 type LogConfig struct {
 	DefaultLogName string `mapstructure:"default_log_name"`
-
-	// MaxBatchSize defines the maximum size in bytes for a batched request of log entries
-	// 0 defaults to the Cloud Logging API's maximum request size (10 MB)
-	MaxBatchSize int `mapstructure:"max_batch_size"`
-	// MaxEntrySize defines the maximum size in bytes for an individual log entry.
-	// 0 uses the default as defined by the Cloud Logging API (256 KB)
-	MaxEntrySize int `mapstructure:"max_entry_size"`
 }
 
 // Known metric domains. Note: This is now configurable for advanced usages.
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -104,6 +104,13 @@ type ResourceFilter struct {
 
 type LogConfig struct {
 	DefaultLogName string `mapstructure:"default_log_name"`
+
+	// MaxBatchSize defines the maximum size in bytes for a batched request of log entries
+	// 0 defaults to the Cloud Logging API's maximum request size (10 MB)
+	MaxBatchSize int `mapstructure:"max_batch_size"`
+	// MaxEntrySize defines the maximum size in bytes for an individual log entry.
+	// 0 uses the default as defined by the Cloud Logging API (256 KB)
+	MaxEntrySize int `mapstructure:"max_entry_size"`
 }
 
 // Known metric domains. Note: This is now configurable for advanced usages.
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -31,9 +31,15 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/genproto/googleapis/api/monitoredres"
 	logpb "google.golang.org/genproto/googleapis/logging/v2"
+	"google.golang.org/protobuf/proto"
 )
 
-const HTTPRequestAttributeKey = "com.google.httpRequest"
+const (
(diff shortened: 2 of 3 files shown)


exporter/collector/logs.go
```diff
@@ -109,8 +115,39 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+		}
+	}
+
+	if len(errors) > 0 {
+		return multierr.Combine(errors...)
+	}
+	return nil
+}
+
+func (l logMapper) createBatches(ld plog.Logs) ([][]*logpb.LogEntry, error) {
```
dashpole, line 138: nit: it might be cleaner to keep the logic simpler here by returning a `[]*logpb.LogEntry`, and then send it in chunks like the monitoring batching code does: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/metrics.go#L208

damemi: Yeah that is how the Stanza implementation did it too

exporter/collector/config.go
```diff
@@ -104,6 +104,13 @@ type ResourceFilter struct {
 
 type LogConfig struct {
 	DefaultLogName string `mapstructure:"default_log_name"`
+
+	// MaxBatchSize defines the maximum size in bytes for a batched request of log entries
+	// 0 defaults to the Cloud Logging API's maximum request size (10 MB)
+	MaxBatchSize int `mapstructure:"max_batch_size"`
+	// MaxEntrySize defines the maximum size in bytes for an individual log entry.
+	// 0 uses the default as defined by the Cloud Logging API (256 KB)
+	MaxEntrySize int `mapstructure:"max_entry_size"`
```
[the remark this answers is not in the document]
damemi: This is a good question, I assumed this was a requirement based on the Stanza exporter and included it in the doc. I'll remove it for now since it's easy enough to reimplement if we can get more details on a use case for this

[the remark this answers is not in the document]
damemi: Removed these options

exporter/collector/logs.go
damemi: Updated, we can't use the direct indexing that metrics does since this is based on size rather than count. But I think I was still able to refactor it to do this more in-place. Please let me know what you think

exporter/collector/logs.go
```diff
@@ -109,8 +116,51 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+	errors := []error{}
+	entry := 0
+	currentBatchSize := 0
+	// Send entries in WriteRequest chunks
+	for len(entries) > 0 {
+		// TODO(damemi): Add integration test for batch request processing
+		if entry < len(entries) && currentBatchSize+proto.Size(entries[entry]) < defaultMaxRequestSize {
+			// if adding the current entry to the current batch doesn't go over the request size,
+			// increase the index and account for the new request size, then continue
+			currentBatchSize = currentBatchSize + proto.Size(entries[entry])
```
dashpole, line 133: can we combine into a single proto.Size call per-entry?

dashpole, line 133: nit: use `+=` instead of X = X +

exporter/collector/logs.go
```diff
@@ -109,8 +116,51 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
...
+		// if the current entry goes over the request size (or we have gone over every entry, i.e. index=len),
+		// write the list up to but not including the current entry's index
+		_, err := l.writeLogEntries(ctx, entries[:entry])
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		// if there are still entries to parse, reassign entries to a
+		// shortened list starting at the current index
+		if entry < len(entries) {
```
dashpole, line 147: Is this check is necessary?  If there aren't any entries left to parse, won't `len(entries) > 0` above return false?

dashpole (approved):
lgtm after nits

exporter/collector/logs.go
dashpole: ah, yeah.  That is unfortunate, but I think the current code is good

exporter/collector/logs.go
damemi: Not necessary, think I was off by one protecting for panics

exporter/collector/logs.go
damemi: Yep, I tried being lazy to guard for panics here by throwing the `entry < len(entries)` check first into this if block but I just broke that out

PR #365, opened 2022-04-22
Add missing otel attributes to Spans exported to Cloud Trace

### LINKED ISSUE #332: Add span attributes for library name and version
Based on our exporter spec, they should be added as `otel.instrumentation_library.name` and `otel.instrumentation_library.version`.   Although, I believe that plan should be altered by https://github.com/open-telemetry/opentelemetry-specification/pull/2276, which changed them to suggest `otel.scope.name` and `otel.scope.version`.

jsuereth:
Fixes #332

Turns out we never had a "raw" `Span -> Span` translation test before now.  Imported a few testing libs to make this a thing, and may expand on unit testing other span conversion aspects.

- Creates new "generic" monitored resource mapping internal library
- Modify Collector exporter to leverage this module with adapters.
- Leverage monitored resource mapping library to annotate trace spans with monitored resource labels.
- Add InstrumnetationLibrary labels to span attributes.


exporter/collector/integrationtest/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/internal/integrationtest
 
-go 1.17
+go 1.16
```
dashpole, line 3: can we leave this at 1.17?

exporter/trace/go.mod
```diff
@@ -31,8 +33,11 @@ require (
 	github.com/googleapis/gax-go/v2 v2.2.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	go.opencensus.io v0.23.0 // indirect
+	go.opentelemetry.io/collector/model v0.49.0 // indirect
```
dashpole, line 36: This is somewhat unfortunate... The collector and the go SDK have two different packages for semconv.  E.g. https://github.com/open-telemetry/opentelemetry-go/tree/main/semconv/v1.9.0 vs https://github.com/open-telemetry/opentelemetry-collector/tree/main/semconv/v1.9.0.  I think we should probably use `go.opentelemetry.io/otel/semconv/v1.9.0`, rather than the collector one.  It seems less strange for the collector to depend on the SDK than for the SDK exporter to depend on the collector.

internal/resourcemapping/resourcemapping.go
```diff
@@ -0,0 +1,186 @@
...
+type GceResource struct {
+	Type   string
+	Labels map[string]string
+}
+
+type ReadOnlyAttributes interface {
+	GetString(string) (string, bool)
+}
+
+// Converts from a set of OTEL resource attributes into a
```
dashpole, line 123:
```suggestion
// ResourceAttributesToMonitoredResource converts from a set of OTEL resource attributes into a
```
Function comments should start with the name of the function.

internal/resourcemapping/resourcemapping.go
```diff
@@ -0,0 +1,186 @@
...
+		},
+	}
+)
+
+type GceResource struct {
+	Type   string
+	Labels map[string]string
+}
+
+type ReadOnlyAttributes interface {
```
dashpole, line 119: nit: add a comment starting with // ReadOnlyAttributes ...

dashpole (approved):
lgtm after import is fixed.

exporter/trace/go.mod
jsuereth: Ah, good point.  I didn't even realize I was pulling the wrong one....

exporter/collector/integrationtest/go.mod
jsuereth: Sure.   It's kinda wierd how half the project is 1.16 and the rest 1.17.  Should we have the whole thing be 1.17 or is this a backward compatible thing?

dashpole: We want the whole thing on 1.17, which I thought I had already done (but must've missed one).  Feel free to include that in this PR.

exporter/trace/go.mod
jsuereth: Hmm, I'm having trouble finding a published version from the SDK -

https://pkg.go.dev/go.opentelemetry.io/otel/semconv/v1.9.0?tab=versions

```shell
$ go get go.opentelemetry.io/otel/semconv/v1.9.0
go get: module go.opentelemetry.io/otel@upgrade found (v1.6.3), but does not contain package go.opentelemetry.io/otel/semconv/v1.9.0
```

Should I move back to a version that exists?   How often does go publish these?

exporter/collector/integrationtest/go.mod
jsuereth: Fixed

exporter/trace/go.mod
jsuereth: Migrated to v1.7.0 (latest released version) of semconv.

internal/resourcemapping/resourcemapping.go
jsuereth: added.

PR #368, opened 2022-04-26
Add logging integration tests

damemi:
This adds unit and integration tests for the logging exporter

xref b/225001760


exporter/collector/integrationtest/cmd/recordfixtures/log.go
```diff
@@ -0,0 +1,77 @@
...
+
+	for _, test := range integrationtest.LogsTestCases {
+		if test.Skip {
+			continue
+		}
+		func() {
+			logs := test.LoadOTLPLogsInput(t, startTime)
+			testServerExporter := testServer.NewExporter(ctx, t, test.CreateConfig())
+
+			require.NoError(t, testServerExporter.PushLogs(ctx, logs), "failed to export metrics to local test server")
```
dashpole, line 67:
```suggestion
			require.NoError(t, testServerExporter.PushLogs(ctx, logs), "failed to export traces to local test server")
```

exporter/collector/integrationtest/logging_test_server.go
```diff
@@ -0,0 +1,113 @@
...
+) *collector.LogsExporter {
+
+	cfg.LogConfig.ClientConfig.Endpoint = l.Endpoint
+	cfg.LogConfig.ClientConfig.UseInsecure = true
+	cfg.ProjectID = "fakeprojectid"
+
+	exporter, err := collector.NewGoogleCloudLogsExporter(
+		ctx,
+		cfg,
+		zap.NewNop(),
```
dashpole, line 108: I know metrics does this, but would it be more helpful to have log messages during test runs?  Should we use `zap.NewDevelopment()` instead?

exporter/collector/integrationtest/testcase.go
```diff
@@ -48,6 +50,92 @@ var (
 	}
 )
 
+type LogsTestCase struct {
```
dashpole, line 53: Is this exactly the same as MetricsTestCase?  I'm all for refactoring later, but for identical structs, it is probably worth unifying them here.

damemi: done, they have all similar functions that I just renamed for now but can probably be refactored to a common interface later

exporter/collector/integrationtest/cmd/recordfixtures/main.go
```diff
@@ -45,25 +46,66 @@ func (t *FakeTesting) Helper()      {}
...
-	t := &FakeTesting{}
 	ctx := context.Background()
 	endTime := time.Now()
 	startTime := endTime.Add(-time.Second)
+	t := &FakeTesting{}
+
+	for _, arg := range os.Args[1:] {
+		switch arg {
+		case "logs":
+			logs(ctx, t, endTime)
```
dashpole, line 57: Is there a reason why we only want to record one of them?  Should we just record both metrics and logs when this is run?

exporter/collector/integrationtest/logging_test_server.go
dashpole: Or we could use NewProduction like the other test below does

exporter/collector/integrationtest/testdata/fixtures/logs_apache_access_expected.json
```diff
@@ -0,0 +1,341 @@
...
+            "remoteIp": "127.0.0.1",
+            "protocol": "HTTP/1.1"
+          }
+        },
+        {
+          "logName": "projects/fakeprojectid/logs/my-log-name-foo",
+          "resource": {
+            "type": "gce_instance",
+            "labels": {
+              "instance_id": "",
```
dashpole, line 190: it would be great to use input data (you can manually update it if you need to) which has the resource attributes required to fill these resource labels in.

dashpole (approved):
feedback can be addressed in follow-ups, if you would prefer that

exporter/collector/integrationtest/cmd/recordfixtures/main.go
damemi: Good point

PR #372, opened 2022-04-29
Merge collector_logging into main

damemi:
/hold


damemi:
Replacing with https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/373 to just manually pick the logic

PR #379, opened 2022-05-03
Add struct tag for sum of squared deviations estimate and factory test.

dashpole:
This should fix failures in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9679, and adds a test to make sure we don't hit them again.


aabmass (approved):
🚢

PR #381, opened 2022-05-04
collector: Sanitize invalid UTF-8 in label values.

igorpeshansky:
Invalid UTF-8 values in Cloud Monitoring data structures will break gRPC, causing it to fail to encode the resulting request protobufs (`grpc: error while marshaling: string field contains invalid UTF-8`). I added the flag in case it was preferable to fail rather than ingest modified values, but we could also unconditionally sanitize those strings instead.

See also #382.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -106,6 +106,9 @@ type MetricConfig struct {
 	// so we don't send it by default, and don't expose it to users. For some uses, it is
 	// expected, however.
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
+
+	// This enables sanitization of UTF-8 characters in labels.
+	EnforceUTF8 *bool `mapstructure:"enforce_utf8"`
 }
 
 type ResourceFilter struct {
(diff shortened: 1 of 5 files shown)


exporter/collector/config.go
```diff
@@ -106,6 +106,9 @@ type MetricConfig struct {
 	// so we don't send it by default, and don't expose it to users. For some uses, it is
 	// expected, however.
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
+
+	// This enables sanitization of UTF-8 characters in labels.
+	EnforceUTF8 *bool `mapstructure:"enforce_utf8"`
```
dashpole, line 111: Same comment as the other PR.  Do we need this?

dashpole, line 111: Can this be a bool instead of `*bool` (if we keep it)?

We can set the default to true in `DefaultConfig()` below.

igorpeshansky: As requested in the other PR, removed the config option altogether for now.

igorpeshansky (commented):
Addressed review comments. PTAL.

dashpole (approved):

PR #383, opened 2022-05-04
Add GCP resource detector

dashpole:
This implements resource detection for GCE, GKE, Cloud Run, Cloud Functions, and App Engine.  It doesn't split out logic into a non-otel package, which can be done as a follow-up.  

Breaking changes from the [opentelemetry-go-contrib gcp detector](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/gcp):

* Detectors for each platform are combined into a single detector
* No longer supports detecting k8s.pod.name, container.name, k8s.namespace.name.  These are expected to be set outside of this resource detector.
* No longer sets service.namespace to "cloud-run-managed"
* Sets faas.name instead of service.name in cloud run
* Sets faas.id instead of service.instance.id in cloud run

Additions:

* Adds detector for Google App Engine
* Now correctly sets cloud.platform on GKE and Cloud Run
* Now detects cloud.availability_zone/region when running on GKE with workload identity enabled (the GCE metadata server is not accessible in that scenario)
* Now detects faas.version on cloud run and cloud functions.


dashpole:
I raised this at the Go sig.  I'll plan to split this between our repo, and upstream.  This repo should not depend on semconv.

dashpole:
Closing in favor of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/394

PR #389, opened 2022-05-11
Log exporter individual log splitting

damemi:
This sizes and splits individual log entries based on the default maximum quotas per cloud logging

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -318,7 +318,8 @@ func (l logMapper) logToSplitEntries(
 
 	httpRequestAttr, ok := log.Attributes().Get(HTTPRequestAttributeKey)
 	if ok {
-		httpRequest, err := l.parseHTTPRequest(httpRequestAttr)
+		var httpRequest *logging.HTTPRequest
+		httpRequest, err = l.parseHTTPRequest(httpRequestAttr)
 		if err != nil {
 			l.obs.log.Debug("Unable to parse httpRequest", zap.Error(err))
 		}
@@ -371,10 +372,9 @@ func (l logMapper) logToSplitEntries(
 			endIndex = int(math.Floor((float64(i+1) / splits) * float64(len(payloadString))))
 		}
 		return entries, logName, nil
-	} else {
-		entry.Payload = payload
 	}
 
+	entry.Payload = payload
 	return []logging.Entry{entry}, logName, nil
 }
 
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -25,13 +25,15 @@ import (
 
 	"cloud.google.com/go/logging"
 	loggingv2 "cloud.google.com/go/logging/apiv2"
+	"google.golang.org/genproto/googleapis/api/monitoredres"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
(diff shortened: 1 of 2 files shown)


exporter/collector/logs.go
```diff
@@ -283,41 +318,80 @@ func (l logMapper) logToEntry(
...
+		}
+		return entries, logName, nil
+	}
+
+	entry.Payload = payload
+	return []logging.Entry{entry}, logName, nil
 }
 
-func parseEntryPayload(logBody pcommon.Value) (interface{}, error) {
+func parseEntryPayload(logBody pcommon.Value, maxEntrySize int) (interface{}, float64, error) {
```
dashpole, line 381: Should we return an int instead of a float for the number of splits?

exporter/collector/logs.go
```diff
@@ -283,41 +318,80 @@ func (l logMapper) logToEntry(
...
+
+		// Start by assuming all splits will be even (this may not be the case)
+		startIndex := 0
+		endIndex := int(math.Floor((1.0 / splits) * float64(len(payloadString))))
+		for i := 1; i <= int(splits); i++ {
+			currentSplit := payloadString[startIndex:endIndex]
+
+			// If the current split is larger than the entry size, iterate until it is within the max
+			// (This may happen since not all characters are exactly 1 byte)
+			for len([]byte(currentSplit)) > l.maxEntrySize {
```
dashpole, line 363: This seems like it could be expensive in cases with high numbers of non-1-byte characters...  But it is probably fine to start

exporter/collector/logs.go
```diff
@@ -235,17 +262,25 @@ func (l logMapper) getLogName(log plog.LogRecord) (string, error) {
...
-) (logging.Entry, error) {
+) ([]logging.Entry, string, error) {
 	entry := logging.Entry{
 		Resource: mr,
 	}
 
+	// We can't just set logName on these entries otherwise the conversion to internal will fail
+	// We also need the logName here to be able to accurately calculate the overhead of entry
+	// metadata in case the payload needs to be split between multiple entries.
+	logName, err := l.getLogName(log)
```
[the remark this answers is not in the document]
damemi: makes sense, done

exporter/collector/logs.go
damemi: yeah, it's the same number of casts either way but it's a little easier to read as an int

exporter/collector/logs.go
damemi: I was thinking about this, the best I could come up with is a binary search. think that would optimize it?

dashpole: yes, but we can leave that as a follow-up if neccessary

exporter/collector/logs.go
```diff
@@ -182,38 +190,67 @@ func (l logMapper) createEntries(ld plog.Logs) ([]*logpb.LogEntry, error) {
...
 					continue
 				}
 
-				internalLogEntry, err := logging.ToLogEntry(entry, fmt.Sprintf("projects/%s", l.cfg.ProjectID))
-				if err != nil {
-					errors = append(errors, err)
-					continue
-				}
-				logName, err := l.getLogName(log)
+				splitEntries, logName, err := l.logToSplitEntries(
```
dashpole, line 202: we don't need to return logName from this anymore, right?

dashpole (approved):

PR #393, opened 2022-05-12
Update golangci-lint and reorder structs to satisfy the linter

dashpole:
Updating the linter produced warnings about linter deprecations.  `govet` also added a new linter error for struct alignment.

Reordering fields in public structs can be breaking if consumers initialize a struct without using named member arguments.  Public structs this impacts includes:

* exporter/collector.Config
* exporter/collector.ClientConfig
* exporter/collector.MetricConfig
* exporter/collector.LogConfig
* exporter/collector/integrationtest.MetricsTestServer
* exporter/collector/integrationtest.LogsTestServer
* exporter/collector/integrationtest.TestCase


exporter/collector/config.go
```diff
@@ -69,58 +66,19 @@ type AttributeMapping struct {
...
-	// customizing the naming of metrics. baseName already includes type
-	// suffixes for summary metrics, but does not (yet) include the domain prefix
-	GetMetricName func(baseName string, metric pmetric.Metric) (string, error)
-
-	// MapMonitoredResource is not exposed as an option in the configuration, but
-	// can be used by other exporters to extend the functionality of this
-	// exporter. It allows overriding the function used to map otel resource to
-	// monitored resource.
-	MapMonitoredResource func(pcommon.Resource) *monitoredrespb.MonitoredResource
+	GetMetricName                    func(baseName string, metric pmetric.Metric) (string, error)
```
bogdandrutu, line 69: Not sure about your rules, but this may be a breaking change if someone initializes a struct without using named member arguments.

dashpole: Yes, we do consider that a breaking change.  The exporter/collector package is currently unstable (and is only used by collector-contrib to my knowledge), so this is acceptable for now.  I called that out more clearly in the description

damemi (approved):

PR #396, opened 2022-05-19
Update end-to-end test server to implement resource detection scenario

dashpole:
This PR:

* Fixes the resource detection module name
* Changes naming of resource detection functions to match semantic convention names
* Adds a zone/region function for app engine
* Falls-back to region when zone is not found (needed to pass resource detection tests).
* Implements the detectResource scenario.

The e2e test below passes the resourceDetection scenario.


dashpole:
@aabmass where is "local" actually being run?  It seems to be on GCP somewhere, as the metadata server is available, but it doesn't appear to be on GCE, since machine-type isn't being found as an attribute.

aabmass:
> @aabmass where is "local" actually being run? It seems to be on GCP somewhere, as the metadata server is available, but it doesn't appear to be on GCE, since machine-type isn't being found as an attribute.

It just connects to the local docker daemon, so in CI it's running on Cloud Build. We are doing this so that the inner container can get ADC
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/749b5eac90b76b6d47254f8067f067213c4f024a/cloudbuild-e2e-local.yaml#L33

See https://cloud.google.com/build/docs/build-config-file-schema#network.

dashpole:
I ended up adding a function to detect when we are on GCE.  That should help filter out other platforms (like cloud build) that we haven't implemented detection for.

dashpole:
All passing now!

e2e-test-server/endtoendserver/server.go
```diff
@@ -133,21 +137,113 @@ func (s *Server) respond(ctx context.Context, testID string, res *response) erro
 	return err
 }
 
-func newTraceProvider() (*sdktrace.TracerProvider, error) {
+func newTraceProvider(res *resource.Resource) (*sdktrace.TracerProvider, error) {
```
aabmass, line 140:
```suggestion
func newTracerProvider(res *resource.Resource) (*sdktrace.TracerProvider, error) {
```

aabmass (approved):

e2e-test-server/endtoendserver/server.go
dashpole: done

PR #398, opened 2022-05-20
optionally configure service account impersonation

jmcarp:
Note: this replaces https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/10173 per a suggestion from @dashpole.

GCP supports authentication using short-lived credentials obtained by
"impersonating" a service account. This is useful for cross-project
access, for example: a service account in one project is trusted to
impersonate a service account in another project and export metrics on
its behalf. This patch adds support for service account impersonation
for metrics, logs, and traces.

See https://cloud.google.com/iam/docs/impersonating-service-accounts for
details.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -30,11 +30,12 @@ const (
 
 // Config defines configuration for Google Cloud exporter.
 type Config struct {
-	LogConfig    LogConfig    `mapstructure:"log"`
-	ProjectID    string       `mapstructure:"project"`
-	UserAgent    string       `mapstructure:"user_agent"`
-	TraceConfig  TraceConfig  `mapstructure:"trace"`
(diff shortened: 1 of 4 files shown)


dashpole:
you can run `make fieldalignment` to fix the linter issue.  It might remove comments, so make sure to add any of those back in

exporter/collector/config.go
```diff
@@ -113,6 +114,12 @@ type MetricConfig struct {
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
 }
 
+type ImpersonateConfig struct {
```
dashpole, line 117: Can you add a comment for this struct with basically the same content as in your PR title?

dashpole (commented):
looks good overall

dashpole (approved):

PR #399, opened 2022-05-20
Allow overriding location, cluster, and namespace using metric labels

dashpole:
With GMP, if you set "location", "cluster" or "namespace" labels on your metrics, they are [moved](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/c4e6189ecd21fcc5333ec4c04f116b98701ff8ff/pkg/export/series_cache.go#L520) to the monitored resource from the set of labels.

To achieve the same behavior in the otel collector, use a processor to move those labels from metric to resource, and then map those from otel resource to monitored resource.  They take precedence over other ways of setting those labels (i.e. a "namespace" label on a KSM metric will be used over the namespace of the KSM exporter.

cc @xichen2020

### DIFF
--- a/exporter/collector/googlemanagedprometheus/monitoredresource.go
+++ b/exporter/collector/googlemanagedprometheus/monitoredresource.go
@@ -29,26 +29,24 @@ func MapToPrometheusTarget(res pcommon.Resource) *monitoredrespb.MonitoredResour
 	if serviceNamespace != "" {
 		job = serviceNamespace + "/" + job
 	}
-	location := getStringOrEmpty(attrs, semconv.AttributeCloudAvailabilityZone)
-	if location == "" {
-		// fall back to region if we don't have a zone.
-		location = getStringOrEmpty(attrs, semconv.AttributeCloudRegion)
-	}
 	return &monitoredrespb.MonitoredResource{
 		Type: "prometheus_target",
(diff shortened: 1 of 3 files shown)


exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -29,26 +29,24 @@ func MapToPrometheusTarget(res pcommon.Resource) *monitoredrespb.MonitoredResour
...
-		// fall back to region if we don't have a zone.
-		location = getStringOrEmpty(attrs, semconv.AttributeCloudRegion)
-	}
 	return &monitoredrespb.MonitoredResource{
 		Type: "prometheus_target",
 		Labels: map[string]string{
-			"location":  location,
-			"cluster":   getStringOrEmpty(attrs, semconv.AttributeK8SClusterName),
-			"namespace": getStringOrEmpty(attrs, semconv.AttributeK8SNamespaceName),
+			"location":  getStringOrEmpty(attrs, "location", semconv.AttributeCloudAvailabilityZone, semconv.AttributeCloudRegion),
```
xichen2020, line 35: nit: Make "location" etc consts.

xichen2020 (approved):
LGTM

Issue #364, opened 2022-04-22
Support writing to multiple projects if project is set in resource


dashpole:
A user has a use-case in which metrics are sent with a project ID set in metadata.  They don't know which projects they will need to send to ahead of time.  This would likely have implications for batching, as I don't believe metrics for multiple projects can be batched together.

PR #402, opened 2022-05-25
Support writing to multiple gcp projects

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/364.

This keeps things simple.  As we learned with https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/358, we can't do batching in the exporter asynchronously without losing self-observability metrics and the ability to return errors.  This PR splits the batch of metrics/traces/logs by project, and then sends each of those batches separately.

The precedence for determining project ID is:  `resource > config > credentials`

So if config is set to `project-1`, telemetry with `gcp.project.id=project-2` will be sent to project-2, and telemetry without  that attribute (or with `gcp.project.id=project-1`) will be sent to project-1.

One interesting thing that came up is that it is now possible to create the same metric name in two different projects.  To accommodate this, the CreateMetricDescriptor cache now uses `project/metric_type` as the key instead of just `metric_type`.

This does not implement multi-project mapping for the SDK metric exporter, as that will be rewritten once the new go SDK is released.


exporter/collector/integrationtest/metrics_integration_test.go
```diff
@@ -61,6 +64,7 @@ func TestIntegrationMetrics(t *testing.T) {
 		t.Run(test.Name, func(t *testing.T) {
 			test.SkipIfNeeded(t)
 			metrics := test.LoadOTLPMetricsInput(t, startTime, endTime)
+			setSecondProjectInMetrics(t, metrics)
```
aabmass, line 67: IIUC, this will override all of the metrics to send to the second project. Does anything get written to the first project anymore?

aabmass (approved):

exporter/collector/integrationtest/metrics_integration_test.go
dashpole: This will override all _existing_ `gcp.project.id` attributes in the input to send to the second project.  Everything that doesn't have a `gcp.project.id` attribute still goes to the first project.

exporter/collector/integrationtest/logs_integration_test.go
```diff
@@ -69,3 +73,18 @@ func TestIntegrationLogs(t *testing.T) {
...
 }
+
+// setSecondProjectInLogs overwrites the gcp.project.id resource attribute
+// with the contents of the SECOND_PROJECT_ID environment variable. This makes
+// the integration test send those logs to the specified second project.
+func setSecondProjectInLogs(t *testing.T, logs plog.Logs) {
+	for i := 0; i < logs.ResourceLogs().Len(); i++ {
+		rl := logs.ResourceLogs().At(i)
+		secondProject := os.Getenv(secondProjectEnv)
+		require.NotEmpty(t, secondProject, "set the SECOND_PROJECT_ID environment to run this test")
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/integrationtest/metrics_integration_test.go
```diff
@@ -72,3 +76,18 @@ func TestIntegrationMetrics(t *testing.T) {
...
 }
+
+// setSecondProjectInMetrics overwrites the gcp.project.id resource attribute
+// with the contents of the SECOND_PROJECT_ID environment variable. This makes
+// the integration test send those metrics to the specified second project.
+func setSecondProjectInMetrics(t *testing.T, metrics pmetric.Metrics) {
+	for i := 0; i < metrics.ResourceMetrics().Len(); i++ {
+		rm := metrics.ResourceMetrics().At(i)
+		secondProject := os.Getenv(secondProjectEnv)
+		require.NotEmpty(t, secondProject, "set the SECOND_PROJECT_ID environment to run this test")
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/integrationtest/metrics_integration_test.go
dashpole: I updated the comment on `setSecondProjectInMetrics` and `setSecondProjectInLogs` to make that clearer

PR #403, opened 2022-05-25
Test project ID in fixtures

dashpole:
We were previously normalizing the project ID because it was set via an env var for the fixture-based tests.  We can test this by simply setting it to a constant instead, and removing the normalization.  This will make it easier to test multi-project exporting: #402.

This also removes some extra cloud.account.id resource attributes from fixtures.

### DIFF
--- a/exporter/collector/integrationtest/testcase.go
+++ b/exporter/collector/integrationtest/testcase.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"sort"
 	"testing"
 	"time"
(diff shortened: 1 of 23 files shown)


exporter/collector/integrationtest/testcase.go
```diff
@@ -349,7 +344,7 @@ func (tc *TestCase) SkipIfNeeded(t testing.TB) {
 func (tc *TestCase) CreateMetricConfig() collector.Config {
 	cfg := collector.DefaultConfig()
 	// If not set it will use ADC
```
aabmass, line 346: can remove this comment

aabmass (approved):
🚢

exporter/collector/integrationtest/testcase.go
dashpole: done

PR #404, opened 2022-05-26
Normalize timestamps in log tests

dashpole:
This way the time doesn't change each time the fixtures are regenerated.  This matches what we do in metric tests


damemi (approved):
thanks, this bugged me too but I wasn't sure how to fix it

Issue #405, opened 2022-05-31
Concurrent map read and map write in collector exporter


dashpole:
Two different occurrences:

First
![image](https://user-images.githubusercontent.com/3262098/171203397-6f7721ad-90f6-4080-9c03-14bf1e96435b.png)

Second
![image](https://user-images.githubusercontent.com/3262098/171203489-8292eec7-9d6b-42e2-92f0-fead59e16487.png)

PR #417, opened 2022-06-09
Make the datapoint storage cache safe for concurrent use

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/405

This adds one RW lock for each cache.  For GetFoo() functions, it takes a RLock, and for SetFoo() functions, it takes a write Lock.  However, since the "used" bool is set in GetFoo() functions (it indicates if the item has been touched since the last GC), this still produced a data race.  I could have switched to a full write-lock for Get functions, but thought it would be more performant to use a concurrency-safe bool.  That also simplifies the implementation of GC using Compare and swap.

I added tests for concurrency to verify that there are no data races.

### DIFF
--- a/exporter/collector/internal/datapointstorage/datapointcache.go
+++ b/exporter/collector/internal/datapointstorage/datapointcache.go
@@ -83,7 +83,6 @@ func (c *Cache) GetNumberDataPoint(identifier string) (*pmetric.NumberDataPoint,
 	point, found := c.numberCache[identifier]
 	if found {
 		point.used.Store(true)
-		c.numberCache[identifier] = point
 	}
 	return point.point, found
 }
@@ -103,7 +102,6 @@ func (c *Cache) GetSummaryDataPoint(identifier string) (*pmetric.SummaryDataPoin
 	point, found := c.summaryCache[identifier]
 	if found {
 		point.used.Store(true)
-		c.summaryCache[identifier] = point
 	}
 	return point.point, found
 }
@@ -123,7 +121,6 @@ func (c *Cache) GetHistogramDataPoint(identifier string) (*pmetric.HistogramData
 	point, found := c.histogramCache[identifier]
 	if found {
 		point.used.Store(true)
-		c.histogramCache[identifier] = point
 	}
 	return point.point, found
 }
@@ -143,7 +140,6 @@ func (c *Cache) GetExponentialHistogramDataPoint(identifier string) (*pmetric.Ex
 	point, found := c.exponentialHistogramCache[identifier]
 	if found {
 		point.used.Store(true)
-		c.exponentialHistogramCache[identifier] = point
 	}
 	return point.point, found
 }
@@ -164,8 +160,10 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
 		// garbage collect the numberCache
 		c.numberLock.Lock()
 		for id, point := range c.numberCache {
-			// If used is true, swap it to false. Otherwise, delete the point.
-			if !point.used.CAS(true, false) {
+			// for points that have been used, mark them as unused
+			if point.used.Load() {
+				point.used.Store(false)
+			} else {
 				// for points that have not been used, delete points
 				delete(c.numberCache, id)
 			}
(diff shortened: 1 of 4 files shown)


dashpole:
@jsuereth you had mentioned wanting to review this at standup.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -71,118 +77,130 @@ func NewCache(shutdown <-chan struct{}) Cache {
...
 // or not it was found
-func (c Cache) GetSummaryDataPoint(identifier string) (*pmetric.SummaryDataPoint, bool) {
+func (c *Cache) GetSummaryDataPoint(identifier string) (*pmetric.SummaryDataPoint, bool) {
+	c.summaryLock.RLock()
+	defer c.summaryLock.RUnlock()
 	point, found := c.summaryCache[identifier]
 	if found {
-		point.used = true
+		point.used.Store(true)
 		c.summaryCache[identifier] = point
```
jsuereth, line 106: Does this need a write lock?

IIUC - if summaryCache is storing references/pointers then just `point.used.Store(true)` is all you need here.  Otherwise you probably want a write lock on summary cache for consistency.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -71,118 +77,130 @@ func NewCache(shutdown <-chan struct{}) Cache {
...
 // or not it was found
-func (c Cache) GetHistogramDataPoint(identifier string) (*pmetric.HistogramDataPoint, bool) {
+func (c *Cache) GetHistogramDataPoint(identifier string) (*pmetric.HistogramDataPoint, bool) {
+	c.histogramLock.RLock()
+	defer c.histogramLock.RUnlock()
 	point, found := c.histogramCache[identifier]
 	if found {
-		point.used = true
+		point.used.Store(true)
 		c.histogramCache[identifier] = point
```
jsuereth, line 126: Same concern here.   You're writing to histogram cache without a write lock, but I'm not sure you need to write to it.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -71,118 +77,130 @@ func NewCache(shutdown <-chan struct{}) Cache {
...
 // or not it was found
-func (c Cache) GetExponentialHistogramDataPoint(identifier string) (*pmetric.ExponentialHistogramDataPoint, bool) {
+func (c *Cache) GetExponentialHistogramDataPoint(identifier string) (*pmetric.ExponentialHistogramDataPoint, bool) {
+	c.exponentialHistogramLock.RLock()
+	defer c.exponentialHistogramLock.RUnlock()
 	point, found := c.exponentialHistogramCache[identifier]
 	if found {
-		point.used = true
+		point.used.Store(true)
 		c.exponentialHistogramCache[identifier] = point
```
jsuereth, line 146: Same nit as before.

jsuereth (commented):
A few concurrency nits / questions on standard Go usage.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -71,118 +77,130 @@ func NewCache(shutdown <-chan struct{}) Cache {
...
 		// garbage collect the numberCache
+		c.numberLock.Lock()
 		for id, point := range c.numberCache {
-			if point.used {
-				// for points that have been used, mark them as unused
-				point.used = false
-				c.numberCache[id] = point
-			} else {
+			// If used is true, swap it to false. Otherwise, delete the point.
+			if !point.used.CAS(true, false) {
```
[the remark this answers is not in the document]
dashpole: Yep.  I still need to set used to false if we are keeping it, but I can change to that.

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -71,118 +77,130 @@ func NewCache(shutdown <-chan struct{}) Cache {
...
 		// garbage collect the summaryCache
+		c.summaryLock.Lock()
 		for id, point := range c.summaryCache {
-			if point.used {
-				// for points that have been used, mark them as unused
-				point.used = false
-				c.summaryCache[id] = point
-			} else {
+			// If used is true, swap it to false. Otherwise, delete the point.
+			if !point.used.CAS(true, false) {
```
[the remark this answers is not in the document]
dashpole: done

jsuereth (approved):

Issue #410, opened 2022-06-06
Self-observability metrics should count individual points, rather than batches


dashpole:
The [googlecloudmonitoring/point_count](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/1f22cbd7de2d7f06806d78899b6dc2aaf3d46a01/exporter/collector/observability.go#L29) self-observability metric assumes that all metrics sent in an unsuccessful batch are have a failed status.

The message returned usually contains the number of successful vs failed points: `desc = total_point_count:200 success_point_count:196 errors:{status:{code:3} point_count:4}"} `.  We should see if it is possible to extract that information (in a non-fragile way) to correctly report the number of success vs failed points.

PR #418, opened 2022-06-09
Record failed points in point_count metric

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/410

### DIFF
--- a/exporter/collector/integrationtest/metric_test_server.go
+++ b/exporter/collector/integrationtest/metric_test_server.go
@@ -111,7 +111,7 @@ func (f *fakeMetricServiceServer) CreateTimeSeries(
 		code = codes.NotFound
 	}
 
-	statusResp := status.New(code, "foo")
+	statusResp := status.New(code, "FORCED_TEST_ERROR")
 	if code == codes.NotFound {
 		statusResp, _ = statusResp.WithDetails(&monitoringpb.CreateTimeSeriesSummary{
 			TotalPointCount:   int32(len(req.TimeSeries)),
--- a/exporter/collector/integrationtest/metric_test_server.go
+++ b/exporter/collector/integrationtest/metric_test_server.go
@@ -17,6 +17,7 @@ package integrationtest
 import (
 	"context"
 	"net"
+	"strings"
 	"sync"
 	"testing"
 
@@ -25,6 +26,8 @@ import (
 	"google.golang.org/genproto/googleapis/api/metric"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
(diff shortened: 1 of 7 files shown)


exporter/collector/integrationtest/metric_test_server.go
```diff
@@ -150,7 +159,6 @@ func (m *MetricsTestServer) NewExporter(
 ) *collector.MetricsExporter {
 	cfg.MetricConfig.ClientConfig.Endpoint = m.Endpoint
 	cfg.MetricConfig.ClientConfig.UseInsecure = true
-	cfg.ProjectID = "fakeprojectid"
```
damemi, line 153: removed this because it was redundant with the setting in [`testcase.CreateMetricConfig()`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/5ca7a23/exporter/collector/integrationtest/testcase.go#L360) and overwrote any attempt to set `ProjectID` with a `Configure()` func

damemi:
Spent a lot of time trying to get this tested by passing a desired return code through `context.WithValue` but no luck. Ended up settling on this approach instead

exporter/collector/integrationtest/metric_test_server.go
dashpole: Hmmm... was this required?  The test doesn't pass anymore

damemi: weird, it passes for me locally.

the failure also looks like it's just the entries aren't in the expected order: https://app.circleci.com/pipelines/github/GoogleCloudPlatform/opentelemetry-operations-go/1387/workflows/a7049ae8-6a76-4a78-9baa-a842d951eaee/jobs/1382?invite=true#step-105-585

could that just be the case?

dashpole: ahh... that makes sense

dashpole: thanks

exporter/collector/metrics.go
```diff
@@ -245,7 +245,17 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			for _, detail := range s.Details() {
+				if summary, ok := detail.(*monitoringpb.CreateTimeSeriesSummary); ok {
+					// if success != total, we're reporting a failed create and change to using the diff
+					if summary.TotalPointCount != summary.SuccessPointCount {
+						summaryPointCount = int(summary.TotalPointCount - summary.SuccessPointCount)
+					}
+				}
+			}
+
+			recordPointCountDataPoint(ctx, summaryPointCount, st)
```
dashpole, line 258: can we always record the number of successful points, even if a few failed?  It would be useful for creating an SLO, for example.

damemi: trying to add this I actually caught that `s.Details()` isn't returning anything (at least in our test cases). it was just passing because the total was only `1`. so I need to look at this more, but I will add that as part of it

exporter/collector/metrics.go
```diff
@@ -245,7 +245,20 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			succeededPoints := len(ts)
+			failedPoints := 0
+			for _, detail := range s.Details() {
+				if summary, ok := detail.(*monitoringpb.CreateTimeSeriesSummary); ok {
+					failedPoints = int(summary.TotalPointCount - summary.SuccessPointCount)
+					succeededPoints = int(summary.SuccessPointCount)
+				}
+			}
+
+			// always record the number of successful points
```
dashpole, line 257: Does recording 0 do anything (e.g. cause a new metric to show up/ not show up)?  My instinct is that we should always record both successful points and failed points, but i'm not sure what the impact is.

damemi: The problem is that there are many different failure codes, so we could send `0` but we would need to pick a failure code to go with it. Otherwise in the case of 0 failures we'll be sending "`X succeeded and 0 succeeded`" if we just used the code we parse from the success response

dashpole: ah, got it.  This works then.

Issue #206, opened 2021-08-30
Add support for Cloud Monitoring Distributions in the metrics exporter


ronilichtman:
The current implementation for `ValueRecorderInstrumentKind` is `lastvalue`. This prevents clients from having an effective way to chart latency metrics.

By populating the RPC request as described in https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#Distribution, it should be possible to implement these distribution metrics.

aabmass:
This last value behavior has been changed to histogram in the OTel metrics specification, but the Go SDK hasn't implemented it yet. This will probably take a while, but we will definitely send distribution metrics once it is ready.

dashpole:
The spec-compliant go SDK is tracked in https://github.com/orgs/open-telemetry/projects/22.  Once that is done, we will update this exporter--including supporting distributions.

PR #432, opened 2022-06-15
Add support for histograms in the SDK metric exporter

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/206

This borrows from the otelmetric exporter [handling of metric aggregation kinds](https://github.com/open-telemetry/opentelemetry-go/blob/1f5b159161e0ff54df4677284fa7d6718dc9186d/exporters/otlp/otlpmetric/internal/metrictransform/metric.go#L235), switches to the "standard" simple.NewWithHistogramDistribution selector to actually give our exporter histograms, and borrows from the collector exporters implementation of [histogramPoint](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/be0741488714937be40c696f720f4e45f1774788/exporter/collector/metrics.go#L547) for the translation logic.

I didn't implement exemplars in this exporter, which can be done as a follow-up.

This leaves the implementation of gauges and counters exactly as-is, even though there are a few improvements (it uses time.Now() as the end time for counters?) I noticed.

I can now get a heatmap when I run the example:

Screen Shot 2022-06-15 at 4 50 11 PM


ymotongpoo (approved):
@dashpole thank you for adding histogram support and updating old type of type checking. LGTM.

Issue #416, opened 2022-06-08
Update lint checks in CI


damemi:
We currently have at least one [unused function](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c567ee/exporter/collector/observability.go#L47), [duplicated](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c567eed530ce47839387e7cb415813b712b0900/exporter/collector/logs.go#L28-L29) [imports](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c567eed530ce47839387e7cb415813b712b0900/exporter/metric/metric.go#L37-L38), and other failures which are caught by tools like `staticcheck`:

```
opentelemetry-operations-go/exporter/collector$ staticcheck .
logs.go:28:2: package "google.golang.org/genproto/googleapis/api/monitoredres" is being imported more than once (ST1019)
	logs.go:29:2: other import of "google.golang.org/genproto/googleapis/api/monitoredres"
logs.go:181:37: should use make([]*logpb.LogEntry, 0) instead (S1019)
logs.go:343:34: error strings should not be capitalized (ST1005)
metrics.go:1150:28: func (*MetricsExporter).processItem is unused (U1000)
observability.go:47:6: func recordPointCount is unused (U1000)

opentelemetry-operations-go/exporter/metric$ staticcheck .
cloudmonitoring.go:62:16: error strings should not be capitalized (ST1005)
cloudmonitoring.go:65:16: error strings should not be capitalized (ST1005)
metric.go:37:2: package "google.golang.org/api/option" is being imported more than once (ST1019)
	metric.go:38:2: other import of "google.golang.org/api/option"
metric_test.go:27:2: package "go.opentelemetry.io/otel/sdk/metric/controller/basic" is being imported more than once (ST1019)
	metric_test.go:28:2: other import of "go.opentelemetry.io/otel/sdk/metric/controller/basic"

opentelemetry-operations-go/exporter/trace$ staticcheck .
cloudtrace.go:72:2: field location is unused (U1000)
cloudtrace.go:79:2: field batchSpanProcessorOptions is unused (U1000)
cloudtrace.go:210:6: func createKeyValueAttributes is unused (U1000)
```

Our lint ci should be updated to catch these. We could also make changes to break lint checks out of circleci/build into their own checks, and use Github Actions

damemi:
[...]

PR #438, opened 2022-06-22
Add unused code check in linter

damemi:
Ref https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/416

This adds a check for unused code. It also updates the `golangci.yml` config to skip a false positive, where the `secondProjectEnv` const defined in `testcase.go` is not detected as being used (it is called in 2 `_test.go` files, which I believe the linter is ignoring and it doesn't have any config option to include them afaict)


dashpole (approved):

aabmass:
@damemi can you push a new empty commit? I think that'll fix it.

PR #439, opened 2022-06-22
Add lint check for duplicated imports

damemi:
Ref https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/416

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -27,7 +27,6 @@ import (
 	"cloud.google.com/go/logging"
 	loggingv2 "cloud.google.com/go/logging/apiv2"
 	"google.golang.org/genproto/googleapis/api/monitoredres"
-	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	logpb "google.golang.org/genproto/googleapis/logging/v2"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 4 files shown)


exporter/collector/logs.go
```diff
@@ -27,7 +27,6 @@ import (
 	"cloud.google.com/go/logging"
 	loggingv2 "cloud.google.com/go/logging/apiv2"
 	"google.golang.org/genproto/googleapis/api/monitoredres"
-	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
```
damemi, line 30: there are more I need to remove, but I want to make sure CI fails without those updates

golangci.yml
```diff
@@ -106,6 +106,9 @@ linters-settings:
     - metres
     - kilometre
     - kilometres
+  revive:
```
damemi, line 109: @dashpole we have revive [enabled](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/439/files#diff-02f3aab86a2dfffc9cf61ee7bd633da2c63efcd7a7c5d668932b93d7612d9dfaR118) but by default it looks like it doesn't do anything (https://github.com/mgechev/revive#configuration):

> By default revive will enable only the linting rules that are named in the configuration file

So we should look at the list of available checks to see if there are any others we should enable

dashpole (approved):

PR #440, opened 2022-06-22
Parse logs traceID to correct format

damemi:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/11436

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -324,7 +324,7 @@ func (l logMapper) logToSplitEntries(
 
 	// parse TraceID and SpanID, if present
 	if !log.TraceID().IsEmpty() {
-		entry.Trace = log.TraceID().HexString()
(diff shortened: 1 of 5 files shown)


exporter/collector/logs.go
```diff
@@ -324,7 +324,7 @@ func (l logMapper) logToSplitEntries(
 
 	// parse TraceID and SpanID, if present
 	if !log.TraceID().IsEmpty() {
-		entry.Trace = log.TraceID().HexString()
+		entry.Trace = fmt.Sprintf("projects/%s/traces/%s", l.cfg.ProjectID, log.TraceID().HexString())
```
dashpole, line 327: lets use the projectID variable instead of l.cfg.ProjectID to make it work with multi-project export

damemi: done

dashpole (approved):

Issue #442, opened 2022-06-23
Exporter does not recognize log levels or any extra attributes 


kushal-ti:
I'm using the open telemetry filelog receiver to collect logs (since using the otlp log receiver is failing in the current release due to the issue mentioned here https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/435.) 
The output of the filelog receiver logs, after being passed through the processor in the otel collector, looks like this
```JSON
{
    "resourceLogs": [
        {
            "resource": {},
            "scopeLogs": [
                {
                    "scope": {},
                    "logRecords": [
                        {
                            "observedTimeUnixNano": "1655957310265130687",
                            "severityText": "INFO",
                            "body": {
                                "stringValue": "Entering healthCheckPost with list of 2 arguments"
                            },
                            "attributes": [
                                {
                                    "key": "gcp.log_name",
                                    "value": {
                                        "stringValue": "otel-collector-pipeline"
                                    }
                                },
                                {
                                    "key": "gcp.arguments",
                                    "value": {
                                        "stringValue": "[{\"type\": \"String\", \"value\": \"kushal\"}, {\"type\": \"CustomBody\", \"value\": {\"member1\":10,\"member2\":100,\"member3\":[10,100]}}]"
                                    }
                                } ... (truncated for brevity)
                            ],
                            "traceId": "e8f84199ef8f4e1779cd21577aaacbc1",
                            "spanId": "ee43d418bc51ca1f"
                        }
                    ]
                }
            ]
        }
    ]
}
``` 
However, this same log on the console looks like so 
![Screenshot from 2022-06-23 09-56-11](https://user-images.githubusercontent.com/88190553/175208976-ac6b1741-2d49-49e8-ab56-f4f5ad39afca.png)

According to the mapping in the logs data model mentioned here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#google-cloud-logging any extra attributes should be mentioned as gcp.* , but gcp.arguments attributes in the above log were dropped. https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/424 talks about non gcp.* attributes so i was under the implication that gcp.* attributes should have been retained, in the current release itself.

Additionally, it also does not retain the log severity, (INFO in this case), instead, the log severity I see on gcp console is default
[...]

PR #446, opened 2022-06-29
Parse basic logging severity text values

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/442

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -78,6 +78,39 @@ var severityMapping = []logging.Severity{
 	logging.Emergency, // 24 -> Emergency
 }
 
+// otelSeverityTexts maps the generic aliases of SeverityTexts to SeverityNumbers.
+// This can be useful if SeverityText is manually set to one of the values from the data
+// model in a way that doesn't automatically parse the SeverityNumber as well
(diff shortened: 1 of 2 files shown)


exporter/collector/logs.go
```diff
@@ -349,7 +382,14 @@ func (l logMapper) logToSplitEntries(
 	if log.SeverityNumber() < 0 || int(log.SeverityNumber()) > len(severityMapping)-1 {
 		return []logging.Entry{entry}, fmt.Errorf("Unknown SeverityNumber %v", log.SeverityNumber())
 	}
-	entry.Severity = severityMapping[log.SeverityNumber()]
+	logSeverity := severityMapping[log.SeverityNumber()]
```
punya, line 385: The nested boolean logic is a bit hard for me to follow. Would you consider instead:

```go
func cloudLoggingSeverity(log plog.LogRecord) logging.Severity {
  severityNumber := log.SeverityNumber()
  if severityNumber == 0 {
    severityNumber := otelSeverityTexts[strings.ToLower(log.SeverityText())]
  }
  return severityMapping[severityNumber]
}
```

The double negative in the comment is also difficult, could you rephrase it? e.g., document how severity mapping works overall, rather than this particular sub-case.

PR #452, opened 2022-07-07
Remove GMP factory functions and readme

damemi:
With the GMP exporter available in the contrib repo and distribution, the config and factory code is no longer needed here. Users can now build/use directly from collector-contrib, so the README (which only had instructions on how to build it yourself) is also removed. Other files are kept as they contain the functions used by the exporter


exporter/collector/integrationtest/testdata/fixtures/basic_counter_metrics_expect.json
```diff
@@ -600,74 +600,6 @@
       }
     ],
     "createMetricDescriptorRequests": [
-      {
-        "name": "projects/myproject",
```
damemi, line 604: cc @dashpole (when you're back) these fixture updates look weird to me, not sure if this is expected or I removed too much code from here. either way can you ptal and confirm?

dashpole: See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/342#discussion_r853158436.  The fixture changes here are the reverse of the fixture changes in the PR that added the GMP exporter, and exist because we no longer import exporterhelper.

Issue #454, opened 2022-07-12
Request: Expose gRPC Pool Size for configuration in NewGoogleCloudLogsExporter


cpheps:
We've been benchmarking the OTel `googlecloudexporter` paired with the `filelogreceiver`, details of setup are listed below. 

**Benchmarking Setup:**
- Generated Logs at 200K logs/s over 5 mintues (log lines were identical and 1KB in size)
- Wrote logs to a single file
- Used following Otel Config:
```yaml
receivers:
  filelog:
    include: ["/tmp/*.log"]
exporters:
  googlecloud:
    retry_on_failure:
      enabled: true
    log:
      default_log_name: perf-benchmark-test-otel
    sending_queue:
      num_consumers: 40
processors:
  memory_limiter:
    check_interval: 1s
    limit_mib: 10000
    spike_limit_mib: 2000
  batch:
    send_batch_max_size: 6000
    send_batch_size: 6000
service:
  pipelines:
    logs:
      receivers: [filelog]
      processors: [memory_limiter, batch]
      exporters: [googlecloud]
```

**Initial Results**:
With the above setup we observed the max log throughput to GCP to be ~120K logs/sec. We measured this by taking the rate of the `log_entry_count` metric over 1 minute intervals.  Here's a graph of the results

<img width="1706" alt="googlecloud_config_only" src="https://user-images.githubusercontent.com/11877763/178570873-a58b47f3-9dc9-46bc-a181-adccefa4662d.png">

**Investigation**:

We noticed that when creating the loggingClient [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L142) that there is a gRPC connection pool in the client. We hooked up a debugger and observed that there was only ever one connection in the pool. We hard coded in a connection pool with 20 connections and saw a large performance increase from a max throughput of ~120K logs/sec to ~150K logs/sec.

Here is the hardcoded change we tested:

```go
...
clientOpts, err := generateClientOptions(ctx, &cfg.LogConfig.ClientConfig, cfg.UserAgent, cfg.ImpersonateConfig)
if err != nil {
	return nil, err
}

// Hardcode connection pool
clientOpts = append(clientOpts, option.WithGRPCConnectionPool(20))

loggingClient, err := loggingv2.NewClient(ctx, clientOpts...)
if err != nil {
	return nil, err
}
...
```

Here is a graph of the same benchmark but with the connection pool increased:

[...]

PR #455, opened 2022-07-13
Add client option for GRPC connection pool size

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/454

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -51,6 +51,8 @@ type ClientConfig struct {
 	Endpoint string `mapstructure:"endpoint"`
 	// Only has effect if Endpoint is not ""
 	UseInsecure bool `mapstructure:"use_insecure"`
+	// GRPCPoolSize sets the size of the connection pool in the GCP client
+	GRPCPoolSize int `mapstructure:"grpcPoolSize"`
 }
 
 type TraceConfig struct {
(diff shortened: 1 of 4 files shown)


exporter/collector/config.go
```diff
@@ -51,6 +51,8 @@ type ClientConfig struct {
 	Endpoint string `mapstructure:"endpoint"`
 	// Only has effect if Endpoint is not ""
 	UseInsecure bool `mapstructure:"use_insecure"`
+	// GRPCPoolSize sets the size of the connection pool in the GCP client
+	GRPCPoolSize int `mapstructure:"grpcPoolSize"`
```
dashpole, line 55:
```suggestion
	GRPCPoolSize int `mapstructure:"grpc_pool_size"`
```

We use underscores in config

dashpole (commented):
can you also add an integration test in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtest/config/config_test.go for config parsing?

exporter/collector/config.go
damemi: updated

dashpole (approved):
lgtm after tests pass

PR #457, opened 2022-07-21
Add traces integration tests

damemi:
This adds a basic expect/fixture test and integration tests for traces

What's interesting is that we thought this would need some unique traceId generation for the integration test to work (ie, GCP to not reject the duplicate trace id from earlier runs). But running this locally, it seems to work fine.


dashpole (approved):
looks great!

PR #464, opened 2022-07-29
Prepare for v0.32.4 and v1.8.4

dashpole:


bogdandrutu (approved):

punya (approved):

dashpole:
@damemi note that the version update now updates the expected fixture.  Is there a way to not have that happen?

damemi:
> @damemi note that the version update now updates the expected fixture. Is there a way to not have that happen?

Yeah we can probably fix that. What commands did you run here? Is this just the result of `make fixtures` reading the new versions and updating based on that?

dashpole:
After updating the version in `release.go`, and running `make prepare-release`, the fixure-based unit test failed.  Then, I ran `make fixtures`, and committed the results.

PR #465, opened 2022-07-29
detectors/gcp: support Cloud Run jobs

zchee:
Support Cloud Run jobs using `CLOUD_RUN_JOB` and `CLOUD_RUN_EXECUTION`.

### DIFF
--- a/detectors/gcp/detector.go
+++ b/detectors/gcp/detector.go
@@ -36,7 +36,6 @@ const (
 	GKE
 	GCE
 	CloudRun
-	CloudRunJobs
 	CloudFunctions
 	AppEngineStandard
 	AppEngineFlex
@@ -49,8 +48,6 @@ func (d *Detector) CloudPlatform() Platform {
 		return GKE
 	case d.onCloudRun():
 		return CloudRun
-	case d.onCloudRunJobs():
-		return CloudRunJobs
 	case d.onCloudFunctions():
 		return CloudFunctions
 	case d.onAppEngineStandard():
--- a/detectors/gcp/detector.go
+++ b/detectors/gcp/detector.go
@@ -36,6 +36,7 @@ const (
 	GKE
 	GCE
 	CloudRun
+	CloudRunJobs
 	CloudFunctions
 	AppEngineStandard
 	AppEngineFlex
@@ -46,10 +47,12 @@ func (d *Detector) CloudPlatform() Platform {
 	switch {
 	case d.onGKE():
 		return GKE
-	case d.onCloudFunctions():
-		return CloudFunctions
 	case d.onCloudRun():
 		return CloudRun
+	case d.onCloudRunJobs():
+		return CloudRunJobs
+	case d.onCloudFunctions():
+		return CloudFunctions
 	case d.onAppEngineStandard():
 		return AppEngineStandard
 	case d.onAppEngine():
--- a/detectors/gcp/detector.go
+++ b/detectors/gcp/detector.go
@@ -36,6 +36,7 @@ const (
 	GKE
 	GCE
 	CloudRun
+	CloudRunJob
 	CloudFunctions
 	AppEngineStandard
 	AppEngineFlex
@@ -50,6 +51,8 @@ func (d *Detector) CloudPlatform() Platform {
 		return CloudFunctions
 	case d.onCloudRun():
 		return CloudRun
+	case d.onCloudRunJob():
+		return CloudRunJob
 	case d.onAppEngineStandard():
 		return AppEngineStandard
 	case d.onAppEngine():
--- a/detectors/gcp/detector.go
+++ b/detectors/gcp/detector.go
@@ -46,10 +46,10 @@ func (d *Detector) CloudPlatform() Platform {
 	switch {
 	case d.onGKE():
 		return GKE
-	case d.onCloudRun():
-		return CloudRun
 	case d.onCloudFunctions():
 		return CloudFunctions
+	case d.onCloudRun():
+		return CloudRun
 	case d.onAppEngineStandard():
 		return AppEngineStandard
 	case d.onAppEngine():
--- a/detectors/gcp/faas.go
+++ b/detectors/gcp/faas.go
@@ -40,10 +40,11 @@ func (d *Detector) onCloudFunctions() bool {
 }
 
 func (d *Detector) onCloudRun() bool {
-	if _, found := d.os.LookupEnv(cloudRunConfigurationEnv); found {
-		return found
-	}
+	_, found := d.os.LookupEnv(cloudRunConfigurationEnv)
+	return found
+}
 
+func (d *Detector) onCloudRunJob() bool {
 	_, found := d.os.LookupEnv(cloudRunJobsEnv)
 	return found
 }
--- a/detectors/gcp/faas.go
+++ b/detectors/gcp/faas.go
@@ -21,21 +21,21 @@ import (
 
 const (
 	// See the Cloud Run env vars:
-	// https://cloud.google.com/run/docs/reference/container-contract
+	// https://cloud.google.com/run/docs/container-contract#services-env-vars
 	// and the Cloud Functions env vars:
-	// https://cloud.google.com/functions/docs/configuring/env-var#runtime_environment_variables_set_automatically
-	cloudRunConfigEnv      = "K_CONFIGURATION"
-	cloudRunJobsEnv        = "CLOUD_RUN_JOB"
-	cloudFunctionTargetEnv = "FUNCTION_TARGET"
-	faasServiceEnv         = "K_SERVICE"
-	faasRevisionEnv        = "K_REVISION"
-	jobsRevisionEnv        = "CLOUD_RUN_EXECUTION"
-	jobsTaskIndexEnv       = "CLOUD_RUN_TASK_INDEX"
-	regionMetadataAttr     = "instance/region"
+	// https://cloud.google.com/functions/docs/configuring/env-var#python_37_and_go_111
+	cloudFunctionsTargetEnv  = "FUNCTION_TARGET"
+	cloudRunConfigurationEnv = "K_CONFIGURATION"
+	cloudRunJobsEnv          = "CLOUD_RUN_JOB"
+	faasServiceEnv           = "K_SERVICE"
+	faasRevisionEnv          = "K_REVISION"
+	jobsRevisionEnv          = "CLOUD_RUN_EXECUTION"
+	jobsTaskIndexEnv         = "CLOUD_RUN_TASK_INDEX"
+	regionMetadataAttr       = "instance/region"
 )
 
 func (d *Detector) onCloudRun() bool {
-	if _, found := d.os.LookupEnv(cloudRunConfigEnv); found {
+	if _, found := d.os.LookupEnv(cloudRunConfigurationEnv); found {
 		return found
 	}
 
@@ -44,7 +44,7 @@ func (d *Detector) onCloudRun() bool {
 }
 
 func (d *Detector) onCloudFunctions() bool {
-	_, found := d.os.LookupEnv(cloudFunctionTargetEnv)
+	_, found := d.os.LookupEnv(cloudFunctionsTargetEnv)
 	return found
 }
 
--- a/detectors/gcp/faas.go
+++ b/detectors/gcp/faas.go
@@ -34,17 +34,17 @@ const (
 	regionMetadataAttr       = "instance/region"
 )
 
+func (d *Detector) onCloudFunctions() bool {
+	_, found := d.os.LookupEnv(cloudFunctionsTargetEnv)
+	return found
+}
+
 func (d *Detector) onCloudRun() bool {
 	if _, found := d.os.LookupEnv(cloudRunConfigurationEnv); found {
 		return found
 	}
 
 	_, found := d.os.LookupEnv(cloudRunJobsEnv)
-	return found
-}
-
-func (d *Detector) onCloudFunctions() bool {
-	_, found := d.os.LookupEnv(cloudFunctionsTargetEnv)
 	return found
 }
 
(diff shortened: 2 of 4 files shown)


damemi (commented):
/gcbrun

zchee:
Note that Cloud Run jobs do not add `K_XXX` environment variables automatically. Instead of `CLOUD_RUN_JOB` and `CLOUD_RUN_EXECUTION`.

zchee:
@damemi not running gcbrun...?

damemi:
hm that seemed to work

zchee:
thanks

damemi:
cc @dashpole

dashpole:
I don't see those set in https://cloud.google.com/run/docs/container-contract#jobs-env-vars.  Are the docs incorrect?

zchee:
@dashpole I think doc is lacking. I was tested on real environment.

detectors/gcp/detector.go
```diff
@@ -36,6 +36,7 @@ const (
 	GKE
 	GCE
 	CloudRun
+	CloudRunJobs
```
dashpole, line 39: I don't think we need a new cloud run jobs constant here, since the detector will use the same functions (e.g. FaasName) to add resource attributes.

dashpole: Can you instead update the onCloudRun to return true on cloudRunJobs as well?

zchee:
@dashpole PTAL

zchee:
@dashpole gentle ping

dashpole:
Yes, sorry.  We are discussing as a team to make sure this is the right approach.

zchee:
@dashpole Ah, I see. I waiting your discuss result.

dashpole:
Hey @zchee, if you make a job with multiple tasks, what is the value in CLOUD_RUN_EXECUTION?  Does each task end up with the same job name and execution?  If so, we may need some additional information to prevent collisions between executions.

zchee:
@dashpole I'll investigate it.

Just in case, in the case of the `gcloud bate run jobs create` command, it means `--tasks=` flag, not `--parallelism` flag, right?

dashpole:
If you can, try both, and also look at what CLOUD_RUN_TASK_INDEX and CLOUD_RUN_TASK_ATTEMPT are set to.  Really appreciate the help

zchee:
@dashpole okay, will check both situations today.

Also, I'll report Cloud Run jobs docs lacking that's 2 envs via Mercari's Google TAM (I work for Mercari, Inc).

zchee:
@dashpole replied from Google support, said "seems to serverless team forgot update document", So I think would be better to waiting to update. Maybe your concern is also cleared.

zchee:
@dashpole I found otel side hardcoded `K_SERVICE`, need fix it also.

- https://github.com/open-telemetry/opentelemetry-go-contrib/blob/v1.9.0/detectors/gcp/cloud-run.go#L110

dashpole:
Note that the old cloud run (which [that code](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/v1.9.0/detectors/gcp/cloud-run.go#L110) references) is deprecated in favor of this code base.

zchee:
@dashpole updated document

zchee:
@dashpole gentle ping

I want to hear your opinions.

dashpole:
What were CLOUD_RUN_TASK_INDEX and CLOUD_RUN_TASK_ATTEMPT set to?

Also, what is in d.metadata.InstanceID() (faas.id)?

Basically, we want to make sure that faas.name, faas.version, and faas.id uniquely identify an instance of a cloud run job.  Otherwise, if two things are running in parallel with the same faas.* attributes, there could be conflicts.

zchee:
@dashpole I'll reply details later.

zchee:
@dashpole 

Use https://github.com/GoogleCloudPlatform/golang-samples/tree/main/run/jobs

```diff
diff --git a/run/jobs/main.go b/run/jobs/main.go
index 05e3195..a8a6d0a 100644
--- a/run/jobs/main.go
+++ b/run/jobs/main.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"strconv"
 	"time"
+
+	"cloud.google.com/go/compute/metadata"
 )
 
 type Config struct {
@@ -99,6 +101,12 @@ func main() {
 	}
 
 	log.Printf("Completed Task #%s, Attempt #%s", config.taskNum, config.attemptNum)
+
+	instanceID, err := metadata.InstanceID()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("instanceID: %s\n", instanceID)
 }
 
 // Throw an error based on fail rate
```

```console
$ gcloud alpha run jobs create --tasks=5 --parallelism=3 --region=asia-northeast1 jobs
# execute...
```

```
labels: {
run.googleapis.com/execution_name: "jobs-ph5p9"
run.googleapis.com/task_attempt: "0"
run.googleapis.com/task_index: "0"
}
```

```
labels: {
run.googleapis.com/execution_name: "jobs-ph5p9"
run.googleapis.com/task_attempt: "0"
run.googleapis.com/task_index: "1"
}
```

```
labels: {
run.googleapis.com/execution_name: "jobs-ph5p9"
run.googleapis.com/task_attempt: "0"
run.googleapis.com/task_index: "2"
}
```

```
textPayload: "2022/09/10 08:28:37 instanceID: 00c527f6d4ab9f193b199268ccb9d99d05ebc33d8983a56d2faef9169e9b42adf8fb6caaf62a35d1bff6df6b375c4282310b634da8218f77e706845ac6f4d5f27714"
```

dashpole:
So, it looks like we have:

Just job name
`CLOUD_RUN_JOB=my-job-name`

job name with execution suffix
`CLOUD_RUN_EXECUTION=my-job-name-ph5p9`

task index
`CLOUD_RUN_TASK_INDEX=0`

job UID
`instanceID=00c527f6d4ab9f193b199268ccb9d99d05ebc33d8983a56d2faef9169e9b42adf8fb6caaf62a35d1bff6df6b375c4282310b634da8218f77e706845ac6f4d5f27714`

Edit: just realized that they updated the docs to show what each environment variable has: https://cloud.google.com/run/docs/container-contract#jobs-env-vars.

dashpole:
My current thought is that we don't really have anything that could be construed as `faas.version`, right?  Seems like we only have `faas.id`, and `faas.name`, but no `faas.version`...

dashpole:
Looking at this some more, it does look like the execution is basically the version, in that it change each time a new version is rolled out.

It would be really nice to include the CLOUD_RUN_TASK_INDEX somewhere, as then you could use execution + index to uniquely identify a job

zchee:
@dashpole okay, fix asap

zchee:
@dashpole use `fmt.Sprintf("%s/%s", version, index)`, PTAL.

dashpole:
I discussed this with the team, and unfortunately I think we will need to update the OTel spec for cloud run jobs before we can merge this.  https://github.com/open-telemetry/opentelemetry-specification/blob/94c9c75c4f82fbda08bddff02ce9a0c582fdd55c/specification/resource/semantic_conventions/faas.md

zchee:
@dashpole 
> I discussed this with the team, and unfortunately I think we will need to update the OTel spec for cloud run jobs before we can merge this

I see.
Who is contribute this change to otel?

zchee:
@dashpole gentle ping

damemi:
Hi @zchee, sorry this got lost. @dashpole is on leave at the moment so I'll pick this back up.

Catching up, I don't think there were any issues with this besides adding the convention for Cloud Run Job versions to the specification. I've opened a PR there to do this: https://github.com/open-telemetry/opentelemetry-specification/pull/3378

Once we merge that, I think we can move forward with this again. Thanks for your patience

zchee:
@damemi Oh wow, that's great news :D And thanks for creating PR to OTel side.
I'll waiting for merge above with resolve conflicts.

zchee:
@damemi fixed lint issue, PTAL.

damemi:
/gcbrun
thanks @zchee! just going to wait for https://github.com/open-telemetry/opentelemetry-specification/pull/3378, but I'm tracking this now so it won't get lost again :)

detectors/gcp/detector.go
```diff
@@ -46,10 +46,10 @@ func (d *Detector) CloudPlatform() Platform {
 	switch {
 	case d.onGKE():
 		return GKE
-	case d.onCloudFunctions():
-		return CloudFunctions
 	case d.onCloudRun():
 		return CloudRun
+	case d.onCloudFunctions():
```
zchee, line 51: just sorted by enum order.

zchee:
@damemi Failed require `ops-go-e2e-cloud-functions` integration test. just flaky, or should I fix something?

detectors/gcp/detector.go
```diff
@@ -46,10 +46,10 @@ func (d *Detector) CloudPlatform() Platform {
...
+		return CloudFunctions
```
damemi, line 52: Ah I think this needs to have cloud functions first, see https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/552:

> The detector was incorrectly detecting the cloud run platform instead of the cloud functions platform. This is because cloud functions v2 is implemented on top of cloud run. This commit makes the detector check for cloud functions first to fix this issue.

We should at least leave a comment here about why this ordering is important. It would be better if we can find a way to more soundly fix this rather than relying on the order, but I'm not sure that will be possible if cloud functions is built on cloud run.

damemi (changes requested):

detectors/gcp/detector.go
zchee: I remembered. will fix

damemi:
@zchee sorry for the delay in this. The spec changes just merged and from the discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/3378 this will be slightly different than we talked about here.

Instead, we will have separate GCP-specific resource attributes for Cloud Run execution and Job ID (https://github.com/open-telemetry/opentelemetry-go/pull/4079 will add these to the Go semconv package).

I think we should probably just make these available via their own functions, and maybe add 2 new functions `IsJob()` and `IsService()`. wdyt?

zchee:
make sense. I'll implements.

detectors/gcp/detector.go
zchee: Cloud Functions gen2 also reserved the `K_CONFIGURATION` environment variable, but undocumented... :(
Well, therefore wrong detection if change the `switch...case` order.

As you said, we should leave a details comment here.

damemi (changes requested):
Thanks for updating this @zchee. We just need to make a few changes to match [the spec for Cloud Run Jobs](https://github.com/open-telemetry/semantic-conventions/blob/8fc3d034c7bdf4c975fa64a436ea92ef5192cccc/specification/resource/semantic_conventions/cloud_provider/gcp/cloud_run.md) that was approved upstream.

They did not like the idea of using `execution+task` as `faas.version`, so those will be their own individual attributes instead. That means we need separate functions for each of them.

To know that we need to call those functions, Cloud Run Jobs have to be treated like their own environment, separate from Services. So a new detector function for that as well.

imo, `faas.name` is fine to be shared between Jobs and Services. These are semantically providing the same info, even though it is technically different environment variables.

To summarize, we need the following new functions:
* `onCloudRunJob()` - [detect](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/9c6bfe4427dee294b17f8311c521f24072c26bf3/detectors/gcp/detector.go#L51) if we are running on Cloud Run Jobs
* `FaasJobExecution()` - return the Execution ID of the Job
* `FaasJobTaskIndex()` - return the Task Index for the execution

Our [Collector resource detection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0f8e1282b5cc7b8c1639d3117db36a5d31a3cccd/processor/resourcedetectionprocessor/internal/gcp/gcp.go#L61) will be updated to write to the new `gcp.cloud_run.*` attributes when running on Cloud Run Jobs, and our exporter's [monitored resource mapping](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/9c6bfe4427dee294b17f8311c521f24072c26bf3/internal/resourcemapping/resourcemapping.go#L135) will be updated to parse `execution+task` into a unique task ID.

zchee:
> We should at least leave a comment here about ...

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/465#discussion_r1225734878

Should I include that to this PR?

damemi:
> > We should at least leave a comment here about ...
> 
> [#465 (comment)](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/465#discussion_r1225734878)
> 
> Should I include that to this PR?

yes please do, this will prevent the same headache in the future :)

for the naming inconsistencies, I don't really have a good answer. I think the usage depends on the context, ie when referring to the product "Cloud Run jobs" makes sense, when referring to multiple jobs the plural makes sense vs a singular job. Either way, I would always prefer the `cloudRunJob` camel case vs `cloudRunjob` in code

detectors/gcp/faas.go
```diff
@@ -19,50 +19,82 @@ import (
...
 func (d *Detector) FaaSVersion() (string, error) {
 	if version, found := d.os.LookupEnv(faasRevisionEnv); found {
 		return version, nil
 	}
 	return "", errEnvVarNotFound
 }
 
-// FaaSID returns the instance id of the cloud run instance or cloud function.
+// FaaSJobExecution returns the execution id of the Cloud Run jobs.
+func (d *Detector) FaaSJobExecution() (string, error) {
```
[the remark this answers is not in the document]
damemi: Something shorter like `JobExecution()` and `JobTaskIndex()` are probably fine too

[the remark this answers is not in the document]
zchee: I choose the first one.

zchee:
@damemi PTAL.

damemi (approved):
thanks @zchee for working on this!

PR #466, opened 2022-08-04
Remove deprecated Flags API

TylerHelmuth:
Upgrades the collector libraries to 0.57.2 and removes the deprecated Flags API in favor of FlagsStruct.  Let me know if the package upgrade needs to be done in a separate PR.


dashpole:
/gcbrun
Thanks!

dashpole (approved):

dashpole:
looks like it needs a `make gotidy`?

TylerHelmuth:
> looks like it needs a make gotidy?

@dashpole done

Issue #421, opened 2022-06-13
Fix expected diff check in integration tests


damemi:
The integration tests that check expected output against a local fake server can fail based on the order of the output from the server. See runs: [fail](https://app.circleci.com/pipelines/github/GoogleCloudPlatform/opentelemetry-operations-go/1387/workflows/a7049ae8-6a76-4a78-9baa-a842d951eaee/jobs/1382), [pass](https://app.circleci.com/pipelines/github/GoogleCloudPlatform/opentelemetry-operations-go/1387/workflows/80e141fe-e8f8-4d46-982a-0bb6bd018737/jobs/1384) (from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/418)

output:
```
   metrics_test.go:73: 
        	Error Trace:	metrics_test.go:73
        	Error:      	Expected requests fixture and actual GCM requests differ
        	Test:       	TestMetrics/Multi-project_metrics
        	Messages:   	  (*integrationtest.MetricExpectFixture)(Inverse(protocmp.Transform, protocmp.Message{
        	            	  	"@type":                             s"fixtures.MetricExpectFixture",
[...]

PR #468, opened 2022-08-16
Upgrade otel collector and go library

bogdandrutu:
Also removed usage of deprecated `ioutil` package

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


dashpole (approved):

dashpole:
Would you mind running `make fixtures` and committing the result?  It should fix the unit test failure.

dashpole:
:/. Seems like a recurrence of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/421.

Can you run `make fixtures` again? Sorry

bogdandrutu:
It is not that, since `make fixtures` run multiple times does not produce any diff.

damemi:
I think it could still be #421, the diff looks like we are sorting on the MetricDescriptorRequest names, but not the names of the MetricDescriptors within those requests. Working on a fix to test it. Unfortunately this is tough to reproduce locally because it's essentially random which sorted order you're going to get, so it could keep passing locally and showing no diff but fail in CI

damemi:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/469 is passing but @bogdandrutu it looks like your changes in [Sort all the metrics](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/468/commits/89bfa35a33fccf607bc3feee290a8e5fdcfaef7a) will also work. If that goes green we can just close my PR (but you might still have to run `make fixtures` again to pick up the new sorting)

bogdandrutu:
@damemi just merge that and I will rebase.

damemi:
@bogdandrutu sorry for the delay, merged now

PR #486, opened 2022-09-21
Include the status code we compute in the error message

dashpole:
If the error returned is not a GRPC error, the status gets set to UNKNOWN and is used in metrics.  But since the original error message wasn't from gRPC, it isn't possible to tell where the UNKNOWN errors are coming from.

This makes the error include the gRPC status even when it isn't a gRPC error to match our self-observability metrics.  That will make it easier to link metrics to the resulting log messages.

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -242,7 +242,7 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
 			}
 
 			var st string
-			s, _ := status.FromError(err)
+			s := status.Convert(err)
 			st = statusCodeToString(s)
 
 			succeededPoints := len(ts)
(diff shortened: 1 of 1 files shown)


punya (approved):

exporter/collector/metrics.go
```diff
@@ -260,7 +260,7 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
 				recordPointCountDataPoint(ctx, failedPoints, st)
 			}
 			if err != nil {
-				errs = append(errs, fmt.Errorf("failed to export time series to GCM: %v", err))
+				errs = append(errs, fmt.Errorf("failed to export time series to GCM: %+v", s))
```
punya, line 263: For a further refinement: should we include the original error as a cause of what we're generating?

```go
errs = append(errs, fmt.Errorf("failed to export time series to GCM (status %+v): %w, s, err)
```
```

Issue #490, opened 2022-09-30
Metrics example logs errors


dashpole:
When running the metrics example, I get these log messages:

```
2022/09/29 20:40:59 Most recent data: counter 117, observer 17.29; histogram 4.95
2022/09/29 20:41:09 Most recent data: counter 116, observer 16.04; histogram 3.7
2022/09/29 20:41:19 Most recent data: counter 111, observer 15.44; histogram 3.1
2022/09/29 20:41:29 Most recent data: counter 153, observer 15.34; histogram 3
2022/09/29 20:41:39 Most recent data: counter 129, observer 13.89; histogram 1.55
2022/09/29 20:41:49 Most recent data: counter 159, observer 17.09; histogram 4.75
2022/09/29 20:41:49 rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Request was missing field timeSeries[0].metric: The TimeSeries must contain a metric.; Request was missing field timeSeries[2].metric: The TimeSeries must contain a metric.; Request was missing field timeSeries[4].metric: The TimeSeries must contain a metric.
error details: name = Unknown  desc = total_point_count:6 success_point_count:3 errors:{status:{code:3} point_count:3}
error details: name = DebugInfo detail = One or more TimeSeries could not be written: Request was missing field timeSeries[0].metric: The TimeSeries must contain a metric.: timeSeries[0]; Request was missing field timeSeries[2].metric: The TimeSeries must contain a metric.: timeSeries[2]; Request was missing field timeSeries[4].metric: The TimeSeries must contain a metric.: timeSeries[4] stack =
```

This is likely a regression from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/487

PR #491, opened 2022-09-30
Fix TS must contain a metric errors, and cleanup error aggregation

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/490

The reason for the errors was that we were re-setting the capacity of the tss list for each point.  This resulted in empty TimeSeries structs being added to the list of timeseries, and produced errors on export.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -436,7 +436,6 @@ func (me *metricExporter) recordToTspb(m metricdata.Metrics, mr *monitoredrespb.
 	errs := []error{}
 	switch a := m.Data.(type) {
 	case metricdata.Gauge[int64]:
-		tss = make([]*monitoringpb.TimeSeries, len(a.DataPoints))
 		for _, point := range a.DataPoints {
 			ts, err := gaugeToTimeSeries[int64](point, m, mr)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


exporter/metric/metric.go
```diff
@@ -433,32 +427,29 @@ func (me *metricExporter) recordToMpb(metrics metricdata.Metrics, attributes att
 // ref. https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries
 func (me *metricExporter) recordToTspb(m metricdata.Metrics, mr *monitoredrespb.MonitoredResource, library instrumentation.Scope) ([]*monitoringpb.TimeSeries, error) {
 	var tss []*monitoringpb.TimeSeries
-	errs := []error{}
+	var aggErr error
 	switch a := m.Data.(type) {
 	case metricdata.Gauge[int64]:
-		tss = make([]*monitoringpb.TimeSeries, len(a.DataPoints))
```
dashpole, line 439: This was the problematic line.  We already initialized tss above.

damemi (approved):
lgtm
couldn't an alternative be `make([]error, 0, 0)`? I think it's a wash on the end result, just curious

dashpole:
The error handling changes aren't part of the bug-fix.  It probably makes more sense to put those in another PR.

dashpole:
Moved error handling changes to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/492

PR #492, opened 2022-09-30
Improve error handling

dashpole:
This has a number of improvements:

* If we fail to write metric descriptors, we now still try to send the timeseries.
* If we fail to write one metric descriptor, we now still try to send other metric descriptors
* If we fail to convert one metric point in a metric, we now still send the other valid points
* [Breaking] If we encounter errors during conversion, we no longer pass the error to a user-provided error handler.  Instead, we return the error to the periodic reader.  The reader will handle the error with otel.Handle, and users can [set the otel error handler](https://github.com/open-telemetry/opentelemetry-go/blob/main/handler.go#L90) the same way they used to configure our error handler.
* Use multierr.Append or multierr.Combine instead of []error to simplify code.  This is slightly more efficient, and handles `nil` errors by not appending, so we aren't required to check for `nil` anymore.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -17,7 +17,6 @@ package metric
 import (
 	"context"
 	"fmt"
-	"log"
 	"math"
 	"reflect"
 	"strings"
@@ -148,14 +147,10 @@ func newMetricExporter(o *options) (*metricExporter, error) {
 
 // Export exports OpenTelemetry Metrics to Google Cloud Monitoring.
 func (me *metricExporter) Export(ctx context.Context, rm metricdata.ResourceMetrics) error {
-	if err := me.exportMetricDescriptor(ctx, rm); err != nil {
-		return err
-	}
-
-	if err := me.exportTimeSeries(ctx, rm); err != nil {
-		return err
-	}
-	return nil
+	return multierr.Combine(
+		me.exportMetricDescriptor(ctx, rm),
+		me.exportTimeSeries(ctx, rm),
+	)
 }
 
 // exportMetricDescriptor create MetricDescriptor from the record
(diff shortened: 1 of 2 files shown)


exporter/metric/metric.go
```diff
@@ -148,14 +147,10 @@ func newMetricExporter(o *options) (*metricExporter, error) {
 
 // Export exports OpenTelemetry Metrics to Google Cloud Monitoring.
 func (me *metricExporter) Export(ctx context.Context, rm metricdata.ResourceMetrics) error {
-	if err := me.exportMetricDescriptor(ctx, rm); err != nil {
-		return err
```
dashpole, line 152: this return would've prevented sending TS if we failed to write descriptors.

exporter/metric/metric.go
```diff
@@ -182,14 +177,15 @@ func (me *metricExporter) exportMetricDescriptor(ctx context.Context, rm metricd
 	// goroutines to send CreateMetricDescriptorRequest asynchronously in the case
 	// the descriptor does not exist in global cache (me.mdCache).
 	// See details in #26.
+	var err error
 	for kmd, md := range mds {
-		err := me.createMetricDescriptorIfNeeded(ctx, md)
-		if err != nil {
-			return err
```
dashpole, line 188: this return would've prevented other descriptors from being written

exporter/metric/metric.go
```diff
@@ -217,34 +213,18 @@ func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md
...
 	mr := me.resourceToMonitoredResourcepb(rm.Resource)
-	var errs []error
+	var aggError error
 
 	for _, scope := range rm.ScopeMetrics {
 		for _, metrics := range scope.Metrics {
 			ts, err := me.recordToTspb(metrics, mr, scope.Scope)
-			if err != nil {
-				errs = append(errs, err)
-			} else {
```
dashpole, line 227: this else prevented appending timeseries if recordToTspb returned a partial error (an error, and some timeseries)

exporter/metric/metric.go
```diff
@@ -217,34 +213,18 @@ func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md
...
-	}
-
-	var aggError error
-	if len(errs) > 0 {
-		aggError = multierr.Combine(errs...)
-	}
-
-	if aggError != nil {
-		if me.o.onError != nil {
-			me.o.onError(aggError)
```
dashpole, line 240: this is removed in-favor of returning the error.

damemi (approved):

PR #494, opened 2022-10-03
Run collector metrics integration tests on the SDK metrics exporter

dashpole:
It does this by converting from pdata to the sdk metric exporter types.

Since it currently fails the tests (because of differences between the exporters), i've disabled all the tests.  I'll reenable them once-at-a-time as I fix the differences.


exporter/collector/integrationtest/metrics_test.go
```diff
@@ -111,3 +115,93 @@ func TestMetrics(t *testing.T) {
...
+			require.NoError(t, err)
+
+			for _, m := range testcases.ConvertResourceMetrics(metrics) {
+				err = testServerExporter.Export(ctx, m)
+				if !test.ExpectErr {
+					require.NoError(t, err, "Failed to export metrics to local test server")
+				} else {
+					require.Error(t, err, "Did not see expected error")
+				}
+			}
```
dashpole, line 153: This is the only section that differs from TestCollectorMetrics above.

exporter/collector/integrationtest/testcases/conversion.go
```diff
@@ -0,0 +1,204 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric/unit"
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/resource"
+)
+
+// ConvertResourceMetrics converts a (collector) pdata metrics to an SDK ResourceMetrics
+// This is useful for testing the SDK with the same data used to test the collector.
+func ConvertResourceMetrics(pdataMetrics pmetric.Metrics) []metricdata.ResourceMetrics {
```
damemi, line 29: the conversion all looks pretty 1:1, but are you worried at all about maintaining these conversions? The metrics sdk is only alpha, so it could still change right?

damemi (approved):

exporter/collector/integrationtest/testcases/conversion.go
dashpole: It is, but it is spec-compliant (and the spec is stable), and the underlying OTLP datamodel is also stable, so I don't think it is likely to change.  pdata APIs are also not stable, so its also a risk.  But i'm not too worried.

PR #495, opened 2022-10-04
[breaking] Change default metric prefix for SDK exporter to workload.googleapis.com

dashpole:
This aligns with the collector exporter.  It is one difference found by integration tests.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


dashpole:
Users can revert to the previous behavior using `WithMetricDescriptorTypeFormatter`, and adding a prefix `custom.googleapis.com/opentelemetry` in their formatter function.

Otherwise, this will change the name of metrics sent using this exporter

damemi (approved):

PR #498, opened 2022-10-04
Correctly sanitize metric label keys in the SDK metric exporter

dashpole:
This is an inconsistency discovered by the collector integration tests.  This updates the sanitization to use the same function.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"strings"
 	"time"
+	"unicode"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/sdk/instrumentation"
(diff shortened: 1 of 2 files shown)


exporter/metric/metric.go
```diff
@@ -627,5 +628,21 @@ func numberDataPointToValue[N int64 | float64](
 // > * Label name must start with a letter or digit.
 // > * The maximum length of a label name is 100 characters.
 func normalizeLabelKey(s string) string {
-	return strings.ReplaceAll(s, ".", "_")
+	if len(s) == 0 {
+		return s
+	}
+	s = strings.Map(sanitizeRune, s)
+	if unicode.IsDigit(rune(s[0])) {
```
damemi, line 635: the comments above say `Label name must start with a letter or digit`, is that not accurate?

damemi: there's also no check on the length, do we just truncate?

damemi (approved):

dashpole:
Some experiments:

Sending label key: `7bar`

```error details: name = DebugInfo detail = One or more TimeSeries could not be written: Field timeSeries[2].metric.labels had an invalid value of "7bar": Label key contains invalid characters.: timeSeries[2] stack =```


Sending label key: `key_7bar`

Success

So I think the comment was incorrect.

**Comment by dashpole** [comment]
Sending label with 101 character-long label:

```2022/10/05 18:48:49 rpc error: code = InvalidArgument desc = Field labels[0].key had an invalid value of "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff": When creating metric custom.googleapis.com/opentelemetry/observer-a:  the label key is more than 100 characters long.
error details: name = DebugInfo detail = Field labels[0].key had an invalid value of "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff": When creating metric custom.googleapis.com/opentelemetry/observer-a:  the label key is more than 100 characters long. stack =```

**Comment by dashpole** [comment]
In the collector exporter, I see the following description:

```golang
// Replaces non-alphanumeric characters to underscores. Note, this does not truncate label keys
// longer than 100 characters or prepend "key" when the first character is "_" like OpenCensus
// did.
```

Which seems to indicate that the decision not to truncate was intentional.  I'm going to leave it as-is for now, but update the comment.

PR #503, opened 2022-10-05
[WIP] Enable testing the SDK exporter using collector integration tests

dashpole:
This adds SDK exporter options to tests so we can configure project, etc.

It either enables a test, or documents the reason why it can't be enabled.


exporter/metric/metric.go
```diff
@@ -469,8 +469,7 @@ func gaugeToTimeSeries[N int64 | float64](point metricdata.DataPoint[N], metrics
 		ValueType:  valueType,
 		Points: []*monitoringpb.Point{{
 			Interval: &monitoringpb.TimeInterval{
-				StartTime: timestamp,
-				EndTime:   timestamp,
+				EndTime: timestamp,
```
dashpole, line 472: StartTime is not required for gauges.  This matches the collector exporter's behavior.

dashpole:
This is ready for review.

damemi (approved):

Issue #506, opened 2022-10-06
SDK Metric exporter should add service_instance_id and service_name labels


dashpole:
This will match the behavior of the collector.  There should be an option to disable these extra labels.

The reason for adding them in the first place is to prevent collisions if the monitored resource doesn't uniquely identify the entity writing the metric.

PR #508, opened 2022-10-10
Add service labels to GCM metrics exporter

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/506

### Design

This adds a new Option: `WithFilteredResourceAttributes(filter attribute.Filter)`, which allows users to specify which resource attributes should be added to the metric as metric labels.

It changes the default behavior of the exporter by adding service.name, service.namespace, and service.instance.id as metric labels by default.  This matches the default behavior of the collector exporter.

It also provides a helper function for disabling this functionality entirely: `NoAttributes()`, which can be used with: `WithFilteredResourceAttributes(NoAttributes())`.

### Alternative

Instead of `WithFilteredResourceAttributes(NoAttributes())`, we could just add `WithServiceLabelsDisabled()` (the first commit only).  This would make it simpler to disable the functionality, but with less configurability for users.  Since the collector exporter allows specifying prefixes of resource attributes to promote, allowing a filter seems useful.

### DIFF
--- a/exporter/metric/cloudmonitoring.go
+++ b/exporter/metric/cloudmonitoring.go
@@ -29,7 +29,10 @@ import (
 
 // New creates a new Exporter thats implements metric.Exporter.
 func New(opts ...Option) (metric.Exporter, error) {
-	o := options{context: context.Background()}
+	o := options{
+		context:                 context.Background(),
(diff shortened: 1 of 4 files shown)


damemi (approved):
agree with the abstract filter approach vs the alternative, but it was helpful to see both

PR #510, opened 2022-10-10
Deprecated unused constants

dashpole:
These were probably used for resource detection, but are mostly replaced by the OTel semconv package.

### DIFF
--- a/exporter/metric/constants.go
+++ b/exporter/metric/constants.go
@@ -23,45 +23,75 @@ package metric
 // set to any meaningful value within the environment. For example,
 // GKE clusters have a name which can be used for this label.
 const (
-	CloudKeyProvider  = "cloud.provider"
+	// Deprecated: use semconv.CloudProviderKey instead
+	CloudKeyProvider = "cloud.provider"
+	// Deprecated: use semconv.CloudAccountIDKey instead
 	CloudKeyAccountID = "cloud.account.id"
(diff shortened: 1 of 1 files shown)


exporter/metric/constants.go
```diff
@@ -23,45 +23,75 @@ package metric
...
+	K8SPod = "k8s_pod"
+	// Deprecated: Use "k8s_cluster" instead
+	K8SCluster = "k8s_cluster"
+	// Deprecated: Use "gce_instance" instead
+	GCEInstance = "gce_instance"
+	// Deprecated: Use "aws_ec2_instance" instead
 	AWSEC2Instance = "aws_ec2_instance"
-	GenericTask    = "generic_task"
+	// Deprecated: Use "generic_task" instead
+	GenericTask = "generic_task"
```
damemi, line 96: why are we deprecating these constants to tell users to use the strings instead? Can't we keep providing these constants?

damemi (approved):

exporter/metric/constants.go
dashpole: There really isn't a reason (I can think of) why someone would need to use these constants with the exporter, which is why I'd prefer not to keep them.  If someone does need that string for some reason, they can just define the constant themselves.

If we ever support custom monitored resource mappings (which might be a useful feature), we would re-introduce similar constants at that point.

Issue #525, opened 2022-11-01
Histogram type metrics with only one `+inf` bucket fails to export


ridwanmsharif:
For `prometheus` metrics, The `+Inf` bucket must always be present, and will match the value of the `_count`.
When using GMP and using OTel to send metrics there we encounter an interesting error. The following metric fails to get exported to GMP even though it is valid:
```
# HELP kv_prober_write_latency Latency of successful KV write probes
# TYPE kv_prober_write_latency histogram
kv_prober_write_latency_bucket{le="+Inf"} 0
kv_prober_write_latency_sum 0
kv_prober_write_latency_count 0
```

This is from a CockroachDB prometheus exporter. It fails with the following error:
 ```
 {
            "error": "failed to export time series to GCM: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Field timeSeries[26].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.; Field timeSeries[6].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.; Field timeSeries[10].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.; Field timeSeries[142].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.; Field timeSeries[58].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.; Field timeSeries[111].points[0].distributionValue had an invalid value: Distribution |explicit_buckets.bounds| does not have at least one entry.\nerror details: name = Unknown  desc = total_point_count:200  success_point_count:194  errors:{status:{code:3}  point_count:6}"
        },
 ```

I believe this situation can be hit in the following way: If none of the buckets have a non-zero value they are not reported by exporters but the `+inf` bucket always is in prometheus. That would lead to this error.
 
Looks like our prometheus engine discards these: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/83d012e29b74306dbee3e367270476ae86f4531a/pkg/export/transform.go#L292

Perhaps we should do something similar. Or adding in a `math.inf` bound in this case? Would that work?

dashpole:
We don't want to deviate from the behavior of prometheus-engine, so for now we should discard those metrics.

damemi:
[...]

PR #533, opened 2022-11-18
Drop histogram metrics with no non-zero bucket values

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/525 by dropping metric points that have no bucket bounds, consistent with GMP.

Testdata was generated with the following sample metric:
```
# HELP kv_prober_write_latency Latency of successful KV write probes
# TYPE kv_prober_write_latency histogram
kv_prober_write_latency_bucket{le="+Inf"} 0
kv_prober_write_latency_sum 0
kv_prober_write_latency_count 0
```
served on a local endpoint, scraped with the collector's prometheus receiver, and written with the file exporter.

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -657,8 +657,9 @@ func (m *metricMapper) histogramToTimeSeries(
 	point pmetric.HistogramDataPoint,
 	projectID string,
 ) []*monitoringpb.TimeSeries {
-	if point.Flags().NoRecordedValue() || !point.HasSum() {
+	if point.Flags().NoRecordedValue() || !point.HasSum() || point.ExplicitBounds().Len() == 0 {
 		// Drop points without a value or without a sum
+		m.obs.log.Debug("Metric has no value, sum, or explicit bounds. Dropping the metric.", zap.Any("metric", metric))
 		return nil
 	}
 	t, err := m.metricNameToType(metric.Name(), metric)
--- a/exporter/collector/integrationtest/testcases/testcases_metrics.go
+++ b/exporter/collector/integrationtest/testcases/testcases_metrics.go
@@ -205,5 +205,12 @@ var MetricsTestCases = []TestCase{
 		// Multi-project exporting is not supported in the SDK exporter
 		SkipForSDK: true,
 	},
+	{
+		// see https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/525
+		Name:                 "Metrics with only one +inf bucket",
+		OTLPInputFixturePath: "testdata/fixtures/metrics/prometheus_empty_buckets.json",
+		ExpectFixturePath:    "testdata/fixtures/metrics/prometheus_empty_buckets_expected.json",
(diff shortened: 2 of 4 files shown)


exporter/collector/metrics.go
```diff
@@ -657,7 +657,7 @@ func (m *metricMapper) histogramToTimeSeries(
 	point pmetric.HistogramDataPoint,
 	projectID string,
 ) []*monitoringpb.TimeSeries {
-	if point.Flags().NoRecordedValue() || !point.HasSum() {
+	if point.Flags().NoRecordedValue() || !point.HasSum() || point.ExplicitBounds().Len() == 0 {
```
damemi, line 660: this change affected the `_expected` test fixture:
```diff
diff --git a/exporter/collector/integrationtest/testdata/fixtures/metrics/prometheus_empty_buckets_expected.json b/exporter/collector/integrationtest/testdata/fixtures/metrics/prometheus_empty_buckets_expected.json
index 6633318..96c3f83 100644
--- a/exporter/collector/integrationtest/testdata/fixtures/metrics/prometheus_empty_buckets_expected.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/metrics/prometheus_empty_buckets_expected.json
@@ -3,44 +3,6 @@
     {
       "name": "projects/fakeprojectid",
       "timeSeries": [
-        {
-          "metric": {
-            "type": "workload.googleapis.com/kv_prober_write_latency",
-            "labels": {
-              "service_instance_id": "0.0.0.0:2112",
-              "service_name": "otel-collector"
-            }
-          },
-          "resource": {
-            "type": "generic_task",
-            "labels": {
-              "job": "otel-collector",
-              "location": "global",
-              "namespace": "",
-              "task_id": "0.0.0.0:2112"
-            }
-          },
-          "metricKind": "CUMULATIVE",
-          "valueType": "DISTRIBUTION",
-          "points": [
-            {
-              "interval": {
-                "endTime": "1970-01-01T00:00:00Z",
-                "startTime": "1970-01-01T00:00:00Z"
-              },
-              "value": {
-                "distributionValue": {
-                  "bucketOptions": {
-                    "explicitBuckets": {}
-                  },
-                  "bucketCounts": [
-                    "0"
-                  ]
-                }
-              }
-            }
-          ]
-        },
```

dashpole: Thats interesting.  How were integration tests passing before if there was an invalid histogram?

damemi: Sorry, this wasn't a test that existed before. I added it, and did a before and after with the logic change to see that the (new) fixture was updated.

dashpole, line 660: Can we add a debug-level log?  It would also be nice if it was possible to use self-obs metric to detect cases like these where we drop a metric, but we shouldn't block this bug fix for that.

damemi: sure, added

dashpole (approved):

Issue #447, opened 2022-06-30
Collector Exporter: Support regex matching for metric resource_filters


jsirianni:
It would be nice if metric.resource_filters supported regex matching in addition to string prefix. 

If I want to always copy all resource attributes, I need to define resource_filters like this:
```yaml
...
          metric:
            resource_filters:
              - prefix: a
              - prefix: A
              - prefix: b
              - prefix: B
              ...
              - prefix: z
              - prefix: Z
...
```

It would be nice if I could accomplish the same thing using the filter `*` or have an option to skip matching entirely and just copy all resource attributes to metric labels.

This is a big friction point for customers because they need to define a resource prefix for every receiver in their pipeline. Failing to do so will result in duplicate metric / time series errors. For example, Rabbitmq has metrics on a per queue basis. If The system has multiple queues, the Google API will reject these metrics (because they appear are duplicates) if the user has not configured the correct resource filter prefix.  

This configuration will not work until the user has added the correct resource filter prefix:
```yaml
receivers:
  rabbitmq:
    endpoint: http://localhost:15672
    collection_interval: 60s
    # From environment. See README.md.
    username: $RABBITMQ_USERNAME
    password: $RABBITMQ_PASSWORD

exporters: 
  googlecloud:

service:
  pipelines:
    metrics:
      receivers:
      - rabbitmq
      exporters:
      - googlecloud
  ```

These resource attributes are critical but omitted by default.

**Comment by dashpole** [comment]
Just curious, does this work?
```yaml
          metric:
            resource_filters:
              - prefix: ""
```

PR #535, opened 2022-11-21
Add regex matching to resource attribute filters

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/447 by adding a `Regex` field to resource filters. There are no restrictions on this field, so for example it is ok to set `Prefix` and `Regex`. The code will simply do an OR check if one of them matches.


exporter/collector/monitoredresource.go
```diff
@@ -71,7 +75,15 @@ func resourceToLabels(
 		}
 		// Matches one of the resource filters
 		for _, resourceFilter := range resourceFilters {
-			if strings.HasPrefix(k, resourceFilter.Prefix) {
+			matchesRegex := false
+			if len(resourceFilter.Regex) > 0 {
+				matched, err := regexp.Match(resourceFilter.Regex, []byte(k))
+				if err != nil {
+
```
dashpole: handle error here?

dashpole (commented):
I would personally expect AND behavior if I put two conditions together, and OR behavior if I put two conditions separately.

dashpole:
But then again, a regex is strictly more expressive than a prefix, so its probably fine to take either approach.  Or we could only allow one of [Prefix, regex] and leave our options open for the future

damemi:
I like it. To your point about expressiveness, I can't think of a case where someone would want to AND a prefix+regex that couldn't be expressed in one regex. But I could see the separation making configs a little more readable, if someone chooses to do that.

Made the change, please double check my logic makes sense.

exporter/collector/monitoredresource.go
```diff
@@ -71,7 +75,15 @@ func resourceToLabels(
 		}
 		// Matches one of the resource filters
 		for _, resourceFilter := range resourceFilters {
-			if strings.HasPrefix(k, resourceFilter.Prefix) {
+			matchesRegex := true
+			if len(resourceFilter.Regex) > 0 {
+				matched, err := regexp.Match(resourceFilter.Regex, []byte(k))
+				if err != nil {
+					log.Debug("Unable to parse resource filter regex.", zap.Error(err), zap.Any("regex", resourceFilter.Regex))
```
dashpole, line 82: should parse errors be found on startup (and be fatal), instead of logged each time we try to match something?

exporter/collector/monitoredresource.go
```diff
@@ -71,7 +75,15 @@ func resourceToLabels(
...
+				}
+				matchesRegex = matched
+			}
+			if strings.HasPrefix(k, resourceFilter.Prefix) && matchesRegex {
```
dashpole, line 86:
```suggestion
				if !matched {
				         continue
				}
			}
			if strings.HasPrefix(k, resourceFilter.Prefix) {
```
Seems simpler not to track matchesRegex separately

damemi: Taking out the error handling made it easier for me to just do both checks in the `if`. I had trouble reusing the `matched` variable from the regexp call because I was declaring a new `err` on the same line. It's better now

PR #537, opened 2022-11-28
Update linters to match upstream settings

### LINKED ISSUE #524: Update and review enabled linters
Upstream otel-go recently had to change the linters they enable: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2896

We should update golangci-lint, and probably make some of the same changes.

damemi:
Fixes #524 
This updates our linter settings to match [upstream](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/2aae3a65934017de3c5b65de95a7f1a2ddad84f8/.golangci.yml)

### DIFF
--- a/e2e-test-server/endtoendserver/constants.go
+++ b/e2e-test-server/endtoendserver/constants.go
@@ -32,16 +32,27 @@ const (
 )
 
 var (
-	subscriptionMode        = mustGetenv("SUBSCRIPTION_MODE")
-	projectID               = mustGetenv("PROJECT_ID")
-	requestSubscriptionName = mustGetenv("REQUEST_SUBSCRIPTION_NAME")
-	responseTopicName       = mustGetenv("RESPONSE_TOPIC_NAME")
+	subscriptionMode        string
+	projectID               string
+	requestSubscriptionName string
+	responseTopicName       string
 )
 
-func mustGetenv(key string) string {
-	v := os.Getenv(key)
-	if v == "" {
-		log.Fatalf("environment variable %q must be set", key)
+func init() {
+	subscriptionMode = os.Getenv("SUBSCRIPTION_MODE")
+	if subscriptionMode == "" {
+		log.Fatalf("environment variable SUBSCRIPTION_MODE must be set")
 	}
-	return v
+	projectID = os.Getenv("PROJECT_ID")
+	if projectID == "" {
+		log.Fatalf("environment variable PROJECT_ID must be set")
+	}
+	requestSubscriptionName = os.Getenv("REQUEST_SUBSCRIPTION_NAME")
+	if requestSubscriptionName == "" {
+		log.Fatalf("environment variable REQUEST_SUBSCRIPTION_NAME must be set")
+	}
+	responseTopicName = os.Getenv("RESPONSE_TOPIC_NAME")
+	if responseTopicName == "" {
+		log.Fatalf("environment variable RESPONSE_TOPIC_NAME must be set")
+	}
 }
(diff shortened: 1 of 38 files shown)


golangci.yml
```diff
@@ -107,19 +137,168 @@ linters-settings:
...
+        disabled: false
+      # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#time-equal
+      - name: time-equal
+        disabled: false
+      # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming
+      - name: var-naming
+        disabled: false
+        arguments:
+          - ["ID"] # AllowList
+          - ["Otel", "Aws", "Gcp"] # DenyList
```
damemi, line 256: this might not be relevant to us, but I want to see what it does in CI

e2e-test-server/endtoendserver/constants.go
```diff
@@ -32,16 +32,27 @@ const (
...
+	if projectID == "" {
+		log.Fatalf("environment variable PROJECT_ID must be set")
+	}
+	requestSubscriptionName = os.Getenv("REQUEST_SUBSCRIPTION_NAME")
+	if requestSubscriptionName == "" {
+		log.Fatalf("environment variable REQUEST_SUBSCRIPTION_NAME must be set")
+	}
+	responseTopicName = os.Getenv("RESPONSE_TOPIC_NAME")
+	if responseTopicName == "" {
+		log.Fatalf("environment variable RESPONSE_TOPIC_NAME must be set")
```
damemi, line 56: this doesn't look great, but I couldn't find another way around the deep-exit linter. It doesn't have any configuration that I can find, so I can't skip this package. I tried calling `mustGetenv` from `init()` hoping that would solve it, but that still failed.

dashpole: You can disable the linter per-line if you think it isn't worth it (e.g. for a test).  See this as an example: https://github.com/open-telemetry/opentelemetry-go/blob/aa868d506b7e4de2f534f47b168b27b7f8824351/sdk/metric/metricdata/temporality.go#L24

exporter/collector/metrics.go
```diff
@@ -119,8 +119,14 @@ func NewGoogleCloudMetricsExporter(
 	version string,
 	timeout time.Duration,
 ) (*MetricsExporter, error) {
-	view.Register(MetricViews()...)
-	view.Register(ocgrpc.DefaultClientViews...)
+	err := view.Register(MetricViews()...)
```
dashpole, line 122: Does this mean you can't use more than one googlecloud exporter in a collector?

exporter/collector/traces.go
```diff
@@ -41,7 +41,10 @@ func (te *TraceExporter) Shutdown(ctx context.Context) error {
 }
 
 func NewGoogleCloudTracesExporter(ctx context.Context, cfg Config, version string, timeout time.Duration) (*TraceExporter, error) {
-	view.Register(ocgrpc.DefaultClientViews...)
+	err := view.Register(ocgrpc.DefaultClientViews...)
```
dashpole, line 44: same comment as for monitoring.

dashpole (approved):

dashpole (changes requested):
sorry, didn't mean to approve

exporter/collector/metrics.go
damemi: does that cause `view.Register()` to throw an error? if so we can change this

dashpole: I believe that does.  You can test it by adding a second view.Register() call in this function :).

damemi: Per offline discussion, tried adding a double call and didn't see the expected failure. So, reverting this and skipping errcheck on these lines (ie, current behavior) until we can understand why this didn't fail when we expected it to.

dashpole (approved):

PR #542, opened 2022-12-02
Do not mutate incoming metrics

dmitryax:
The method Sort mutates the incoming data which can lead to subtle bugs like https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16499 given that googlecloudexporter marked as not mutating


damemi (commented):
/gcbrun

damemi (approved):

Issue #547, opened 2022-12-19
Automatic Credential Lookup in collector exporter make it not possible to set credentials


cpheps:
[This](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/config.go#L254-L266) block of code, add in #527, in `exporter/collector/config.go` does an automatic credentials lookup where it checks the existence of the `GOOGLE_APPLICATION_CREDENTIALS` ENV or the `application_default_credentials.json` file. If one is present it loads the credentials if not it errors. 

In our collector we have wrapped the `googlecloud` exporter to allow customers to specify a `credentials_file` or embed the credentials JSON in the config via a `credentials` parameter. Here's is our [config documentation](https://github.com/observIQ/observiq-otel-collector/tree/main/exporter/googlecloudexporter#configuration). We have customers that require this functionality for configuring more than one `googlecloud` exporter in an OTel collector. Primarily if they want to send data to multiple projects they have an exporter for each.

With the recent change to check for credentials we can no longer add our own credential config option as it is not taken into account in this logic and causes one of two behaviors:

1. If there are no default credentials it errors that none were found
2. If there are default credentials the gRPC client errors that multiple credentials were set (default one the ones we pass in).

It would be nice if this default lookup logic only occurred if no other credentials options were set.

cpheps:
@dashpole 

> Does this problem exist for any option? Or are credentials the only place where multiple options is not permitted?

I believe it's just credentials. The validation logic lives in [google.golang.org/api](https://github.com/googleapis/google-api-go-client/blob/main/internal/settings.go#L80). Looks like it only worries about multiple credentials being set.

> Option 3 seems the least invasive, but option 4 might be worth considering as well.

I agree with this assessment. For option 4, I think we brought up adding the ability to specify non-default credentials before and at the time it didn't seem like this was something Google wanted to support at the time. Maybe that has since changed.

dashpole:
Let's go with option 3 for now to fix the issue.

> We have customers that require this functionality for configuring more than one googlecloud exporter in an OTel collector. Primarily if they want to send data to multiple projects they have an exporter for each.

[...]

PR #548, opened 2022-12-19
Fix GetClientOptions conflicts with default credentials

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/547

Omit default credentials if additional options are provided, as these could conflict.


damemi (approved):

cpheps:
@dashpole Is it possible to update the googlecloudexporter [go.mod](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter/go.mod#L6) with a version of the collector module that has this fix?

I'm not sure if there any other changes since the last version that isn't ready to be upstreamed yet.

dashpole:
We can try to tag a release this week.

Issue #208, opened 2021-09-23
opentelemetry-operations-go/exporter/trace rpc error: code = Unavailable


eqinox76:
Hey all,

our monitoring sometimes shows the following error message from the traces exporter:

```
rpc error: code = Unavailable desc = The service is currently unavailable.

stack:
/home/go/pkg/mod/go.opentelemetry.io/otel@v1.0.0/handler.go:106                                   Handle
/home/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.0.0/trace/batch_span_processor.go:248            (*batchSpanProcessor).processQueue
/home/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.0.0/trace/batch_span_processor.go:111            NewBatchSpanProcessor.func1
/opt/go/1.16.3/x64/src/runtime/asm_amd64.s:1371                                                   goexit
```

We are using the following library versions:
```
	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.24.0
	go.opentelemetry.io/otel v1.0.0
	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.0.0
	go.opentelemetry.io/otel/sdk v1.0.0
	go.opentelemetry.io/otel/trace v1.0.0
	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v1.0.0-RC2
```

I think the error is common among grpc systems when something is restarting or connections are lost. Maybe providing a retry option like [grpc_retry](https://github.com/grpc-ecosystem/go-grpc-middleware) would help to reduce data loss and the occurrence of such low level alerts.

Regards,
Carsten

damemi:
It looks like the cloud trace client has [retry logic built in by default](https://github.com/googleapis/google-cloud-go/blob/trace/v1.2.0/trace/apiv2/trace_client.go#L62). I'm wondering if I followed that correctly, or if maybe something in the batch processor is dropping that retry handling?

dashpole:
Can we test it with our mock cloud trace implementation?

damemi:
Re-reading it, I think that retry logic is only on by default for `CreateSpans`, but our exporter uses `BatchWriteSpans`. Still need to test this, but adding a note for when I can get to it

damemi:
[...]

PR #550, opened 2022-12-21
Bump Cloud Trace library to v1.8.0

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/208

This bumps the cloud trace client to v1.8.0, which offers default batch retry on failed writes.

Currently the trace client [retries failed calls](https://github.com/googleapis/google-cloud-go/blob/trace/v1.2.0/trace/apiv2/trace_client.go#L59) for `CreateSpans` but not `BatchWriteSpans`.


dashpole:
Should we open an upstream issue asking for the default to be changed?  Or ask why the default is set the way it is?

If we can avoid adding a new config surface, that seems best.  It also would seem odd for us to have a different default from the upstream library.

damemi:
Upstream issue raised at https://github.com/googleapis/google-cloud-go/issues/7184

damemi:
update: we implemented this as a default setting in the upstream trace client instead. but, I think we can still use the unit tests I added here to make sure attempts are retried. once the upstream generated clients land, I'll use this PR to bump the dependency and add these tests. Holding until then

dashpole (approved):

Issue #411, opened 2022-06-06
Multi-project export uses quota from the credential's project.


dashpole:
If you have a service account from project A, and override to send telemetry to project B using `gcp.project.id`, it will use quota from project A.  This probably isn't what users expect.

There is an [option](https://github.com/googleapis/google-api-go-client/blob/2bdb4a9808454c2e43c015b99011a42d084aa02d/internal/settings.go#L53) when we Dial the grpc client called QuotaProject we could use to override the project used for quota.

We need to decide what the best way to handle this is.

aabmass:
@damemi taking a look

damemi:
I don't think this is possible without a change to the cloud trace/monitoring/logging libraries, or possibly grpc (https://github.com/googleapis/google-api-go-client/issues/1780).

* the grpc QuotaProject option won't work because it is a DialOption, and we want to set this dynamically for each call
* the `x-goog-user-project` header could work, but this is always set by the underlying grpc libraries to be the default credential's project. 
* we don't have a way to override this and it appears like hardcoded behavior in the grpc libraries.
* in the future, we might be able to use a chained interceptor to override this

I'm going to follow the issue I created in grpc to see if I'm missing something obvious. Otherwise, we'll close this as out of our current scope and we can reopen it if needed.

damemi:
Looking more into this it seems like running a collector without a default quota project defined in creds does work, so I opened https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/555

PR #555, opened 2023-01-05
Support multi-project quota usage

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/411

Need to figure out a way to add a test for this (only done manually so far). But ultimately it's just setting the header on requests.

Note that this will not work if the collector's credentials already define a quota project. We could probably automatically parse that out when this feature is enabled, but it feels hacky so for a first release of this I think documenting the requirement, allowing it to explicitly fail, and alerting the user is good enough.

The logs collector exporter will need to be refactored to use this, because we currently batch requests of logs from potentially multiple projects into a single request.

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -137,6 +137,7 @@ The following configuration options are supported:
     User should calculate this as `num_seconds * requests_per_second` where:
     - `num_seconds` is the number of seconds to buffer in case of a backend outage
     - `requests_per_second` is the average number of requests per seconds.
+- `destination_project_quota` (optional): Enables [multi-project quota usage](#multi-project-exporting) for traces and metrics.
 
 Note: These `retry_on_failure` and `sending_queue` are provided (and documented) by the [Exporter Helper](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#configuration)
 
@@ -316,6 +317,13 @@ additional flexibility in parsing log severity from incoming entries.
 ## Multi-Project exporting
 
 By default, the exporter sends telemetry to the project specified by `project` in the configuration. This can be overridden on a per-metrics basis using the `gcp.project.id` resource attribute. For example, if a metric has a label `project`, you could use the `groupbyattrs` processor to promote it to a resource label, and the `resource` processor to rename the attribute from `project` to `gcp.project.id`.
+
+### Multi-Project quota usage
+
+The `gcp.project.id` label can be combined with the `destination_project_quota` option to attribute quota usage to the project parsed by the label. This feature is currently only available
+for traces and metrics.
+
+Note that this option will not work  if a quota project is already defined in your Collector's GCP credentials. In this case, the telemetry will fail to export with a "project not found" error.
 
 ## Recommendations
 
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -137,7 +137,7 @@ The following configuration options are supported:
     User should calculate this as `num_seconds * requests_per_second` where:
     - `num_seconds` is the number of seconds to buffer in case of a backend outage
     - `requests_per_second` is the average number of requests per seconds.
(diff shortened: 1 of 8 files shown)


exporter/collector/README.md
```diff
@@ -317,6 +318,13 @@ additional flexibility in parsing log severity from incoming entries.
 
 By default, the exporter sends telemetry to the project specified by `project` in the configuration. This can be overridden on a per-metrics basis using the `gcp.project.id` resource attribute. For example, if a metric has a label `project`, you could use the `groupbyattrs` processor to promote it to a resource label, and the `resource` processor to rename the attribute from `project` to `gcp.project.id`.
 
+### Multi-Project quota usage
+
+The `gcp.project.id` label can be combined with the `destination_project_quota` option to attribute quota usage to the project parsed by the label. This feature is currently only available
```
dashpole, line 323: If I don't set gcp.project.id, but set `project` to a different value than my credentials' project, will this option still use quota from my destination project, or will it use quota from the credentials' project?

exporter/collector/metrics.go
```diff
@@ -68,12 +68,21 @@ type MetricsExporter struct {
...
+	cfg         Config
 	// goroutines tracks the currently running child tasks
 	goroutines sync.WaitGroup
 	timeout    time.Duration
 }
 
+// requestInfo is meant to abstract info from CreateMetricsDescriptorRequests and
+// CreateTimeSeriesRequests that is shared by requestOpts functions.
+type requestInfo struct {
+	name string
```
dashpole, line 83: maybe call this projectName or something more descriptive?

exporter/trace/cloudtrace.go
```diff
@@ -81,6 +84,14 @@ func WithProjectID(projectID string) func(o *options) {
 	}
 }
 
+// WithDestinationProjectQuota enables per-request usage of the destination
+// project's quota. For example, when setting gcp.project.id on a span.
```
dashpole, line 88: gcp.project.id is a resource attribute, rather than a span attribute, right?

exporter/collector/README.md
damemi: Yes, good call out. Updated the description of the collector setting above

exporter/collector/metrics.go
damemi: Sure. Tbh I'm not super happy with this approach in the first place, but it was the best way I could think of to re-use the functionality across the different metric export functions. I'm open to other architectural suggestions for this, but otherwise I did the rename. This could be useful in the future if we want to do other modifications before sending the request

exporter/trace/cloudtrace.go
damemi: Yeah, updated

damemi:
I tested this on a vm to see if default service account credentials also work (before I had only been testing with my local json credentials), and didn't have any problems.

exporter/collector/metrics.go
dashpole: maybe `type requestProject string`?  I'm fine with whatever you think is best.

exporter/collector/README.md
```diff
@@ -317,6 +318,13 @@ additional flexibility in parsing log severity from incoming entries.
...
+for traces and metrics. The Collector's default service account will need `roles/serviceusage.serviceUsageConsumer` IAM permissions in the destination quota project.
+
+Note that this option will not work  if a quota project is already defined in your Collector's GCP credentials. In this case, the telemetry will fail to export with a "project not found" error.
```
dashpole, line 326: It might be nice to provide instructions on how to ensure the quota project isn't set on the credentials.  `gcloud auth application-default set-quota-project ""` doesn't seem to work.

damemi: I just manually removed the line from my `.config/gcloud/application_default_credentials.json` file. Not sure if there is an official/recommended way to do this but I can at least add that step

dashpole (approved):

Issue #554, opened 2023-01-05
Migrate presubmits from circleci to github actions


dashpole:
The rest of the OpenTelemetry community uses github actions, and it is probably best for us to standardize on that.

dashpole:
Marking p2 given the recent security alert.  We weren't impacted, but it makes it better to do sooner rather than later.

PR #557, opened 2023-01-06
Add github actions tests

dashpole:
Part of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/554

This keeps all of the individual steps the same, but instead of running make ci, like we do in circleci, it runs each step (generate, build, lint, fixtures, check-clean-work-tree) as a separate step.

It also expands the compatibility matrix we test against to match what the opentelemetry-go repo uses to include testing against two go versions (1.19 and 1.18), testing on linux, windows, and mac.

It also adopts `gocovmerge` (github.com/wadey/gocovmerge) to properly merge code coverage reports from each module into a single one.

Tests are based on the github actions tests in opentelemetry-go.

After this, I will change the required tests in the repo, and then open a follow-up to remove circleci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,146 @@
+name: ci
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+env:
+  # Path to where test results will be saved.
+  TEST_RESULTS: /tmp/test-results
+  # Default minimum version of Go to support.
+  DEFAULT_GO_VERSION: 1.18
+jobs:
+  lint:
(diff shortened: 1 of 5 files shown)


dashpole:
not actually ready for review.  trying to figure out why GHA isn't running on the PR

dashpole:
See https://github.com/dashpole/opentelemetry-operations-go/pull/1 for the tests running on github actions.

dashpole:
This is now ready for review.

damemi (approved):

PR #551, opened 2022-12-21
WIP extending integration tests to more platforms

dashpole:

PR #562, opened 2023-01-09
Add resource mappings for cloud run, cloud functions, and GAE

dashpole:
Adds support for gae_instance, cloud_run_revision, and cloud_function monitored resources, but disallow their use for monitoring since those resources are not writable.  This is required to add e2e tests on those platforms.

This maintains the mappings if using the service.* labels (e.g. for cloud run).

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -31,6 +31,7 @@ import (
 	"go.opentelemetry.io/otel/sdk/metric/aggregation"
 	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
 	"go.uber.org/multierr"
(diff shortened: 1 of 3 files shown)


dashpole:
This is ready for review. It is a subset of the changes in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/551, which now passes the gae/cloud run presubmits.

exporter/metric/metric.go
```diff
@@ -337,9 +338,19 @@ func (attrs *attributes) GetString(key string) (string, bool) {
...
-	gmr := resourcemapping.ResourceAttributesToMonitoredResource(&attributes{
-		attrs: attribute.NewSet(res.Attributes()...),
-	})
+	attrs := &attributes{attrs: attribute.NewSet(res.Attributes()...)}
+	cloudPlatform, _ := attrs.GetString(string(semconv.CloudPlatformKey))
+	var gmr *resourcemapping.GceResource
+	switch cloudPlatform {
+	case semconv.CloudPlatformGCPCloudRun.Value.AsString():
+		fallthrough
+	case semconv.CloudPlatformGCPCloudFunctions.Value.AsString():
```
damemi, line 347: is there a tradeoff between `fallthrough` here and just making it a single case with multiple matches?

damemi (approved):

exporter/metric/metric.go
dashpole: Ah, thats a good point.

dashpole: done

PR #564, opened 2023-01-10
Include service attributes in logs exporter by default

damemi:
This copies the flow from metrics to include service resource attributes as LogEntry labels


damemi:
(cc @jsuereth)

dashpole:
Is this still a WIP?

damemi:
> Is this still a WIP?

no, sorry. going with the labels for now and opened #565 to determine if that's our long-term solution or not

dashpole (approved):

Issue #543, opened 2022-12-07
GMP exporter should follow OTel -> Prometheus export conventions


dashpole:
That includes:

* [x] Adding a `target_info` metric with resource attributes
* [x] Unit should be added as a suffix in metric names (before type suffixes)
* [x] Counters have a _total suffix added
* [x] otel_scope_info metric from scope
* [x] otel_scope_name and otel_scope_version attribute to all metrics within the scope, and to otel_scope_info

dashpole:
I'm going to work on the metric name portion of this

damemi:
Closing via #574

damemi:
Just noticed we never enabled these upstream. Opened https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24372 to add to contrib

PR #566, opened 2023-01-10
Add target_info metric for GMP exporter

damemi:
Part of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/543

This adds an extension point to the metric exporter's `PushMetrics()` function intended for making arbitrary modifications to metrics before export. The GMP exporter can use the `AddTargetInfo()` function that implements this to insert a new `target_info` metric for each monitored resource as specified in https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1


exporter/collector/config.go
```diff
@@ -96,7 +96,10 @@ type MetricConfig struct {
 	// exporter. It allows overriding the function used to map otel resource to
 	// monitored resource.
 	MapMonitoredResource func(pcommon.Resource) *monitoredrespb.MonitoredResource
-	Prefix               string `mapstructure:"prefix"`
+	// ModifyMetrics is an extension point for exporters to directly modify the set
+	// of ResourceMetrics during the call to PushMetrics.
```
dashpole, line 100: Note that this means we need to mark the exporter as modifying the metrics (means an extra copy before export).  This may be OK for now.  Are there other ways we might be able to accomplish this?

dashpole: See https://github.com/open-telemetry/opentelemetry-collector/blob/main/consumer/consumer.go#L22

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -52,6 +53,27 @@ func MapToPrometheusTarget(res pcommon.Resource) *monitoredrespb.MonitoredResour
 	}
 }
 
+// AddTargetInfo extracts the target_info metric from each ResourceMetric associated with the input pmetric.Metrics
+// and inserts it into each ScopeMetric for that resource, with the matching "job" and "instance" labels as specified in
+// https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1
+func AddTargetInfo(m pmetric.Metrics) {
```
dashpole, line 59: can you put this in its own file?  This one is called monitoredresource.go, and this probably doesn't fit there.

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json
```diff
@@ -641,6 +641,37 @@
               }
             }
           ]
+        },
+        {
+          "metric": {
+            "type": "prometheus.googleapis.com/target_info/gauge",
+            "labels": {
+              "instance": "10.92.5.2:15692",
```
dashpole, line 649: I don't think job/instance should be included in metric labels if they are already in resource labels.  IIRC, GMP will reject this point.

damemi: From the spec (https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1) I thought the point was that these would be included as metric labels?

exporter/collector/googlemanagedprometheus/monitoredresource.go
damemi: Sure, I just put it here because it uses the gmp MR so it felt like it fit

exporter/collector/config.go
damemi: Is there any way around this if we want to inject a new metric? eg, only making this insert the metric rather than an arbitrary modification path. Or is even that going to require this update?

dashpole: You might be able to Change from a `func ModifyMetrics(pmetric.Metrics)` to `func ExtraMetrics(pmetric.Metrics) pmetric.MetricSlice`, where the metricslice isn't ever actually added to the resource metrics, but is used _as if_ it was part of the resource metrics

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json
dashpole: Prometheus has a "flat" label model.  There isn't such a thing as resource labels vs metric labels.  GMP hides the implementation from users by internally representing some labels (job, instance, cluster, namespace) as resource labels rather than metric labels, but they look the same to users at query time.

But in order for that to work, you can't have the same label key in both resource and metric labels in the metric sent to GCM.  Otherwise GMP doesn't know which value to use, so they just reject it.

dashpole: We need to add all _other_ resource labels (except those in the monitored resource) as metric labels on the metric (e.g. service.name -> service_name)

exporter/collector/config.go
damemi: I ended up going with a copy approach, I think that leaves the original metrics in tact. ptal if that's right

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json
damemi: done, also refactored how those labels are tracked a little too. maybe overthinking it

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -0,0 +1,76 @@
...
+				nonSpecialResourceAttributes[attributeKey] = value.AsString()
+			}
+		}
+
+		// copy the Resource to a new ResourceMetric in our target_info slice, so we don't lose its attributes
+		rm.Resource().CopyTo(resourceMetricSlice.AppendEmpty().Resource())
+
+		// loop over this resource's scopeMetrics and copy the scope info to our target_info slice
+		scopeMetricSlice := resourceMetricSlice.At(i).ScopeMetrics()
+		sms := rm.ScopeMetrics()
```
dashpole, line 56: Would this produce a target_info for each scope in the input?  If so, we don't want to do that.  We want to add one target_info for each resource, rather than one for each scope.

dashpole (commented):
wip Review

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -0,0 +1,76 @@
...
+	resourceMetricSlice := pmetric.NewResourceMetricsSlice()
+	rms := m.ResourceMetrics()
+	// loop over input (original) resource metrics
+	for i := 0; i < rms.Len(); i++ {
+		rm := rms.At(i)
+
+		// store the attributes for this resource that will be added as metric labels to target_info
+		nonSpecialResourceAttributes := make(map[string]string)
+		for attributeKey := range rm.Resource().Attributes().AsRaw() {
+			special := false
```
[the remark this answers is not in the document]
damemi: good point, done

exporter/collector/googlemanagedprometheus/extra_metrics.go
damemi: yeah, updated to not do that (and added it to the unit test)

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -0,0 +1,64 @@
...
+		rm.Resource().CopyTo(resourceMetricSlice.AppendEmpty().Resource())
+
+		// create the target_info metric as a Gauge with value 1
+		targetInfoMetric := resourceMetricSlice.At(i).ScopeMetrics().AppendEmpty().Metrics().AppendEmpty()
+		targetInfoMetric.SetName("target_info")
+		targetInfoMetric.SetEmptyGauge().DataPoints().AppendEmpty().SetIntValue(1)
+
+		// copy Resource attributes to the metric except for attributes which will already be present in the MonitoredResource labels
+		for key, value := range nonSpecialResourceAttributes {
+			targetInfoMetric.Gauge().DataPoints().At(0).Attributes().PutStr(key, value)
```
dashpole, line 49:
```suggestion
		dataPoint := targetInfoMetric.SetEmptyGauge().DataPoints().AppendEmpty()
		dataPoint.SetIntValue(1)

		// copy Resource attributes to the metric except for attributes which will already be present in the MonitoredResource labels
		for key, value := range nonSpecialResourceAttributes {
			dataPoint.Attributes().PutStr(key, value)
```
nit: keeping a reference to dataPoint seems slightly more readable

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -31,23 +31,34 @@ const (
 	namespaceLabel = "namespace"
 )
 
+// promTargetKeys are attribute keys which are used in the prometheus_target monitored resource.
+// It is also used by GMP to exclude these keys from the target_info metric.
+var promTargetKeys = map[string][]string{
+	locationLabel:       {locationLabel, semconv.AttributeCloudAvailabilityZone, semconv.AttributeCloudRegion},
+	clusterLabel:        {clusterLabel, semconv.AttributeK8SClusterName},
+	namespaceLabel:      {namespaceLabel, semconv.AttributeK8SNamespaceName},
+	"job":               {semconv.AttributeServiceName},
```
dashpole, line 40: nit: should we make these strings constants above?  Makes typos less likely.

exporter/collector/metrics.go
```diff
@@ -190,7 +190,17 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+	m.ResourceMetrics().CopyTo(rms)
+
+	// add extra metrics from the ExtraMetrics() extension point, combine into a new copy
+	extraResourceMetrics := pmetric.NewResourceMetricsSlice()
+	if me.cfg.MetricConfig.ExtraMetrics != nil {
+		extraResourceMetrics = me.cfg.MetricConfig.ExtraMetrics(m)
+	}
+	extraResourceMetrics.MoveAndAppendTo(rms)
+
 	for i := 0; i < rms.Len(); i++ {
```
dashpole, line 204: It seems like it would be easy enough to make this a function (e.g. pushResourceMetrics(ctx context.Context, rms pmetric.ResourceMetrics) error), and then call it twice: Once for actual metrics, and once for target_info.  Then we could avoid the copy.

exporter/collector/metrics.go
```diff
@@ -190,7 +190,17 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
 	// they need to be sent to. Each project's timeseries are sent in a
 	// separate request later.
 	pendingTimeSeries := map[string][]*monitoringpb.TimeSeries{}
-	rms := m.ResourceMetrics()
+
+	rms := pmetric.NewResourceMetricsSlice()
+	m.ResourceMetrics().CopyTo(rms)
```
dashpole, line 195: One downside of this approach we should avoid if possible: This performs a copy even if `me.cfg.MetricConfig.ExtraMetrics == nil`, which will be the case in our normal exporter.  We should avoid this case if possible. Even if you don't accept my suggestion below.

exporter/collector/googlemanagedprometheus/extra_metrics.go
damemi: sounds good, done

exporter/collector/googlemanagedprometheus/monitoredresource.go
damemi: Yeah they are used in a couple places. done

exporter/collector/metrics.go
damemi: Moved this around to avoid the copy if `ExtraMetrics == nil`, ptal

exporter/collector/metrics.go
damemi: Keeping it in a single call feels cleaner to me than adding a special path. This way whether you're debugging normal exporter or GMP, the logic is the same. I did make your change to avoid unnecessary copies as you suggested below though, so ptal

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -0,0 +1,66 @@
...
+		targetInfoMetric := resourceMetricSlice.At(i).ScopeMetrics().AppendEmpty().Metrics().AppendEmpty()
+		targetInfoMetric.SetName("target_info")
+
+		dataPoint := targetInfoMetric.SetEmptyGauge().DataPoints().AppendEmpty()
+		dataPoint.SetIntValue(1)
+
+		// copy Resource attributes to the metric except for attributes which will already be present in the MonitoredResource labels
+		for key, value := range nonSpecialResourceAttributes {
+			dataPoint.Attributes().PutStr(key, value)
+		}
```
dashpole, line 52:
```suggestion
		// copy the Resource to a new ResourceMetric in our target_info slice, so we don't lose its attributes
		rm.Resource().CopyTo(resourceMetricSlice.AppendEmpty().Resource())

		// create the target_info metric as a Gauge with value 1
		targetInfoMetric := resourceMetricSlice.At(i).ScopeMetrics().AppendEmpty().Metrics().AppendEmpty()
		targetInfoMetric.SetName("target_info")

		dataPoint := targetInfoMetric.SetEmptyGauge().DataPoints().AppendEmpty()
		dataPoint.SetIntValue(1)

		// copy Resource attributes to the metric except for attributes which will already be present in the MonitoredResource labels
		for attributeKey := range rm.Resource().Attributes().AsRaw() {
			if !isSpecialAttribute(attributeKey) {
				value, _ := rm.Resource().Attributes().Get(attributeKey)
				dataPoint.Attributes().PutStr(attributeKey, value.AsString())
			}
		}
```
Avoid creating the nonSpecialResourceAttributes map.

dashpole (approved):
lgtm with a few optimizations

dashpole:
(maybe try rebasing?)

PR #570, opened 2023-01-12
 Remove monitored resource mappings for Cloud run and cloud functions

dashpole:
Partial revert of #562.

It removes MR mappings for cloud run and cloud functions.  It keeps the GAE mappings, since that is useful when writing metrics.

This should break e2e tests until we switch to checking otel resource attributes.


dashpole:
cc @aabmass

dashpole:
This will need https://github.com/GoogleCloudPlatform/opentelemetry-operations-e2e-testing/pull/25 to pass integration tests.

damemi (approved):

PR #571, opened 2023-01-17
Add otel_scope_info metric to GMP exporter

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/543

Following the spec at https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1, this adds an `otel_scope_info` metric:

* `otel_scope_name` and `otel_scope_version` are populated if present on the scope
* additional scope attributes are added as labels
* this is broken into its own function so that it can be optionally configurable through the exporter

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -19,10 +19,11 @@ import (
 	"go.opentelemetry.io/collector/pdata/pmetric"
 )
 
-// AddTargetInfo extracts the target_info metric from each ResourceMetric associated with the input pmetric.Metrics
+// AddTargetInfoMetric inserts target_info for each resource.
+// First, it extracts the target_info metric from each ResourceMetric associated with the input pmetric.Metrics
 // and inserts it into each ScopeMetric for that resource, as specified in
 // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1
-func AddTargetInfo(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
+func AddTargetInfoMetric(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
 	// initialize a new ResourceMetricSlice, which will hold our target_info metrics for each RM/Scope
 	resourceMetricSlice := pmetric.NewResourceMetricsSlice()
 	rms := m.ResourceMetrics()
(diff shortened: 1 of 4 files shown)


exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -51,6 +52,36 @@ func AddTargetInfo(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
...
+			sm := sms.At(j)
+
+			scopeInfoMetric := resourceMetricSlice.At(i).ScopeMetrics().AppendEmpty().Metrics().AppendEmpty()
+			scopeInfoMetric.SetName("otel_scope_info")
+
+			dataPoint := scopeInfoMetric.SetEmptyGauge().DataPoints().AppendEmpty()
+			dataPoint.SetIntValue(1)
+
+			dataPoint.Attributes().PutStr("otel_scope_name", sm.Scope().Name())
+			dataPoint.Attributes().PutStr("otel_scope_version", sm.Scope().Version())
```
damemi, line 75: if name/version aren't present, should we just not add these labels? wasn't sure if that would be compatible with other metrics that do have them

exporter/collector/googlemanagedprometheus/extra_metrics_test.go
```diff
@@ -63,9 +66,103 @@ func TestAddTargetInfo(t *testing.T) {
...
+
+				sm := rm.ScopeMetrics().AppendEmpty()
+				metric := sm.Metrics().AppendEmpty()
+				metric.SetName("target_info")
+				metric.SetEmptyGauge().DataPoints().AppendEmpty().SetIntValue(1)
+				metric.Gauge().DataPoints().At(0).Attributes().PutStr("foo-label", "bar")
+
+				rm2 := rms.AppendEmpty()
+				rm2.Resource().Attributes().PutStr(locationLabel, "us-east")
+				rm2.Resource().Attributes().PutStr("foo-label", "bar")
```
damemi, line 152: note that the way I'm doing this, the resources get copied over (we're basically building 2 separate metricSlices and appending them to the main slice). is this ok?

exporter/collector/googlemanagedprometheus/extra_metrics.go
dashpole: We should still add them.  When doing a join between otel_scope_info and other metrics, the empty scope should still end up matching the metrics contained within the empty scope.

exporter/collector/googlemanagedprometheus/extra_metrics_test.go
dashpole: Of course we rather would not, but if we have to it is fine.  What would the alternative be?  Make the function return a ScopeMetricSlice instead of a ResourceMetricSlice?

damemi: Yeah, I was trying to keep both functions with the same signature as the ExtraMetrics extension point for future flexibility but I guess it's not really necessary

exporter/collector/googlemanagedprometheus/extra_metrics.go
damemi: that's what I thought, but wanted to check 👍

exporter/collector/googlemanagedprometheus/extra_metrics_test.go
dashpole: couldn't target info also be returned as a ScopeMetricSlice?

damemi: We need a ScopeMetricSlice for each Resource though right? How about the changes in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/571/commits/6be421c043c1b9acfe9cc5cfdaa31b676fac582a#diff-0af4ea6286cefe60de85c9cad15924d571a1755ae70b74638c2b36355c46554aR108-R111

dashpole (approved):

PR #574, opened 2023-01-19
Add scope_info labels to all GMP metrics where available

damemi:
Follow up https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/571, due to misinterpreting the spec. `otel_scope_` labels should be applied to all metric data points in a given scope, not just metric data points for the `otel_scope_info` metric.

Previously, the ExtraMetrics extension point would just return a list of *new* metrics, to be merged into the existing list. Since we now want to modify metrics from the original list to add these labels, this had to be changed.

This changes the `AddTargetInfoMetric` and `AddScopeInfoMetric` functions to directly modify the input `pmetric.Metrics` object. To ensure original metrics aren't modified, the ExtraMetrics extension point in `metrics.go` has been updated to handle copying the input metrics and passing that copy to the ExtraMetrics function.

Ref https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/543

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -52,7 +52,6 @@ func AddTargetInfoMetric(m pmetric.Metrics) {
 func AddScopeInfoMetric(m pmetric.Metrics) {
 	rms := m.ResourceMetrics()
 	for i := 0; i < rms.Len(); i++ {
-
 		sms := rms.At(i).ScopeMetrics()
 		for j := 0; j < sms.Len(); j++ {
 			sm := sms.At(j)
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -21,21 +21,16 @@ import (
 
 // AddTargetInfoMetric inserts target_info for each resource.
 // First, it extracts the target_info metric from each ResourceMetric associated with the input pmetric.Metrics
-// and inserts it into each ScopeMetric for that resource, as specified in
+// and inserts it into a new ScopeMetric for that resource, as specified in
 // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1
-func AddTargetInfoMetric(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
-	// initialize a new ResourceMetricSlice, which will hold our target_info metrics for each RM/Scope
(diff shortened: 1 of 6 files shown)


exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -135,12 +135,9 @@ var MetricsTestCases = []TestCase{
...
 			cfg.MetricConfig.MapMonitoredResource = googlemanagedprometheus.MapToPrometheusTarget
 			cfg.MetricConfig.ExtraMetrics = func(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
-				extraMetrics := googlemanagedprometheus.AddTargetInfoMetric(m)
-				scopeInfoMetrics := googlemanagedprometheus.AddScopeInfoMetric(m)
-				for i := 0; i < scopeInfoMetrics.Len(); i++ {
-					scopeInfoMetrics.At(i).ScopeMetrics().MoveAndAppendTo(extraMetrics.At(i).ScopeMetrics())
-				}
-				return extraMetrics
+				googlemanagedprometheus.AddScopeInfoMetric(m)
+				googlemanagedprometheus.AddTargetInfoMetric(m)
```
damemi, line 139: this ordering is important (first scope_info, then target_info), otherwise the `otel_scope_info` metric is also added to the new (empty) ScopeMetrics that holds `target_info` for a resource. I actually wasn't sure on this, but took an educated guess

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -49,36 +44,80 @@ func AddTargetInfoMetric(m pmetric.Metrics) pmetric.ResourceMetricsSlice {
...
 			dataPoint.SetIntValue(1)
-
-			dataPoint.Attributes().PutStr("otel_scope_name", sm.Scope().Name())
-			dataPoint.Attributes().PutStr("otel_scope_version", sm.Scope().Version())
 			sm.Scope().Attributes().Range(func(k string, v pcommon.Value) bool {
 				dataPoint.Attributes().PutStr(k, v.AsString())
 				return true
 			})
+
+			// If present, add scope_name and scope_version attributes to each datapoint (including otel_scope_info)
```
dashpole, line 70: If they aren't present, we should skip the otel_scope_info metric entirely

exporter/collector/integrationtest/testcases/testcases_metrics.go
dashpole: This shouldnt' matter if we adopt my recommendation above.

dashpole (approved):

Issue #544, opened 2022-12-07
Support gzip compression for metric export


dashpole:
For off-GCP workloads, it can be helpful to reduce network bandwidth by enabling compression.  That can be done with something like this:  https://github.com/GoogleCloudPlatform/prometheus-engine/blob/2e01f7369ed284c77a401c63b53b145eeeec87b1/pkg/export/export.go#LL261C4-L261C55

It isn't clear to me if that can be used for all signal types, or just for monitoring

aabmass:
"High p2" because this would be a very useful feature. Mike is still planning to work on this.

PR #576, opened 2023-01-26
Add gzip compression option to metrics and logs

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/544

This copies GMP's [compression setting](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/30ce1d152c6ecfcde8e57b4553dadc895b233708/pkg/operator/apis/monitoring/v1/types.go#L115) by exposing a string option for users to specify the type of compression to enable for gRPC calls. Currently the only supported option is `gzip`.

This only works for logs and metrics, trying to wire it up for Cloud Trace gave this error:
```
2023/01/26 19:31:25 failed to export to Google Cloud Trace: rpc error: code = InvalidArgument desc = Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest
error details: name = DebugInfo detail = Was unable to extract resource containers for Chemist from the incoming request. Error message during extraction: Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest stack =
2023-01-26T19:31:25.980Z	info	exporterhelper/queued_retry.go:426	Exporting failed. Will retry the request after interval.	{"kind": "exporter", "data_type": "traces", "name": "googlecloud", "error": "rpc error: code = InvalidArgument desc = Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest\nerror details: name = DebugInfo detail = Was unable to extract resource containers for Chemist from the incoming request. Error message during extraction: Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest stack =", "interval": "6.722764325s"}
2023/01/26 19:31:32 failed to export to Google Cloud Trace: rpc error: code = InvalidArgument desc = Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest
error details: name = DebugInfo detail = Was unable to extract resource containers for Chemist from the incoming request. Error message during extraction: Error skipping unknown field of a message with tag:31, type:google.devtools.cloudtrace.v2.BatchWriteSpansRequest stack =
```

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -31,12 +31,11 @@ import (
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/encoding/gzip"
 )
 
 const (
 	DefaultTimeout = 12 * time.Second // Consistent with Cloud Monitoring's timeout
-
-	CompressionGzip = "gzip"
 )
 
 // Config defines configuration for Google Cloud exporter.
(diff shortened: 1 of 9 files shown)


exporter/collector/config.go
```diff
@@ -211,6 +216,14 @@ func ValidateConfig(cfg Config) error {
 			return fmt.Errorf("unable to parse resource filter regex: %s", err.Error())
 		}
 	}
+
+	if len(cfg.LogConfig.ClientConfig.Compression) > 0 && cfg.LogConfig.ClientConfig.Compression != CompressionGzip {
+		return fmt.Errorf("unknown compression option '%s', allowed values: '', 'gzip'")
+	}
+	if len(cfg.MetricConfig.ClientConfig.Compression) > 0 && cfg.MetricConfig.ClientConfig.Compression != CompressionGzip {
```
dashpole, line 223: should we disallow setting any option for the trace exporter?

exporter/collector/config.go
```diff
@@ -35,6 +35,8 @@ import (
 
 const (
 	DefaultTimeout = 12 * time.Second // Consistent with Cloud Monitoring's timeout
+
+	CompressionGzip = "gzip"
```
dashpole, line 39: nit: does this need to be exported?

dashpole: also seems like you use `gzip.Name` as a constant elsewhere.  Might be good to use that instead.

dashpole (approved):

damemi:
Trying to add tests to this, my goal is to send gzip-compressed metrics/logs to our mock servers and have the mock return an error (because the mock doesn't have the gzip compressor registered). Then, ideally a 2nd test with a mock server that has gzip decoding enabled would pass.

However it looks like the gzip compressor is registered automatically by importing the `gzip` package https://stackoverflow.com/a/49507459. Calling [`encoding.GetCompressor("gzip")`](https://pkg.go.dev/google.golang.org/grpc/encoding#GetCompressor) in the mock confirms this.

So, I haven't found a way to get the error to trigger. I did add a test that sets compression and expects no error, so that is something at least.

dashpole (approved):

Issue #558, opened 2023-01-09
Refactor collector logs exporter to support multi-project quota


damemi:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/555 added multi-project quota support for traces and metrics, but the logs exporter currently batches log entries from potentially multiple projects into a single request.

Since the cross-project quota is set by a header on the request, the logs exporter will need to be refactored so that batches are composed of logs for the same destination project.

dashpole:
Still useful

PR #579, opened 2023-02-02
Enable destination_project_quota for logs exporter

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/558

Adding this was pretty simple, the current logic builds a single LogEntry list and then batches those entries as needed to keep each batch request under the GCP size limit. We can do this without caring about the project for each entry because, while batch LogEntry requests do have a field for project, the request will honor individual project IDs set on each entry in the batch. (Traces and Metrics don't have that)

However, adding destination project quota brings us back to caring about the project at the request/batch level. That's easy to do using a `map[string]LogEntry`, which is similar to how traces and metrics batch.

But, it would be optimal to revert to the efficient, possibly mixed-project batching when `destination_project_quota` is disabled. The best I could think of to do this is to drop everything in a single key in the map, and reuse the rest of the logic. I'm not very satisfied with that, because I don't like using a magic key and it couples the `destination_project_quota` config option to multiple places in the code. So this might be more refactorable, but for unblocking the feature this will do fine imo, but I'm open to ideas

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/logs/logs_multi_project_destination_quota_expected.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/logs/logs_multi_project_destination_quota_expected.json
@@ -1,0 +1,780 @@
+{
+  "writeLogEntriesRequests": [
+    {
+      "entries": [
+        {
+          "logName": "projects/fakeprojectid/logs/multi-project",
+          "resource": {
+            "type": "gce_instance",
+            "labels": {
+              "instance_id": "",
+              "zone": ""
+            }
+          },
+          "textPayload": "127.0.0.1 - - [26/Apr/2022:22:53:36 +0800] \"GET / HTTP/1.1\" 200 1247",
+          "timestamp": "1970-01-01T00:00:00Z",
+          "httpRequest": {
+            "requestMethod": "GET",
+            "requestUrl": "/",
+            "status": 200,
+            "responseSize": "1247",
+            "remoteIp": "127.0.0.1",
+            "protocol": "HTTP/1.1"
+          },
+          "labels": {
+            "log.file.name": "test.log"
+          }
+        },
+        {
+          "logName": "projects/fakeprojectid/logs/multi-project",
+          "resource": {
+            "type": "gce_instance",
+            "labels": {
+              "instance_id": "",
+              "zone": ""
+            }
+          },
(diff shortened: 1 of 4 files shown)


exporter/collector/integrationtest/testdata/fixtures/logs/logs_multi_project_destination_quota_expected.json
```diff
@@ -0,0 +1,780 @@
...
+          },
+          "textPayload": "127.0.0.1 - - [26/Apr/2022:22:54:43 +0800] \"-\" 408 -",
+          "timestamp": "1970-01-01T00:00:00Z",
+          "labels": {
+            "log.file.name": "test.log"
+          }
+        }
+      ],
+      "partialSuccess": true
+    },
```
damemi, line 390: note that this fixture differs from `logs_multi_project_expected.json` here, with 2 lists of `entries`

exporter/collector/logs.go
```diff
@@ -215,9 +223,12 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld plog.Logs) error {
 	return nil
 }
 
-func (l logMapper) createEntries(ld plog.Logs) ([]*logpb.LogEntry, error) {
+func (l logMapper) createEntries(ld plog.Logs) (map[string][]*logpb.LogEntry, error) {
+	// if destination_project_quota is enabled, projectMapKey will be the name of the project for each batch of entries
+	// otherwise, we can mix project entries for more efficient batching and store all entries in a single list
+	projectMapKey := "google-destination-quota-disabled"
```
dashpole, line 229: could this just always be the destination project id instead of this string?  Then other functions could assume the key of `map[string][]*logpb.LogEntry` is the destination project?

damemi: if destination quota is disabled, entries can be unsorted (wrt project), so using a default key packs them into a single list that is used by the other functions, and the other functions ignore the key (no need to set any project id in the request)

with destination quota enabled, this key does become the destination project and each entry gets sorted into its project's list. then the other function uses that key to set the x-goog header

dashpole: Ah, got it.  I'd prefer using the empty string, as someone might think the constant is important...  But not a big deal.

dashpole (approved):
looks great.

exporter/collector/logs.go
damemi: Oh empty string is much better

PR #581, opened 2023-02-03
Actually add logs fixture sorting

damemi:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/580 didn't work because I was just checking the `request.LogName` field, which our exporter doesn't set (we set individual entry log names ie, `request.Entries[i].LogName`). I also forgot to add the matching sort to the fixture generator.

The tests still passed due to the randomness of the fixture ordering (until post-merge https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/actions/runs/4086503881). But, I should have caught this based on the fact that fixtures weren't updated in #580.

This sorts each *entry* in each *request*, then sorts the requests based on either their LogName or the LogName of the first entry in the request. This should be sufficiently deterministic for our tests.

### DIFF
--- a/exporter/collector/integrationtest/logs_test.go
+++ b/exporter/collector/integrationtest/logs_test.go
@@ -54,15 +54,38 @@ func TestLogs(t *testing.T) {
 				t,
 				timestamp,
 			)
+
+			// sort the entries in each request
+			for listIndex := 0; listIndex < len(expectFixture.WriteLogEntriesRequests); listIndex++ {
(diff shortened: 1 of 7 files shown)


damemi:
failing even harder now, lol. looking into this

damemi:
comparing just by `entry.LogName` didn't always produce the same output either (for the record, these passed locally for me but failed in the CI environment)

I went all out and just compared the String() output of the protobuf message in each entry. that generated even more fixture changes, and it looks like it passed both locally and in CI now.

dashpole (approved):

Issue #584, opened 2023-02-07
GMP metric integration test is broken


dashpole:
As part of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/583, I found that the GMP tests are not passing when run against a real GCM endpoint.

Failure:
```
    metrics_integration_test.go:72: 
        	Error Trace:	/workspace/exporter/collector/integrationtest/metrics_integration_test.go:72
        	Error:      	Received unexpected error:
        	            	failed to export time series to GCM: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Request was missing field timeSeries[21].points[0].interval.endTime: The end time of the interval is required.; Request was missing field timeSeries[1].points[0].interval.endTime: The end time of the interval is required.
        	Test:       	TestIntegrationCollectorLogs/Google_Managed_Prometheus
```

damemi:
I'll take a look into this, thanks @dashpole

damemi:
At first I assumed I just needed to add a timestamp to the new metrics, but it's [already there](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json#L735).

This is set by the exporter calling [`timestamppb.New(point.Timestamp().AsTime())`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c7cfa7a5ababcdd1440d1479cbc9d4df041fe08/exporter/collector/metrics.go#L896). I assume this just defaults to the zero time when it's not set. But, I'm wondering if this is working right for any of our points. For example, [this timestamp](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c7cfa7a5ababcdd1440d1479cbc9d4df041fe08/exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus.json#L340) is `Fri Apr 08 2022 18:45:16 GMT+0000` but comes out as [`1970-01-01T00:00:00Z`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0c7cfa7a5ababcdd1440d1479cbc9d4df041fe08/exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json#L617). The rest of the timestamps in that expect fixture are also the zero time.

[...]

PR #585, opened 2023-02-07
Add timestamp to scope_info and target_info GMP metrics

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/584

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -15,6 +15,8 @@
 package googlemanagedprometheus
 
 import (
+	"time"
+
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 )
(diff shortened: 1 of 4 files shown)


dashpole:
I think it is reasonable to fix it this way for now, and we should try to make sure we get this fixed upstream ASAP, since it produces errors when using the GMP exporter.  We should make an issue to track a solution to the problems this creates below:

It is possible that the collector will be exporting a batch of metrics that are old (e.g. if metrics have been persisted to disk for an hour from on-prem before sending if there was a network outage).  In that case, the timestamp of the target_info metric would be now() but the timestamp on the metrics would be an hour ago, which is odd.  It could also cause the timestamps of multiple batches to be very close together if, for example, i'm backfilling historical data sequentially.  In that case, the timestamps of my regular metrics would be far apart (because they were collected at regular intervals), but if they are uploaded quickly, I would get a bunch of target_info and scope_info metrics that are really close together in time, and could cause write errors.

The [solution adopted by the PRW exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7aac7c8f0fe31ad0eb07484018c22ec954822463/pkg/translator/prometheusremotewrite/metrics_to_prw.go#L47) is to use the most recent timestamp in the metrics contained within the resource/scope.

damemi:
@dashpole no I agree doing it right makes sense now. should we revert the broken release or just publish the new one asap?

dashpole (approved):
nice.  Didn't think the fix would be that quick

dashpole:
feel free to merge with coverage checks failing.

PR #587, opened 2023-02-09
try with lots of attributes

dashpole:
is the limit really 32?


dashpole:
I got labels: 


Label | Value
-- | --
test10 | foo
test11 | foo
test13 | foo
test15 | foo
test16 | foo
test17 | foo
test18 | foo
test19 | foo
test21 | foo
test22 | foo
test23 | foo
test25 | foo
test26 | foo
test27 | foo
test28 | foo
test29 | foo
test3 | foo
test30 | foo
test31 | foo
test32 | foo
test34 | foo
test35 | foo
test36 | foo
test38 | foo
test4 | foo
test40 | foo
test5 | foo
test6 | foo
test7 | foo
test8 | foo
test9 | foo
test_id | 7465461270372835454

It looks like the following labels were dropped somewhere along the way:

* test1
* test2
* test12
* test14
* test20
* test24
* test33
* test37
* test39

PR #591, opened 2023-02-14
Fix infinite loop on log exporter batch write

### LINKED ISSUE #590: PushLogs in exporter can get stuck in infinite loop
In [PushLogs](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L174) the exporter can get stuck in an infinite loop if a batch is large enough to be split over several requests to GCP.

If a batch needs to be split after the first request `currentBatchSize` is not reset with the rest of the loop variables [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L215-L216). This leads to the next iteration that the entry at index 0 will be calculated as too large. ~~I believe there is an off by one error in the slicing logic which will end up sending an empty request.~~

```go
_, err := l.writeLogEntries(ctx, entries[:entry])
if err != nil {
	errors = append(errors, err)
}

entries = entries[entry:]
```

In the above code when `entry` is 0 `entries[:entry]` will be an empty slice and `entries[entry:]` will reset to the same slice. Since we never reset `currentBachSize` and we keep iterating over the same slice, `PushLogs` will keep sending empty requests.

damemi:
Fixes #590 

This could have been prevented if we had originally implemented integration tests for batch request sizing ([as TODO'd here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/a62ae2cc5958c5ea59cbc9064cd4a76a4af25da9/exporter/collector/logs.go#L185))

We do already have unit tests for individual entry sizing, but we chose not to block on request sizing for a few reasons:
1. Requests are sized in `PushLogs()`, which means they likely have to be an integration test as done here.
2. The integration tests work with an actual exporter, so we would either need to create test logs that exceeded the default GCP size limit (10 MB) or configure the test exporter to use a different max size.
3. Configuring the test exporter isn't very easy without exposing that configuration option due to running across Go packages.
4. We didn't feel the need to expose this config option anyway, because the only benefit to users would be to set a *lower* max request size, which would just result in more requests (== higher costs?)
5. 10 MB is a pretty big request, and we didn't realistically expect that to be hit.

We are now running into point number 5. So, the best option I have is to add these options (max_entry_size and max_request_size) to the LogExporter config. Though we don't need to document them, because they aren't very useful, there isn't technically any harm in someone choosing to set these manually.

Another option I suppose would be exposing just the variables and setting those directly. That would keep the collector config cleaner, but imo it's a wash either way.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -46,8 +46,8 @@ type Config struct {
 	ProjectID               string            `mapstructure:"project"`
 	UserAgent               string            `mapstructure:"user_agent"`
 	ImpersonateConfig       ImpersonateConfig `mapstructure:"impersonate"`
-	TraceConfig             TraceConfig       `mapstructure:"trace"`
 	LogConfig               LogConfig         `mapstructure:"log"`
+	TraceConfig             TraceConfig       `mapstructure:"trace"`
 	MetricConfig            MetricConfig      `mapstructure:"metric"`
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -46,9 +46,9 @@ type Config struct {
 	ProjectID               string            `mapstructure:"project"`
 	UserAgent               string            `mapstructure:"user_agent"`
 	ImpersonateConfig       ImpersonateConfig `mapstructure:"impersonate"`
-	LogConfig               LogConfig         `mapstructure:"log"`
 	TraceConfig             TraceConfig       `mapstructure:"trace"`
 	MetricConfig            MetricConfig      `mapstructure:"metric"`
+	LogConfig               LogConfig         `mapstructure:"log"`
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
 
@@ -163,6 +163,12 @@ type LogConfig struct {
 	// Defaults to empty, which won't include any additional resource labels.
 	ResourceFilters []ResourceFilter `mapstructure:"resource_filters"`
 	ClientConfig    ClientConfig     `mapstructure:",squash"`
+	// MaxEntrySize defines the maximum size of an individual LogEntry in bytes. If a LogEntry
+	// is larger than this size, it will be split and sent as multiple entries. Default: 256000 (256KB).
+	MaxEntrySize int `mapstructure:"max_entry_size"`
+	// MaxRequestSize defines the maximum size of a batch WriteLogEntriesRequest in bytes. If a request
+	// is larger than this size, it will be split into multiple requests. Default: 10000000 (10MB).
+	MaxRequestSize int `mapstructure:"max_request_size"`
 	// ServiceResourceLabels, if true, causes the exporter to copy OTel's
 	// service.name, service.namespace, and service.instance.id resource attributes into the Cloud Logging LogEntry labels.
 	// Disabling this option does not prevent resource_filters from adding those labels. Default is true.
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -47,8 +47,8 @@ type Config struct {
 	UserAgent               string            `mapstructure:"user_agent"`
 	ImpersonateConfig       ImpersonateConfig `mapstructure:"impersonate"`
 	TraceConfig             TraceConfig       `mapstructure:"trace"`
-	MetricConfig            MetricConfig      `mapstructure:"metric"`
 	LogConfig               LogConfig         `mapstructure:"log"`
+	MetricConfig            MetricConfig      `mapstructure:"metric"`
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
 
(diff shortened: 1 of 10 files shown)


exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_access_batches_expected.json
```diff
@@ -0,0 +1,438 @@
...
+          "labels": {
+            "log.file.name": "test.log",
+            "single-attribute": "foobar"
+          }
+        }
+      ],
+      "partialSuccess": true
+    },
+    {
+      "entries": [
```
damemi, line 33: the key here: this is splitting into multiple requests ("entries").

exporter/collector/logs.go
```diff
@@ -214,6 +224,7 @@ func (l *LogsExporter) PushLogs(ctx context.Context, ld plog.Logs) error {
 
 			entries = entries[entry:]
 			entry = 0
+			currentBatchSize = 0
```
damemi, line 227: confirmed this was the fix by commenting this line out which reproduced the infinite loop on the new test case

exporter/collector/config.go
```diff
@@ -163,6 +163,12 @@ type LogConfig struct {
 	// Defaults to empty, which won't include any additional resource labels.
 	ResourceFilters []ResourceFilter `mapstructure:"resource_filters"`
 	ClientConfig    ClientConfig     `mapstructure:",squash"`
+	// MaxEntrySize defines the maximum size of an individual LogEntry in bytes. If a LogEntry
+	// is larger than this size, it will be split and sent as multiple entries. Default: 256000 (256KB).
+	MaxEntrySize int `mapstructure:"max_entry_size"`
```
dashpole, line 168: If we don't expect anyone to set this, should we omit it from the config?  Is it actually neccessary?

damemi: We don't need it, it just felt weird to have request size configurable without it

dashpole: Oh, I see.  We need to be able to set MaxRequestSize, but not MaxEntrySize?

damemi: yeah, for the integration test we need to either be able to configure MaxRequestSize (to simulate a smaller limit) or use a fixture with >10MB requests. 

This was the easiest way I could find to pass a request size, but we could do it in ways that don't affect the collector config. Either way though, it's going to involve exporting something around this request size publicly since it's in a different package than the integration test

damemi:
@dashpole I pushed a new commit that uses an internal package so we don't need to expose the sizing config to users. It's a little more wiring in the tests but otherwise pretty drop-in.

I did still have to add an exported function to the logs exporter, but it only takes an internal struct as its param so it's still inaccessible by users.

exporter/collector/internal/logsutil/logsutil.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2023 OpenTelemetry Authors
```
dashpole, line 1: is this the right header for this repo?

exporter/collector/integrationtest/inmemoryocexporter.go
```diff
@@ -188,6 +190,10 @@ func NewLogTestExporter(
 		logger,
 	)
 	require.NoError(t, err)
+
+	if extraConfig != nil {
```
dashpole, line 194: nit: ConfigureExporter already checks for a nil config, so this check isn't strictly necessary.  Same comment in logs_integration_test.go

exporter/collector/config.go
```diff
@@ -46,8 +46,8 @@ type Config struct {
 	ProjectID               string            `mapstructure:"project"`
 	UserAgent               string            `mapstructure:"user_agent"`
 	ImpersonateConfig       ImpersonateConfig `mapstructure:"impersonate"`
-	LogConfig               LogConfig         `mapstructure:"log"`
 	TraceConfig             TraceConfig       `mapstructure:"trace"`
+	LogConfig               LogConfig         `mapstructure:"log"`
```
dashpole, line 50: any reason why this was moved?  Technically, moving within a struct is a breaking change in go.

exporter/collector/internal/logsutil/logsutil.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logsutil
+
+type ExporterConfig struct {
```
dashpole, line 17: Even though this is internal, it might be helpful to have comments explaining what they are used for.

dashpole (commented):
just a few nits

exporter/collector/config.go
damemi: fieldalignment wanted this moved when I added the new fields. Guess it didn't care once I removed them. moved it back

exporter/collector/integrationtest/inmemoryocexporter.go
damemi: good point

exporter/collector/internal/logsutil/logsutil.go
damemi: good catch, I actually copied this from another file in our repo so we may want to do an audit. looks like this is in a lot of places

exporter/collector/internal/logsutil/logsutil.go
damemi: done

dashpole (approved):

PR #601, opened 2023-02-24
Bump golang.org/x/text from 0.3.7 to 0.3.8 in /tools

damemi:
dependabot threw a tantrum in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/597 and stopped responding, so this makes the same change manually


psx95:
There seems to be a minor reduction in codecov, I think because number of lines grew. The PR seems fine to me though.

psx95 (approved):
Not sure **exactly** why the codecov reduced a little.

damemi:
> Not sure **exactly** why the codecov reduced a little.

codecov seems to dip a little on `datapointcache.go` in every PR. I haven't figured out why, might be worth looking into because it's annoying

PR #613, opened 2023-03-14
Only attach instrumentation scope/service labels when non-empty

### LINKED ISSUE #377: Only attach instrumentation scope when it is non-empty
Some receivers don't set instrumentation scope name/version, which produces telemetry with a bunch of empty labels.  It might be a better UX to omit it in that case, even if the exporter has instrumentation labels enabled.  The same point also goes for service labels.

damemi:
Fixes #377 

This updates the metrics and traces exporters (for SDK and collector) to only attach instrumentation library labels to GCP if those labels are non-empty as otel attributes. This applies even if the option to include instrumentation library labels is set to `true`.

This also updates the metrics, traces, and logs exporters to do the same for service labels (also even if the service labels option is set to true).

---

This PR also makes updates to tests to prevent 2 tests overwriting each others' fixtures.

tl;dr I spent way too much time digging through Go internals, proto marshalling, and gzip libraries before I realized that the gzip test added in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/576 overwrote another test's fixture. This prevented me from reproducing the bug's behavior. Now `make fixtures` will fail if it sees 2 tests writing to the same output fixture.


exporter/collector/integrationtest/cmd/recordfixtures/main.go
```diff
@@ -54,12 +54,23 @@ func main() {
 	startTime := endTime.Add(-time.Second)
 	t := &FakeTesting{}
 
-	recordLogs(ctx, t, endTime)
-	recordMetrics(ctx, t, startTime, endTime)
-	recordTraces(ctx, t, startTime, endTime)
+	// track the output fixtures to catch tests overwriting another test's fixtures
+	outputFixtures := make(map[string]string)
```
aabmass, line 58: nit might be cleaner to do this check in a test e.g. add a new `exporter/collector/integrationtest/testcases/test_testcases.go`

exporter/collector/integrationtest/cmd/recordfixtures/main.go
```diff
@@ -54,12 +54,23 @@ func main() {
 	startTime := endTime.Add(-time.Second)
 	t := &FakeTesting{}
 
-	recordLogs(ctx, t, endTime)
-	recordMetrics(ctx, t, startTime, endTime)
-	recordTraces(ctx, t, startTime, endTime)
+	// track the output fixtures to catch tests overwriting another test's fixtures
```
punya, line 57: Could the fixture path be generated deterministically from the test name instead (so that collisions would be impossible)?

If we need to maintain a mapping, perhaps it would be more readable to define a type for the fixture tracker. One reason is that I usually expect functions to modify their receiver, but not to modify parameters that are passed in.

```go
type fixtureTracker map[string]string
func (ft fixtureTracker) register(test testcases.TestCase) error {...}
func (ft fixtureTracker) recordTraces(ctx context.Context, t *FakeTesting, startTime, endTime time.Time) {...}
```
```

File: exporter/collector/integrationtest/testcases/testcases_metrics.go [hunk 3]
```diff
@@ -221,7 +224,7 @@ var MetricsTestCases = []TestCase{
 	{
 		Name:                 "Basic Counter with gzip compression",
 		OTLPInputFixturePath: "testdata/fixtures/metrics/basic_counter_metrics.json",
-		ExpectFixturePath:    "testdata/fixtures/metrics/basic_counter_metrics_expect.json",
+		ExpectFixturePath:    "testdata/fixtures/metrics/basic_counter_metrics_compressed_expect.json",
```
aabmass, line 227: 😭 nice find

exporter/collector/metrics.go
```diff
@@ -388,13 +388,19 @@ func (me *MetricsExporter) createServiceTimeSeries(ctx context.Context, req *mon
...
-	return labels{
-		"instrumentation_source":  sanitizeUTF8(is.Name()),
-		"instrumentation_version": sanitizeUTF8(is.Version()),
+	instrumentationSource := sanitizeUTF8(is.Name())
+	if len(instrumentationSource) > 0 {
+		isLabels["instrumentation_source"] = instrumentationSource
 	}
+	instrumentationVersion := sanitizeUTF8(is.Version())
+	if len(instrumentationVersion) > 0 {
+		isLabels["instrumentation_version"] = instrumentationSource
```
punya:
```suggestion
		isLabels["instrumentation_version"] = instrumentationVersion
```

exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_access_batches_expected.json
```diff
@@ -397,25 +447,13 @@
               "zone": ""
             }
           },
-          "textPayload": "127.0.0.2 - - [26/Apr/2022:22:54:38 +0800] \"GET / HTTP/1.1\" 200 4429",
+          "textPayload": "127.0.0.1 - - [26/Apr/2022:22:54:43 +0800] \"-\" 408 -",
```
aabmass, line 450: Is this diff because the new service labels changed the cutoff for where the batches get split by bytes?

exporter/collector/spandata.go
```diff
@@ -67,30 +68,38 @@ func pdataSpanToOTSpanData(
...
+		endTime:                endTime,
+		name:                   span.Name(),
+		attributes:             pdataAttributesToOTAttributes(span.Attributes(), resource),
+		links:                  pdataLinksToOTLinks(span.Links()),
+		events:                 pdataEventsToOTMessageEvents(span.Events()),
+		droppedAttributes:      int(span.DroppedAttributesCount()),
+		droppedMessageEvents:   int(span.DroppedEventsCount()),
+		droppedLinks:           int(span.DroppedLinksCount()),
+		resource:               r,
+		instrumentationLibrary: instrumentationScopeLabels(is),
```
punya, line 84: Consider renaming this field to `instrumentationScope` in a separate PR.

exporter/collector/spandata.go
```diff
@@ -125,6 +134,13 @@ func pdataAttributesToOTAttributes(attrs pcommon.Map, resource pcommon.Resource)
 	otAttrs := make([]attribute.KeyValue, 0, attrs.Len())
 	appendAttrs := func(m pcommon.Map) {
 		m.Range(func(k string, v pcommon.Value) bool {
+			if k == semconv.AttributeServiceName ||
+				k == semconv.AttributeServiceNamespace ||
+				k == semconv.AttributeServiceInstanceID {
+				if len(v.AsString()) == 0 {
```
aabmass, line 140: nit can this be moved into the if expression?

exporter/collector/integrationtest/cmd/recordfixtures/main.go
damemi: Keeping this here throws an error and stops fixture generation at the source of the problem. Eg, I run `make fixtures` and immediately see that one of my fixtures is badly copy-pasted because the command fails.

If we move this to a separate test, that decouples this warning from what it's trying to warn about. imo this is more of a runtime error in the `recordfixtures` cmd that should stop execution.

exporter/collector/integrationtest/cmd/recordfixtures/main.go
damemi: sgtm, makes it easier to add other checks too if we need to

exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_access_batches_expected.json
damemi: Yeah good catch. I updated the batch size for this fixture and that looks like it got rid of the changes in a new commit https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/613/commits/3bb133f13c6df313788c88948cb3de7f2db8106e

The change against main still looks a little weird but I'm pretty sure that's just how git represented each entry getting shifted down a line: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/613/files#diff-55654849941a179a128e04a60faef5b295c8b07a7acd771f12b3c5dc3265a77e

exporter/collector/integrationtest/cmd/recordfixtures/main.go
```diff
@@ -48,18 +48,31 @@ func (t *FakeTesting) FailNow() {
...
 	ctx := context.Background()
 	endTime := time.Now()
 	startTime := endTime.Add(-time.Second)
 	t := &FakeTesting{}
 
-	recordLogs(ctx, t, endTime)
-	recordMetrics(ctx, t, startTime, endTime)
-	recordTraces(ctx, t, startTime, endTime)
+	// track the output fixtures to catch tests overwriting another test's fixtures
+	ft := make(fixtureRecorder)
```
aabmass, line 68: nit based on initials

```suggestion
	fr := make(fixtureRecorder)
```

aabmass (approved):

exporter/collector/integrationtest/cmd/recordfixtures/main.go
damemi: Yeah, I don't really feel like renaming all of our fixtures in this PR especially since the scope of the changes here are already beyond the original goal. I'll open another issue to track that

damemi: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/615

Issue #575, opened 2023-01-19
Add Batch Limit for Stackdriver Exporter


marvinkite:
On high-velocity metrics, you get:

```
InvalidArgument - Field timeSeries had an invalid value: A maximum of 200 TimeSeries can be written in a single request.
```

Apparently Stackdriver backend has the limit that each request cannot batch more than 200 time series so exporter needs to do some processing.

I have this issue with f0e0c76a62e3 commit

```golang 
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.34.3-0.20230106160205-f0e0c76a62e3
```

See GoogleCloudPlatform/opentelemetry-operations-js/issues/87

dashpole:
Thanks for reporting this!  This is definitely something we need to implement.  It exists on the googlecloud collector exporter for metrics, but not on the SDK exporter right now.

khongSC:
Would it be possible to get this ticket prioritized? I currently can't use the latest version of the exporter in production because it will constantly throw the maximum time series error. Thanks!

dashpole:
@damemi go for it.

gagata:
hi @damemi was there any progress on this?

knightpp:
A naive implementation could be <https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/compare/main...knightpp:opentelemetry-operations-go:issue-575>.

zchee:
@knightpp cc: @dashpole 
Could you create new PR based above change? My company team has same issue.
If can't send PR with any reason, I'll create instead, based by your changes with `Co-Author` mark.

knightpp:
@zchee I've created a PR. I hope the PR is good enough to be merged. Thanks.

damemi:
Fixed by #617, thanks @knightpp for contributing that!

PR #617, opened 2023-03-20
Use batches to call CreateTimeSeries

knightpp:
Based on https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/575 I've updated sending logic to batch time series data. The argument is hardcoded, and I couldn't find mocks for `*monitoring.MetricClient` so no tests.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -223,29 +223,48 @@ func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md
 func (me *metricExporter) exportTimeSeries(ctx context.Context, rm metricdata.ResourceMetrics) error {
 	tss := []*monitoringpb.TimeSeries{}
 	mr := me.resourceToMonitoredResourcepb(rm.Resource)
-	var aggError error
 
+	var aggErrors []error
 	extraLabels := me.extraLabelsFromResource(rm.Resource)
 	for _, scope := range rm.ScopeMetrics {
 		for _, metrics := range scope.Metrics {
 			ts, err := me.recordToTspb(metrics, mr, scope.Scope, extraLabels)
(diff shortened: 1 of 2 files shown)


knightpp:
> just need to you to sign the CLA (see the bot comment)
Hi, I've signed the CLA. There's a checkmark on the PR. Thanks.

damemi:
Ah sorry, I see it now.
/gcbrun

exporter/metric/metric.go
```diff
@@ -223,29 +223,48 @@ func (me *metricExporter) createMetricDescriptorIfNeeded(ctx context.Context, md
...
 	}
 
-	// TODO: When this exporter is rewritten, support writing to multiple
-	// projects based on the "gcp.project.id" resource.
-	req := &monitoringpb.CreateTimeSeriesRequest{
-		Name:       fmt.Sprintf("projects/%s", me.o.projectID),
-		TimeSeries: tss,
+	name := fmt.Sprintf("projects/%s", me.o.projectID)
+
+	const batchSize = 200
```
damemi, line 246: could you define this at the top-level of the file with a comment explaining it, like in the collector exporter https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/e9bb597e4aad734e5fa68ac8847e7000303368de/exporter/collector/metrics.go#L104-L108

damemi (commented):
looks good, one nit and can you add a test too? The test might be easier if you make `batchSize` a configurable (private) field on the exporter, eg so you can run a test with a max batch size of like 1 or 2 instead of having to create 200+ test metrics

knightpp:
> can you add a test too? 

Added tests and refactored `exportTimeSeries`.

PR #618, opened 2023-03-22
Use go workspaces instead of maintaining replace directives

punya:
https://go.dev/doc/tutorial/workspaces


aabmass:
Why was this abandoned?

punya:
I realized I didn't understand the semantics of `go work tidy` and I didn't have time to dig into it further. I still think the work is worth doing.

PR #620, opened 2023-04-05
Update OTEL dependencies to latest release candidates

lrascao:
Relevant changes required from RC [changelog](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.15.0-rc.2):

 * Both the Histogram and HistogramDataPoint are redefined with a generic argument of [N int64 | float64] in go.opentelemetry.io/otel/sdk/metric/metricdata. (https://github.com/open-telemetry/opentelemetry-go/pull/3849)
 * The metric Export interface from go.opentelemetry.io/otel/sdk/metric accepts a *ResourceMetrics instead of ResourceMetrics. (https://github.com/open-telemetry/opentelemetry-go/pull/3853)

Also of note `go.opentelemetry.io/otel/metric` is now stable as of [v1.15.0-rc2](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.15.0-rc.1)

fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/620


exporter/metric/metric.go
```diff
@@ -606,7 +606,7 @@ func toNonemptyTimeIntervalpb(start, end time.Time) (*monitoringpb.TimeInterval,
 	}, nil
 }
 
-func histToTypedValue(hist metricdata.HistogramDataPoint) *monitoringpb.TypedValue {
+func histToTypedValue(hist metricdata.HistogramDataPoint[float64]) *monitoringpb.TypedValue {
 	counts := make([]int64, len(hist.BucketCounts))
 	for i, v := range hist.BucketCounts {
 		counts[i] = int64(v)
```
lrascao, line 612: hm.. now i'm wondering if this should be `int64` instead

lrascao:
> If you don't mind waiting and updating this PR once these tags are published, that would be really helpful.

sure, i'll move this to draft until then

damemi:
@ptxmac thanks for bumping this, now that 1.15 is stable we should merge it

@lrascao can you resolve the conflicts and switch this from draft to ready for review?

lrascao:
Looks like we need to wait for folks at https://github.com/open-telemetry/opentelemetry-collector to cut a new release that also bumps to `v1.15.0`

lrascao:
Apparently there's a regression in the google managed prometheus metric naming convention:
```
--- FAIL: TestGetMetricName (0.00s)
    --- FAIL: TestGetMetricName/sum_without_total (0.00s)
        naming_test.go:143: MetricName(foo, {orig:0xc0001d5080})=foo_total/counter; want foo/counter
    --- FAIL: TestGetMetricName/sum_with_unit (0.00s)
        naming_test.go:143: MetricName(foo, {orig:0xc0001d5100})=foo_seconds_total/counter; want foo_total/counter
FAIL
FAIL    github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/googlemanagedprometheus   0.239s
```

damemi:
@lrascao is this still a draft or is it ready for review? Not sure if you were working on the regression you mentioned in your last comment

damemi:
> Apparently there's a regression in the google managed prometheus metric naming convention:
> 
> ```
> --- FAIL: TestGetMetricName (0.00s)
>     --- FAIL: TestGetMetricName/sum_without_total (0.00s)
>         naming_test.go:143: MetricName(foo, {orig:0xc0001d5080})=foo_total/counter; want foo/counter
>     --- FAIL: TestGetMetricName/sum_with_unit (0.00s)
>         naming_test.go:143: MetricName(foo, {orig:0xc0001d5100})=foo_seconds_total/counter; want foo_total/counter
> FAIL
> FAIL    github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/googlemanagedprometheus   0.239s
> ```

Digging around it looks like this is caused by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20519. Looking into it

lrascao:
> @lrascao is this still a draft or is it ready for review? Not sure if you were working on the regression you mentioned in your last comment

yeah, not really sure on that path forward here, adapt to the change or request rollback of the regression

damemi:
Talked to the team, and it looks like we should accept the change in default naming to match the Prometheus specification. In other words, just update the failing tests to expect the new names (with `_total` suffix).

This will be a breaking change, but it can be disabled using the feature gate flag from the upstream change (`--feature-gates=-pkg.translator.prometheus.NormalizeName`).

lrascao:
> Talked to the team, and it looks like we should accept the change in default naming to match the Prometheus specification. In other words, just update the failing tests to expect the new names (with `_total` suffix).
> 
> This will be a breaking change, but it can be disabled using the feature gate flag from the upstream change (`--feature-gates=-pkg.translator.prometheus.NormalizeName`).

alright, updating

lrascao:
ping @damemi ready for a GCB run

lrascao:
> /gcbrun

tests need a retry it seems

exporter/collector/googlemanagedprometheus/naming_test.go
```diff
@@ -45,16 +45,6 @@ func TestGetMetricName(t *testing.T) {
...
 	}{
-		{
-			desc:     "sum without total",
-			baseName: "foo",
-			metric: func(m pmetric.Metric) {
-				m.SetName("foo")
-				sum := m.SetEmptySum()
-				sum.SetIsMonotonic(true)
-			},
-			expected: "foo/counter",
```
damemi, line 56: I think we should keep this test, just update the expected to show that a sum without total will get _total appended

damemi (changes requested):
just want to add that test back

exporter/metric/metric.go
```diff
@@ -486,7 +486,7 @@ func (me *metricExporter) recordToTspb(m metricdata.Metrics, mr *monitoredrespb.
 			ts.Metric = me.recordToMpb(m, point.Attributes, library, extraLabels)
 			tss = append(tss, ts)
 		}
-	case metricdata.Histogram:
+	case metricdata.Histogram[float64]:
```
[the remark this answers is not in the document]
damemi: Yeah now that histogram uses generics we'll need to support both int64 and float64 (like we do for other metrics: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ceb4457c9caae3705d3f9f85b8fbf6c4c0e6f0e7/exporter/metric/metric.go#L322-L343)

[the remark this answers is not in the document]
lrascao: updated, not sure i've hit all call sites, apparently we don't have test coverage over this

exporter/collector/googlemanagedprometheus/naming_test.go
lrascao: renamed and added it back

ktong:
Could it upgrade to v1.16.0/v0.39.0 instead to avoid another upgrade? https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.16.0

damemi:
> Could it upgrade to v1.16.0/v0.39.0 instead to avoid another upgrade? https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.16.0

We are pretty close with this one, so given the number of changes I think it would be best to update incrementally if possible. So, getting this green then immediately following with another bump would be better imo.

Looking into why some of the unit tests are failing

exporter/metric/metric.go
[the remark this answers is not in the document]
damemi: I don't think this switch case falls through like this in go, explicitly adding the 2 cases fixed the unit test failure for me locally:

```go
	case metricdata.Histogram[float64]:
		for _, point := range a.DataPoints {
			ts, err := histogramToTimeSeries(point, m, mr)
			if err != nil {
				aggErr = multierr.Append(aggErr, err)
				continue
			}
			ts.Metric = me.recordToMpb(m, point.Attributes, library, extraLabels)
			tss = append(tss, ts)
		}
	case metricdata.Histogram[int64]:
		for _, point := range a.DataPoints {
			ts, err := histogramToTimeSeries(point, m, mr)
			if err != nil {
				aggErr = multierr.Append(aggErr, err)
				continue
			}
			ts.Metric = me.recordToMpb(m, point.Attributes, library, extraLabels)
			tss = append(tss, ts)
		}
```

exporter/metric/metric.go
```diff
@@ -389,7 +393,7 @@ func recordToMdpbKindType(a metricdata.Aggregation) (googlemetricpb.MetricDescri
 			return googlemetricpb.MetricDescriptor_CUMULATIVE, googlemetricpb.MetricDescriptor_DOUBLE
 		}
 		return googlemetricpb.MetricDescriptor_GAUGE, googlemetricpb.MetricDescriptor_DOUBLE
-	case metricdata.Histogram:
+	case metricdata.Histogram[float64]:
```
damemi, line 396:
```suggestion
	case metricdata.Histogram[float64], metricdata.Histogram[int64]:
```

exporter/metric/metric.go
```diff
@@ -606,20 +611,20 @@ func toNonemptyTimeIntervalpb(start, end time.Time) (*monitoringpb.TimeInterval,
...
 	counts := make([]int64, len(hist.BucketCounts))
 	for i, v := range hist.BucketCounts {
 		counts[i] = int64(v)
 	}
-	var mean float64
-	if !math.IsNaN(hist.Sum) && hist.Count > 0 { // Avoid divide-by-zero
-		mean = hist.Sum / float64(hist.Count)
+	var mean N
+	if !math.IsNaN(float64(hist.Sum)) && hist.Count > 0 { // Avoid divide-by-zero
+		mean = hist.Sum / N(hist.Count)
```
damemi, line 621: maybe we should keep these as forcing `float64`. if this code is modified to divide by an integer it will change the output

damemi (changes requested):

exporter/metric/metric.go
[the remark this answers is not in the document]
ptxmac: Ohh, interesting!
I guess duplicating the case is the best solution then

exporter/metric/metric.go
lrascao: that was my first thought, unfortunately doesn't work:

`../../metric/metric.go:494:27: a.DataPoints undefined (type "go.opentelemetry.io/otel/sdk/metric/metricdata".Aggregation has no field or method DataPoints)`

which kinda makes sense since `DataPoints []HistogramDataPoint[N]` is also generic

lrascao: going with the ugly dupe

exporter/collector/integrationtest/metrics_test.go
```diff
@@ -203,10 +204,11 @@ func TestSDKMetrics(t *testing.T) {
 				expectFixture,
 			)
 			if diff != "" {
+				fmt.Printf("%s", diff)
 				require.Fail(
 					t,
 					"Expected requests fixture and actual GCM requests differ",
-					diff,
+					"",
```
damemi, line 211: leftover debugging?

exporter/metric/metric.go
damemi: resolved the one below to clear it up

damemi (changes requested):

lrascao:
@damemi needs a splash of `/gcbrun`

exporter/collector/integrationtest/testcases/conversion.go
```diff
@@ -152,13 +152,13 @@ func convertHistogram(h pmetric.Histogram) metricdata.Aggregation {
 	if h.DataPoints().Len() == 0 {
 		return nil
 	}
-	agg := metricdata.Histogram{
+	agg := metricdata.Histogram[float64]{
```
damemi, line 155: As a note, I think it is fine that we are only handling float64 histograms here, because 
1. this code is just handling testcases.
2. `Sum` is the only value used here that can be int or float in `metricdata.Histogram[N]` (we don't read [Min or Max](https://github.com/open-telemetry/opentelemetry-go/blob/e0852d609c4a4205d550e2de45afdbf80d43557b/sdk/metric/metricdata/data.go#L130-L134)) and `Sum()` always returns a float64, so there shouldn't be any runtime issues with this.
3. We [force `float64`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/620/files#diff-0a6b748481b08d3c4a7c5db8799a3b134f70e26fa97566ec8e0c4543d81783e7L615-R630) when we handle Sum anyway, which is consistent with the behavior prior to this update and therefore not breaking.

However, we should follow this up with support for `int64` histograms. Otherwise, users who switch to the GCP exporter might see slightly different values since we force floats in the output calculation.  That would include adding a test case for an int64 histogram, which we don't appear to have currently. Will make a separate issue to track this.

damemi (approved):
Thanks for all of your help @lrascao!

Issue #614, opened 2023-03-16
Authenticating with `gcloud auth login` on GCE does not seem possible


braydonk:
The docs for the [Google Cloud Exporter in opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter#google-cloud-exporter) suggest that I get credentials through the command `gcloud auth application-default login`. When I run `gcloud auth application-default login` on a GCE VM, I get the following JSON shape:
```
{
  "client_id": "redacted",
  "client_secret": "redacted",
  "quota_project_id": "my-project-id",
  "refresh_token": "redacted",
  "type": "authorized_user"
}
```
This causes the exporter to fail with the following error:
```
Error: failed to build pipelines: failed to create "googlecloud" exporter, in pipeline "metrics/*": no project found with application default credentials
```
This is because the creds found by `google.FindDefaultCredentials` don't have a field `ProjectID`. https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/16f0ea8ce52bbbe71d4cbac84e5317c83d7324b6/exporter/metric/cloudmonitoring.go#L38-L47

I am unsure if this is a GCE specific problem. If it isn't, then the documentation should be revisited. However, it is very possible that this is a GCE specific problem and perhaps a note can be added to the documentation if anything.

damemi:
My last comment was a lot of rambling. I opened https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/621 to update our docs with better recommendations for authenticating based on environment. tl;dr we shouldn't be telling people to use `gcloud auth application-default login` except for personal testing. If this looks good I'll add the same change to the upstream docs too.

Longer term, we will need to look more at our auth story (specifically our dependency on project id) and make sure it functionally aligns with the client libraries. For now though, I think this is WAI, just poorly documented.

PR #621, opened 2023-04-07
Update auth docs for Collector exporter

damemi:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/614
(for now)

This adds more detail around the different ways to authenticate the collector for GCP, de-emphasizing the use of `gcloud auth application-default login` and instead recommending the GCP service account.

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -4,22 +4,73 @@ This exporter can be used to send metrics and traces to Google Cloud Monitoring 
 
 ## Getting started
 
+In general, authenticating with the Collector exporter follows the same steps as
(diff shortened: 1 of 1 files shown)


punya:
Consider referencing https://cloud.google.com/docs/authentication/provide-credentials-adc as a canonical doc before summarizing the bits that are most relevant to our use case.

damemi:
> Consider referencing https://cloud.google.com/docs/authentication/provide-credentials-adc as a canonical doc before summarizing the bits that are most relevant to our use case.

sg, added a link at the top

punya (approved):

PR #622, opened 2023-04-07
Fix GMP metric name handling

quentinmit:
If OTLP metric data comes in with a URL-style metric name, the `googlemanagedprometheus` exporter will sanitize the name but forget to prepend the correct `prometheus.googleapis.com/` prefix.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus.json
@@ -457,6 +457,19 @@
                            }
                         ]
                      }
+                  },
+                  {
+                     "name":"prometheus.googleapis.com/otlp.test.prefix6",
+                     "description":"GMP metric names are not special",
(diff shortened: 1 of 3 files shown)


punya (approved):
Thanks Quentin!

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus.json
```diff
@@ -379,10 +379,88 @@
...
+                        "dataPoints":[
+                           {
+                              "timeUnixNano":"1649443516286000000",
+                              "asDouble":0.00699415
+                           }
+                        ]
+                     }
+                  },
+                  {
+                     "name":"WORKLOAD.googleapis.com/otlp.test.prefix5",
```
aabmass, line 449: Can you add a test for `prometheus.googleapis.com/foo` to make it explicit that it will not be stripped ?

aabmass (approved):

PR #626, opened 2023-04-19
add resource mappings for Anthos managed monitored resources

waterloong:


damemi:
@waterloong this MR isn't meant to be external right? If so it shouldn't go here

PR #631, opened 2023-05-23
Add write ahead log to collector metrics exporter

damemi:
This adds optional WAL buffering to the collector metrics exporter, heavily based on a similar implementation in the [Prometheus remote write exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6f1b6b97eb5d317bee3d20bcf1a8a96ceabdbd6b/exporter/prometheusremotewriteexporter/wal.go).

The flow for this is:
* If WAL is enabled, set it up in the user-specified directory under `gcp_metrics_wal`
* Update `PushMetrics()` to write byte-marshalled CreateTimeSeriesRequest objects to the WAL
    * if WAL is disabled, just export the requests as normal
* Start a `walRunner()` go routine that spawns a `readWALAndExport()` loop (the goroutine also handles shutdown signals)
* `readWALAndExport()` continuously reads the latest entry from the WAL and exports it to GCM:
    * If the export is successful, move the read index forward and continue (truncating the log file)
    * If the export fails for DeadlineExceeded or Unavailable, retry forever until success or non-retryable failure
    * If the export fails for any other reason, move the read index forward and continue (truncating the log file and skipping this entry)
* If there are no entries found in the WAL, `watchWAL()` opens a file watch to block for changes to the log file.

Added an integration test with WAL enabled to test this

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -180,8 +180,8 @@ Additional configuration for the metric exporter:
 - `metric.prefix` (optional): MetricPrefix overrides the prefix / namespace of the Google Cloud metric type identifier. If not set, defaults to "custom.googleapis.com/opencensus/"
 - `metric.skip_create_descriptor` (optional): Whether to skip creating the
   metric descriptor.
-- `metric.wal.directory` (optional): Path to local write-ahead-log file.
-- `metric.wal.max_backoff` (optional): Maximum duration to retry entries from
+- `metric.experimental_wal.experimental_directory` (optional): Path to local write-ahead-log file.
+- `metric.experimental_wal.experimental_max_backoff` (optional): Maximum duration to retry entries from
   the WAL on network errors.
 
 Addition configuration for the logging exporter:
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -178,7 +178,11 @@ Note: These `retry_on_failure` and `sending_queue` are provided (and documented)
 Additional configuration for the metric exporter:
 
 - `metric.prefix` (optional): MetricPrefix overrides the prefix / namespace of the Google Cloud metric type identifier. If not set, defaults to "custom.googleapis.com/opencensus/"
-- `metric.skip_create_descriptor` (optional): Whether to skip creating the metric descriptor.
+- `metric.skip_create_descriptor` (optional): Whether to skip creating the
+  metric descriptor.
+- `metric.wal.directory` (optional): Path to local write-ahead-log file.
+- `metric.wal.max_backoff` (optional): Maximum duration to retry entries from
+  the WAL on network errors.
 
 Addition configuration for the logging exporter:
 
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -148,9 +148,9 @@ type MetricConfig struct {
 // it preserves both the data on disk and the order of the data points.
 type WALConfig struct {
 	// Directory is the location to store WAL files.
-	Directory string `mapstructure:"experimental_directory"`
+	Directory string `mapstructure:"directory"`
 	// MaxBackoff sets the length of time to exponentially re-try failed exports.
-	MaxBackoff time.Duration `mapstructure:"experimental_max_backoff"`
+	MaxBackoff time.Duration `mapstructure:"emax_backoff"`
 }
 
 // ImpersonateConfig defines configuration for service account impersonation.
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -46,8 +46,8 @@ type Config struct {
 	ProjectID               string            `mapstructure:"project"`
 	UserAgent               string            `mapstructure:"user_agent"`
 	ImpersonateConfig       ImpersonateConfig `mapstructure:"impersonate"`
-	LogConfig               LogConfig         `mapstructure:"log"`
 	TraceConfig             TraceConfig       `mapstructure:"trace"`
+	LogConfig               LogConfig         `mapstructure:"log"`
 	MetricConfig            MetricConfig      `mapstructure:"metric"`
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
@@ -90,11 +90,6 @@ type AttributeMapping struct {
 }
 
 type MetricConfig struct {
-	// GetMetricName is not settable in config files, but can be used by other
-	// exporters which extend the functionality of this exporter. It allows
-	// customizing the naming of metrics. baseName already includes type
-	// suffixes for summary metrics, but does not (yet) include the domain prefix
-	GetMetricName func(baseName string, metric pmetric.Metric) (string, error)
 	// MapMonitoredResource is not exposed as an option in the configuration, but
 	// can be used by other exporters to extend the functionality of this
 	// exporter. It allows overriding the function used to map otel resource to
@@ -103,7 +98,14 @@ type MetricConfig struct {
 	// ExtraMetrics is an extension point for exporters to add to the set
 	// of ResourceMetrics during a call to PushMetrics.
 	ExtraMetrics func(pmetric.Metrics) pmetric.ResourceMetricsSlice
-	Prefix       string `mapstructure:"prefix"`
+	// GetMetricName is not settable in config files, but can be used by other
+	// exporters which extend the functionality of this exporter. It allows
+	// customizing the naming of metrics. baseName already includes type
+	// suffixes for summary metrics, but does not (yet) include the domain prefix
+	GetMetricName func(baseName string, metric pmetric.Metric) (string, error)
+	// WALConfig holds configuration settings for the write ahead log.
+	WALConfig *WALConfig `mapstructure:"wal_config"`
+	Prefix    string     `mapstructure:"prefix"`
 	// KnownDomains contains a list of prefixes. If a metric already has one
 	// of these prefixes, the prefix is not added.
 	KnownDomains []string `mapstructure:"known_domains"`
@@ -138,6 +140,17 @@ type MetricConfig struct {
 	// deviation.  It isn't correct, so we don't send it by default, and don't expose
 	// it to users. For some uses, it is expected, however.
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
+}
+
+// WALConfig defines settings for the write ahead log. WAL buffering writes data
+// points in-order to disk before reading and exporting them. This allows for
+// better retry logic when exporting fails (such as a network outage), because
+// it preserves both the data on disk and the order of the data points.
+type WALConfig struct {
+	// Directory is the location to store WAL files.
+	Directory string `mapstructure:"directory"`
+	// MaxBackoff sets the length of time to exponentially re-try failed exports.
+	MaxBackoff time.Duration `mapstructure:"max_backoff"`
 }
 
 // ImpersonateConfig defines configuration for service account impersonation.
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -104,7 +104,7 @@ type MetricConfig struct {
 	// suffixes for summary metrics, but does not (yet) include the domain prefix
 	GetMetricName func(baseName string, metric pmetric.Metric) (string, error)
 	// WALConfig holds configuration settings for the write ahead log.
-	WALConfig *WALConfig `mapstructure:"wal_config"`
+	WALConfig *WALConfig `mapstructure:"experimental_wal_config"`
 	Prefix    string     `mapstructure:"prefix"`
 	// KnownDomains contains a list of prefixes. If a metric already has one
 	// of these prefixes, the prefix is not added.
@@ -148,9 +148,9 @@ type MetricConfig struct {
 // it preserves both the data on disk and the order of the data points.
 type WALConfig struct {
 	// Directory is the location to store WAL files.
-	Directory string `mapstructure:"directory"`
+	Directory string `mapstructure:"experimental_directory"`
 	// MaxBackoff sets the length of time to exponentially re-try failed exports.
-	MaxBackoff time.Duration `mapstructure:"max_backoff"`
+	MaxBackoff time.Duration `mapstructure:"experimental_max_backoff"`
 }
 
 // ImpersonateConfig defines configuration for service account impersonation.
--- a/exporter/collector/integrationtest/inmemoryocexporter.go
+++ b/exporter/collector/integrationtest/inmemoryocexporter.go
@@ -162,7 +162,6 @@ func NewMetricTestExporter(
 		ctx,
 		cfg,
 		zap.NewNop(),
-		//zap.NewExample(),
 		"latest",
 		collector.DefaultTimeout,
 	)
--- a/exporter/collector/integrationtest/inmemoryocexporter.go
+++ b/exporter/collector/integrationtest/inmemoryocexporter.go
@@ -162,6 +162,7 @@ func NewMetricTestExporter(
 		ctx,
 		cfg,
 		zap.NewNop(),
+		//zap.NewExample(),
 		"latest",
 		collector.DefaultTimeout,
 	)
--- a/exporter/collector/integrationtest/testcases/testcases_metrics.go
+++ b/exporter/collector/integrationtest/testcases/testcases_metrics.go
@@ -15,7 +15,9 @@
 package testcases
 
 import (
+	"os"
 	"strings"
+	"time"
 
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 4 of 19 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ Thumbs.db
 coverage.*
 
 /example/metric/metric
+/exporter/collector/integrationtest/gcp_metrics_wal/
```
psx95, line 13: Nit: Missing newline.

exporter/collector/config.go
```diff
@@ -140,6 +142,18 @@ type MetricConfig struct {
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
 }
 
+// WALConfig defines settings for the write ahead log. WAL buffering writes data
+// points in-order to disk before reading and exporting them. This allows for
```
jsuereth, line 146: Is there a way to do in-memory only WAL?

exporter/collector/metrics.go
psx95, line 1: A few newly added lines here are not covered by unit test - some of them involve the specific case when WAL is enabled. Would it be worth adding tests that exercise that logic ?

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			}
+			// retry at same read index if retryable (network) error
+			s := status.Convert(err)
+			if s.Code() == codes.DeadlineExceeded || s.Code() == codes.Unavailable {
+				me.obs.log.Error("non-retryable error, skipping request")
+				continue
 			}
+
+			err = me.wal.TruncateFront(me.rWALIndex.Load())
 			if err != nil {
```
jsuereth, line 428: Are you testing `err` to see if it's non-retryable here?

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			switch event.Op {
+			case fsnotify.Remove:
+				me.obs.log.Error("WAL file deleted")
+			case fsnotify.Rename:
+				me.obs.log.Error("WAL file renamed")
+			case fsnotify.Write:
+				wErr = nil
+			}
+
+		case eerr, ok := <-walWatcher.Errors:
```
psx95, line 492: Possible typo - `eerr` ?

exporter/collector/config.go
damemi: Not with this library: https://github.com/tidwall/wal/blob/e70ddec4a12f90db734d4d90df98b5088bd6d417/wal.go#L159

exporter/collector/metrics.go
damemi: Retryable errors from GCM are handled in the `continue` statement above (L423). If this line returns an error then it was a problem with the WAL itself

exporter/collector/metrics.go
damemi: I'll add some unit tests, but just as a note the codecov checks don't know about our integration (fixture) tests, which do cover these lines

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			if ok {
+				wErr = eerr
+			}
+		}
+	}()
+	err = <-watchCh
+	return err
+}
+
+func (me *MetricsExporter) walRunner(ctx context.Context) {
```
psx95, line 502: A comment describing how this function fits in the bigger picture would be helpful to anyone contributing to this file.

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+		select {
+		case <-me.shutdownC:
+			return
+		case <-ctx.Done():
+			wErr = ctx.Err()
+			return
+
+		case event, ok := <-walWatcher.Events:
+			if !ok {
+				return
```
psx95, line 481: What does `!ok` indicate here ?

exporter/collector/integrationtest/testdata/fixtures/metrics/basic_counter_metrics_wal_expect.json
psx95, line 1: The [basic_counter_metrics_expect.json](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/0179cdda02d5c1bb256fa60087883328573f4672/exporter/collector/integrationtest/testdata/fixtures/metrics/basic_counter_metrics_expect.json) has contents differing to this output for the same input. 

No of lines in basic_counter_metrics_expect.json = 693
No of lines in basic_counter_metrics_wal_expect.json = 360

Is this expected ?

damemi: That's a good catch, these should be the same. It looks like the test metrics aren't actually being sent to the fixture in this change (only selfObservabilityMetrics). Something's definitely not right with that

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+		// Must have been ErrNotFound, start a file watch and block waiting for updates.
+		if err := me.watchWAL(ctx); err != nil {
+			return err
 		}
 	}
-	if len(errs) > 0 {
-		return multierr.Combine(errs...)
+}
+
+// watchWAL watches the WAL directory for a write then returns to the
```
punya, line 451: Is this approach commonly used? For example, I don't see it used (at least directly) in the WAL implementation for the PRW exporter.

This would make sense if WAL writes were coming from a different process, but is there a simpler way where we write to a channel after every successful write to the WAL?

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+	return err
+}
+
+func (me *MetricsExporter) walRunner(ctx context.Context) {
+	defer me.goroutines.Done()
+	for {
+		select {
+		case <-me.shutdownC:
+			// do one last final sync/read/export then return
+			err := me.wal.Sync()
```
punya, line 508: Why is this necessary, given that we've set up our WAL in auto-sync mode?

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+		}
+	}()
+	err = <-watchCh
+	return err
+}
+
+func (me *MetricsExporter) walRunner(ctx context.Context) {
+	defer me.goroutines.Done()
+	for {
+		select {
```
punya, line 505: Do we need to handle `ctx.Done()`?

exporter/collector/metrics.go
punya: Also consider giving the function a more descriptive name such as `readAndExportLoop`.

exporter/collector/metrics.go
```diff
@@ -64,14 +71,22 @@ type selfObservability struct {
...
 	// shutdownC is a channel for signaling a graceful shutdown
 	shutdownC chan struct{}
 	// mdCache tracks the metric descriptors that have already been sent to GCM
 	mdCache map[string]*monitoringpb.CreateMetricDescriptorRequest
+	// A channel that receives metric descriptor and sends them to GCM once
+	metricDescriptorC chan *monitoringpb.CreateMetricDescriptorRequest
+	// the "write" index for WAL
+	wWALIndex *atomic.Uint64
+	// the "read" index for WAL
+	rWALIndex *atomic.Uint64
```
punya, line 87: Does this actually need to be atomic, or could it be a local variable in the read and export loop? It seems like it's only updated in that one goroutine.

punya (commented):
Is there a better way to separate the WAL concern from the other parts of our export logic? For example, could you make a self-contained struct that deals with the WAL and has a small API that can be used by the exporter? This could take a couple of forms:

* a struct that has no-op implementations for the nil value (which would remove the repeated nil-checks in the exporter)
* a struct that wraps the "raw" exporter in a version that adds a WAL

exporter/collector/metrics.go
```diff
@@ -272,44 +329,198 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
-			if failedPoints > 0 {
-				recordPointCountDataPoint(ctx, failedPoints, st)
+			err = me.export(ctx, req)
+			if err != nil {
+				me.obs.log.Warn(fmt.Sprintf("error exporting to GCM: %+v", err))
+			}
+			// retry at same read index if retryable (network) error
+			s := status.Convert(err)
+			if s.Code() == codes.DeadlineExceeded || s.Code() == codes.Unavailable {
+				me.obs.log.Error("non-retryable error, skipping request")
```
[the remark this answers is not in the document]
punya: Why do we consider DeadlineExceeded and Unavailable to be unrecoverable errors? E.g., it seems like unavailable could be a transient network blip.

Conversely, if we encounter an ostensibly recoverable error but it happens over and over again - should we implement some form of backoff and eventually give up (drop that request)?

exporter/collector/metrics.go
```diff
@@ -191,13 +214,47 @@ func NewGoogleCloudMetricsExporter(
...
+func (me *MetricsExporter) setupWAL() error {
+	walPath := filepath.Join(me.cfg.MetricConfig.WALConfig.Directory, "gcp_metrics_wal")
+	me.walPath = walPath
+	metricWal, err := wal.Open(walPath, nil)
+	if err != nil {
+		return err
+	}
+	me.wal = metricWal
+
+	// sync existing WAL indices
```
punya, line 243: If there are a nonzero number of entries in the WAL at startup, does it mean we're recovering from a crash/unclean shutdown? In that case it seems useful to log how many entries we're about to try to recover.

(Crash recovery seems like the biggest benefit of file- vs memory-backed WAL.)

exporter/collector/metrics.go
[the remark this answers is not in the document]
punya: Maybe I'm misunderstanding the control flow, but this logic seems like it contradicts the PR description (i.e. retryable vs non-retryable seem flipped).

exporter/collector/config.go
```diff
@@ -103,7 +98,14 @@ type MetricConfig struct {
 	// ExtraMetrics is an extension point for exporters to add to the set
 	// of ResourceMetrics during a call to PushMetrics.
 	ExtraMetrics func(pmetric.Metrics) pmetric.ResourceMetricsSlice
-	Prefix       string `mapstructure:"prefix"`
+	// GetMetricName is not settable in config files, but can be used by other
```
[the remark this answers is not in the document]
damemi: these are caught by the `fieldalignment` filter, and won't pass CI without the formatting changes in this PR

exporter/collector/config.go
jsuereth: I think handling network outages is the priority.  5-min outage with even a potential collector crash/reboot would be handled by the on-disk portion here.

In-memory is a nice to have, unless the WAL is not viable for some reason.

exporter/collector/integrationtest/testdata/fixtures/metrics/basic_counter_metrics_wal_expect.json
damemi: added cross-reference option to testcases

damemi:
Sorry for the push spam, trying to debug some flakey CI test results that I can't reproduce as reliably locally or in a container

internal/cloudmock/metrics.go
```diff
@@ -104,10 +105,16 @@ func (f *fakeMetricServiceServer) CreateTimeSeries(
 	var code codes.Code
 	if strings.Contains(req.Name, "notfound") {
 		code = codes.NotFound
+	} else if strings.Contains(req.Name, "unavailable") {
```
jsuereth, line 108: Please add docs somewhere denoting these magic strings.

exporter/collector/config.go
```diff
@@ -140,6 +142,18 @@ type MetricConfig struct {
...
+// WALConfig defines settings for the write ahead log. WAL buffering writes data
+// points in-order to disk before reading and exporting them. This allows for
+// better retry logic when exporting fails (such as a network outage), because
+// it preserves both the data on disk and the order of the data points.
+type WALConfig struct {
+	// Directory is the location to store WAL files.
+	Directory string `mapstructure:"directory"`
+	// Enabled turns the WAL on or off. When false, the exporter will not use
+	// the WAL and will instead attempt to send data directly to Google Cloud.
+	Enabled bool `mapstructure:"enabled"`
```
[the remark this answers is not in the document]
damemi: changed this to a pointer and check for nil when instantiating

exporter/collector/metrics.go
```diff
@@ -116,6 +131,12 @@ func (me *MetricsExporter) Shutdown(ctx context.Context) error {
 	go func() {
 		// Wait until all goroutines are done
 		me.goroutines.Wait()
+		// Close the WAL if open
+		if me.wal != nil {
+			if err := me.wal.Close(); err != nil {
+				fmt.Printf("error closing WAL: %+v\n", err)
```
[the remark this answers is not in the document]
damemi: done

exporter/collector/metrics.go
damemi: this is intentionally `eerr` to avoid duplicating the `err` variable above ... renamed to be better

exporter/collector/metrics.go
damemi: removed these

exporter/collector/metrics.go
damemi: removed

exporter/collector/metrics.go
```diff
@@ -272,44 +360,252 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+			}
+			// ErrNotFound from wal.Read() means the index is either 0 or out of
+			// bounds (indicating we're probably at the end of the WAL). That error
+			// will trigger a file watch for new writes (below this). For other
+			// errors, fail.
+			if !errors.Is(err, wal.ErrNotFound) {
+				me.obs.log.Error(fmt.Sprintf("error reading WAL and exporting: %+v", err))
+			}
+
+			// Must have been ErrNotFound, start a file watch and block waiting for updates.
```
ridwanmsharif, line 603: The above `if` block doesn't break or return. Won't we get here even if `err` is another error?

exporter/collector/metrics.go
```diff
@@ -272,44 +360,252 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
-			st = statusCodeToString(s)
-
-			succeededPoints := len(ts)
-			failedPoints := 0
-			for _, detail := range s.Details() {
-				if summary, ok := detail.(*monitoringpb.CreateTimeSeriesSummary); ok {
-					failedPoints = int(summary.TotalPointCount - summary.SuccessPointCount)
-					succeededPoints = int(summary.SuccessPointCount)
-				}
+			if !(s.Code() == codes.DeadlineExceeded || s.Code() == codes.Unavailable) {
```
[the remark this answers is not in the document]
damemi: Yes, but also to keep the ordering of the requests

exporter/collector/config.go
```diff
@@ -140,6 +142,17 @@ type MetricConfig struct {
 	EnableSumOfSquaredDeviation bool `mapstructure:"sum_of_squared_deviation"`
 }
 
+// WALConfig defines settings for the write ahead log. WAL buffering writes data
+// points in-order to disk before reading and exporting them. This allows for
+// better retry logic when exporting fails (such as a network outage), because
+// it preserves both the data on disk and the order of the data points.
+type WALConfig struct {
```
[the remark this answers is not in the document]
damemi: Agreed, but I think that is fine as a follow up feature request

exporter/collector/metrics.go
```diff
@@ -191,15 +215,79 @@ func NewGoogleCloudMetricsExporter(
...
+	walPath := filepath.Join(me.cfg.MetricConfig.WALConfig.Directory, "gcp_metrics_wal")
+	me.wal.path = walPath
+	metricWal, err := wal.Open(walPath, &wal.Options{LogFormat: 1})
+	if err != nil {
+		return 0, 0, err
+	}
+	me.wal.Log = metricWal
+
+	// default to 1 hour exponential backoff
+	me.wal.maxBackoff = time.Duration(3600 * time.Second)
```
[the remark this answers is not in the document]
damemi: done

exporter/collector/metrics.go
damemi: The idea was just to log errors that aren't `ErrNotFound`, since not found means we're at the end of the WAL and can be expected sometimes. added a comment explaining this

internal/cloudmock/metrics.go
damemi: These are only used in this internal mock server for testing. But I added a comment for this function explaining that a lot better

exporter/collector/README.md
```diff
@@ -178,7 +178,11 @@ Note: These `retry_on_failure` and `sending_queue` are provided (and documented)
 Additional configuration for the metric exporter:
 
 - `metric.prefix` (optional): MetricPrefix overrides the prefix / namespace of the Google Cloud metric type identifier. If not set, defaults to "custom.googleapis.com/opencensus/"
-- `metric.skip_create_descriptor` (optional): Whether to skip creating the metric descriptor.
+- `metric.skip_create_descriptor` (optional): Whether to skip creating the
+  metric descriptor.
+- `metric.wal.directory` (optional): Path to local write-ahead-log file.
+- `metric.wal.max_backoff` (optional): Maximum duration to retry entries from
```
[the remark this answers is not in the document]
punya: I recall that in the collector, there's a convention of using an `experimental_` prefix to signal such things in configuration.

Do we think that this is going to be replaced with an enhanced version of `queued_retry` in future?

exporter/collector/config.go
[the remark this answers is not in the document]
punya: Filed #646

exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -229,5 +231,78 @@ var MetricsTestCases = []TestCase{
 			cfg.MetricConfig.ClientConfig.Compression = "gzip"
 		},
 	},
+	{
+		Name:                 "Write ahead log enabled",
+		OTLPInputFixturePath: "testdata/fixtures/metrics/basic_counter_metrics.json",
+		ExpectFixturePath:    "testdata/fixtures/metrics/basic_counter_metrics_wal_expect.json",
+		CompareFixturePath:   "testdata/fixtures/metrics/basic_counter_metrics_expect.json",
```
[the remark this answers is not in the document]
damemi: "expect" is what the input for a test converts into, and it verifies that a single test hasn't changed since its last fixture was checked in

"compare" is to cross-reference output fixtures between different tests that should have equal outcomes. In this case, we want to make sure that enabling WAL gives the same output as normal export, based on the same input fixture.

exporter/collector/integrationtest/inmemoryocexporter.go
```diff
@@ -162,6 +162,7 @@ func NewMetricTestExporter(
 		ctx,
 		cfg,
 		zap.NewNop(),
+		//zap.NewExample(),
```
punya, line 165:
```suggestion
```

exporter/collector/README.md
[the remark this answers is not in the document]
damemi: I'm fine with that, but our exporter is beta so I don't think we can change these settings once we publish them

exporter/collector/metrics.go
aabmass: It looks like this context is the one passed to `NewGoogleCloudMetricsExporter()`. Is that one intended for initialization only? We're not even passing it to `go mExp.exportMetricDescriptorRunner()` on L235.

exporter/collector/metrics.go
```diff
@@ -271,44 +365,254 @@ func (me *MetricsExporter) PushMetrics(ctx context.Context, m pmetric.Metrics) e
...
+		}
+	}()
+	err = <-watchCh
+	return err
+}
+
+func (me *MetricsExporter) runWALReadAndExportLoop(ctx context.Context) {
+	defer me.goroutines.Done()
+	runCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
aabmass, line 577: We're not using this cancel anywhere, I think this code is a noop

exporter/collector/metrics.go
```diff
@@ -115,6 +138,12 @@ func (me *MetricsExporter) Shutdown(ctx context.Context) error {
 	go func() {
 		// Wait until all goroutines are done
 		me.goroutines.Wait()
+		// Close the WAL if open
+		if me.wal != nil {
+			if err := me.wal.Close(); err != nil {
+				me.obs.log.Error(fmt.Sprintf("error closing WAL: %+v\n", err))
+			}
```
aabmass, line 145: Consider moving this into into a defer in `runWALReadAndExportLoop()` so the goroutine can handle its own cleanup when it receives the shutdown signal.

aabmass (approved):
I think it might be good to follow up PR with refactoring the WAL into its own struct as `metrics.go` is getting pretty long and I think the sprinkled WAL logic could be encapsulated

exporter/collector/metrics.go
```diff
@@ -182,6 +211,7 @@ func NewGoogleCloudMetricsExporter(
 		shutdownC:         shutdown,
 		timeout:           timeout,
 	}
+	mExp.exportFunc = mExp.exportToTimeSeries
```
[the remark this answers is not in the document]
damemi: `exportToTimeSeries()` is a method on a MetricsExporter struct, so I don't think I can set it in the initializer

exporter/collector/metrics.go
damemi: `shutdownC` is only closed if `Shutdown()` is called, so I think handling the main context closing is good to have to prevent from blocking on other terminations (non-gracefully?)

exporter/collector/metrics.go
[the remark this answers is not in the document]
aabmass: Ah that makes sense

exporter/collector/metrics.go
damemi: sgtm

exporter/collector/metrics.go
damemi: sounds good, removed this and changed the `runCtx` to just `ctx`

exporter/collector/README.md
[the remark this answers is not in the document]
damemi: ah yeah I commented without seeing Punya's update

I changed these to add the `experimental_` prefix, and also to the wal object... it looks a little weird like this:

```yaml
metric:
  experimental_wal:
    experimental_directory: my/path
```

do you think we could just make the `wal` subsection experimental and imply that everything under it is also subject to change? Or just to be safe, mark each option experimental?

```yaml
metric:
  experimental_wal:
    directory: my/path #subject to change
```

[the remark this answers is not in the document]
aabmass: I think the second less verbose option is fine. I see they are doing that [this processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7c0fd3de2aa7cbf61d6360c6208079c761992512/processor/metricsgenerationprocessor/README.md#configuration) as well.

exporter/collector/metrics.go
aabmass: > so I think handling the main context closing is good to have to prevent from blocking on other terminations (non-gracefully?)

That makes sense, but I wasn't sure of the collector's contract here. I figured it would eventually cancel that "initialization" context which would cause this goroutine to exit but that doesn't seem to be the case.

exporter/collector/integrationtest/testcases/testcases_metrics.go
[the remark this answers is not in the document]
damemi: there is also a comment in the definition in testcase.go:

```
 	// CompareFixturePath is a second output fixture that should be equal to this test's output fixture.
	// Used for cross-referencing multiple tests that should have the same output, without overwriting the same fixtures.
```

damemi:
I think I've addressed all the blockers, going to merge this now so we can get an initial release with it out.

PR #641, opened 2023-06-09
add support for Google Bare Metal Solution

alex-basinov:
Adds support for [BMS machines](https://cloud.google.com/bare-metal).

### DIFF
--- a/internal/resourcemapping/resourcemapping.go
+++ b/internal/resourcemapping/resourcemapping.go
@@ -52,6 +52,7 @@ const (
 	serviceName       = "service_name"
 	configurationName = "configuration_name"
 	revisionName      = "revision_name"
+	bmsInstance       = "baremetalsolution.googleapis.com/Instance"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


internal/resourcemapping/resourcemapping.go
```diff
@@ -122,6 +123,10 @@ var (
 			},
 			awsAccount: {otelKeys: []string{string(semconv.CloudAccountIDKey)}},
 		},
+		bmsInstance: {
+			location:   {otelKeys: []string{string(semconv.CloudRegionKey)}},
```
aabmass, line 127: Could location use the zone as well? Many others resources check for zone then region

aabmass (commented):
@alex-basinov is this monitored resource publicly writeable yet? If it is, would you mind adding an integration test for it following https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtests.md

Lmk if you have any questions.

alex-basinov:
> @alex-basinov is this monitored resource publicly writeable yet?

No, not yet but It'll become publicly writable very soon.

alex-basinov:
The monitored resource is now publicly writeable. I've also added an integration test. PTAL ?

aabmass (approved):

PR #667, opened 2023-07-07
Add option to parse error logs to Error Reporting type

### LINKED ISSUE #665: Formatting error messages in logs to report error events to GCP Error Reporting service
The current Google Cloud Exporter for logs exports the `textPayload` field. In order for log lines to appear in GCP's Error Reporting service they need to have the following format according to the documentation [here:](https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text)
`"jsonPayload": {
      "@type": "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent",
      "message": "Text message"
    },`

Supporting this use case would greatly assist in integration GMP with GCP's other services, specifically Error Reporting.

Thanks

damemi:
Fixes #665 
This adds an option to the logs exporter, `log.error_reporting_type`, to automatically [format error logs to GCP Error Reporting](https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text). To do this, it adds the `@type` field to existing JSON payloads, or converts string payloads to JSON (with a `message` field) and then adds `@type`.

I don't think there is a behavior that makes sense for byte payloads, since that would mean a mixture of byte and string values in the json payload (?).

An improvement to this could be using the [`ReportedErrorEvent` protobuf](https://github.com/googleapis/google-cloud-go/blob/c9aebae6d9cd8518bc247b7198ed38365577a872/errorreporting/apiv1beta1/errorreportingpb/report_errors_service.pb.go#L148) directly rather than just setting `@type`, as documented in https://cloud.google.com/error-reporting/docs/formatting-error-messages#reported-error-example. This adds a requirement for a `service` field.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -180,6 +180,9 @@ type LogConfig struct {
 	// service.name, service.namespace, and service.instance.id resource attributes into the Cloud Logging LogEntry labels.
 	// Disabling this option does not prevent resource_filters from adding those labels. Default is true.
 	ServiceResourceLabels bool `mapstructure:"service_resource_labels"`
+	// ErrorReportingType enables automatically parsing error logs to a json payload containing the
+	// type value for GCP Error Reporting. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text.
+	ErrorReportingType bool `mapstructure:"error_reporting_type"`
 }
 
 // Known metric domains. Note: This is now configurable for advanced usages.
--- a/exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_error.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_error.json
@@ -277,8 +277,8 @@
             },
             {
               "timeUnixNano": "1651013314740290000",
-              "severityNumber": "SEVERITY_NUMBER_WARN",
-              "severityText": "Warn",
+              "severityNumber": "SEVERITY_NUMBER_ERROR",
+              "severityText": "Error",
               "body": {
                 "kvlistValue": {
                   "values": [
@@ -291,7 +291,7 @@
                     {
                       "key": "severity",
                       "value": {
-                        "stringValue": "warn"
+                        "stringValue": "error"
                       }
                     },
                     {
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -52,6 +52,9 @@ const (
 	LogNameAttributeKey        = "gcp.log_name"
 	SourceLocationAttributeKey = "gcp.source_location"
 	TraceSampledAttributeKey   = "gcp.trace_sampled"
+
+	GCPTypeKey                 = "@type"
+	GCPErrorReportingTypeValue = "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent"
 )
 
 // severityMapping maps the integer severity level values from OTel [0-24]
@@ -462,7 +465,7 @@ func (l logMapper) logToSplitEntries(
 			logRecord.Body().SetEmptyMap()
 			logRecord.Body().Map().PutStr("message", strValue)
 		}
-		logRecord.Body().Map().PutStr("@type", "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent")
+		logRecord.Body().Map().PutStr(GCPTypeKey, GCPErrorReportingTypeValue)
 	}
 
 	entry.Labels = logLabels
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -376,16 +376,20 @@ func (l logMapper) logToSplitEntries(
 	logName string,
 	projectID string,
 ) ([]logging.Entry, error) {
+	// make a copy in case we mutate the record
+	logRecord := plog.NewLogRecord()
+	log.CopyTo(logRecord)
+
 	entry := logging.Entry{
 		Resource: mr,
 	}
 
-	entry.Timestamp = log.Timestamp().AsTime()
-	if log.Timestamp() == 0 {
+	entry.Timestamp = logRecord.Timestamp().AsTime()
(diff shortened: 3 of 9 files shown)


exporter/collector/logs_test.go
```diff
@@ -210,6 +218,32 @@ func TestLogMapping(t *testing.T) {
...
+				log := plog.NewLogRecord()
+				log.SetSeverityNumber(18)
+				log.Body().SetStr("{\"message\": \"hello!\"}")
+				return log
+			},
+			expectedEntries: []logging.Entry{
+				{
+					Payload: map[string]interface{}{
+						"@type":   "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent",
+						"message": `{"message": "hello!"}`,
```
damemi, line 236: note that "json" bodies (that are actually string values) can end up with a redundant message key. I don't think we should try automatically parsing bodies that are string valued to JSON and making assumptions to remove them

exporter/collector/logs.go
```diff
@@ -376,16 +376,20 @@ func (l logMapper) logToSplitEntries(
 	logName string,
 	projectID string,
 ) ([]logging.Entry, error) {
+	// make a copy in case we mutate the record
```
psx95, line 379: Is this necessary? I don't see the usages of `log` variable in this method indicating that we don't care about the original object and its not a pointer, so I am assuming its pass by value.

exporter/collector/logs.go
```diff
@@ -435,22 +439,32 @@ func (l logMapper) logToSplitEntries(
...
 	entry.Severity = severityMapping[severityNumber]
 
+	// Parse severityNumber > 17 (error) to a GCP Error Reporting entry if enabled
+	if severityNumber >= 17 && l.cfg.LogConfig.ErrorReportingType {
+		if logRecord.Body().Type() != pcommon.ValueTypeMap {
+			strValue := logRecord.Body().AsString()
+			logRecord.Body().SetEmptyMap()
+			logRecord.Body().Map().PutStr("message", strValue)
+		}
+		logRecord.Body().Map().PutStr("@type", "type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent")
```
psx95, line 465: I think making this type String to be a constant might be better for readability. But I do not have a strong preference here since this is not used anywhere else.

exporter/collector/logs_test.go
psx95, line 1: Can we add more test cases -  
1. Where the body is already a map
2. Where body is a simple string (rather than it being in the JSON format). 
3. The config contains `true` for `ErrorReportingType` but the log severity number is not high enough for `@type` to be appended.

exporter/collector/logs.go
psx95: I noticed we are using this field in test as well, I would still defer the decision to you, but I now have a slight preference to making this a const field.

exporter/collector/integrationtest/testdata/fixtures/logs/logs_apache_error.json
psx95, line 1: Clarification: Was this change just to reconcile discrepancy between file name and error severity level ?

exporter/collector/logs.go
damemi: the [underlying pdata](https://github.com/open-telemetry/opentelemetry-collector/blob/38615b2ad813a277e4a4fe0b391f3cc43db975b2/pdata/plog/generated_logrecord.go#L24) is a pointer, so collector components aren't supposed to mutate data which could be shared (like across multiple exporters) https://github.com/open-telemetry/opentelemetry-collector/tree/89e13a1d94c230cd1552d5a73e6730dace710921/exporter#data-ownership

exporter/collector/logs.go
```diff
@@ -52,6 +52,9 @@ const (
 	LogNameAttributeKey        = "gcp.log_name"
 	SourceLocationAttributeKey = "gcp.source_location"
 	TraceSampledAttributeKey   = "gcp.trace_sampled"
+
+	GCPTypeKey                 = "@type"
```
psx95, line 56: Nit: Should this be `GCPErrorReportingTypeKey` or `GCPErrorTypeKey`  and `GCPErrorTypeValue`  - basically consistent naming that indicates its a key-value pair to be used together ?

psx95 (approved):
Small nit (variable naming), but looks good to me overall.

exporter/collector/logs.go
damemi: this key isn't specific to Error Reporting, any protobuf marshalled to json will use it to indicate the object type. Error Reporting lets you parse it manually to bypass setting required fields in the proto, which is what we're doing here.

psx95: I see, that makes sense.

exporter/collector/config.go
```diff
@@ -180,6 +180,9 @@ type LogConfig struct {
 	// service.name, service.namespace, and service.instance.id resource attributes into the Cloud Logging LogEntry labels.
 	// Disabling this option does not prevent resource_filters from adding those labels. Default is true.
 	ServiceResourceLabels bool `mapstructure:"service_resource_labels"`
+	// ErrorReportingType enables automatically parsing error logs to a json payload containing the
+	// type value for GCP Error Reporting. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text.
+	ErrorReportingType bool `mapstructure:"error_reporting_type"`
```
dashpole, line 185: nit: it isn't clear from the name what it does.  Prefer send_errors_to_error_reporting, or something that implies what the effect is.

dashpole (approved):

exporter/collector/config.go
damemi: afaict, adding this attribute doesn't send them "to" error reporting. they're still sent to cloud logging, just reformatted to have a special attribute that Error Reporting picks up. I'm not entirely clear on the ux from that point, will talk to some of the folks who work on ER

dashpole: Sounds good.  Thanks for checking!

PR #674, opened 2023-07-17
Reorganize metric integration tests

dashpole:
My motivation for this is to be able to link to individual fixtures for each metric type (gauge, sum, non-monotonic sum, etc) for the standard and GMP exporter in documentation.

This is probably easiest to review commit-by-commit.

* makes a function for the GMP config
* simplifies naming of fixture files (e.g. removes most instances of "metrics" and "basic" from names)
* adds GMP tests for simple scenarios.
* Enables the normalization feature gate for GMP tests.  This demonstrates correct name translation in fixtures for the GMP exporter.
* Change units of metrics to UCUM so the translation is tested.
* Adds basic resource attributes to simple scenarios.
* Disables service metric labels for fixtures
* Change names of metrics and labels to include `.`s so that it tests sanitization.


dashpole:
rebased

exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -37,22 +37,35 @@ var MetricsTestCases = []TestCase{
 		ExpectFixturePath:    "testdata/fixtures/metrics/counter_expect.json",
 		ConfigureCollector: func(cfg *collector.Config) {
 			cfg.MetricConfig.InstrumentationLibraryLabels = true
+			cfg.MetricConfig.ServiceResourceLabels = false
```
damemi, line 40: could you add a comment explaining why resource labels are disabled? I know you explained it to me, but it's already fuzzy

damemi (approved):
nothing blocking, but some suggestions.

it would also be nice if we added a comment on each test explaining what are the critical things to look for in the input/output fixtures. some are pretty self-explanatory but as they get more complex it gets harder to verify the expected fixtures

dashpole:
clarified all of the test case names to make expectations clearer.

exporter/collector/integrationtest/testcases/testcases_metrics.go
dashpole: added a comment

exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -399,3 +367,20 @@ var MetricsTestCases = []TestCase{
 	},
 	// TODO: Add integration tests for workload.googleapis.com metrics from the ops agent
 }
+
+func configureGMPCollector(cfg *collector.Config) {
```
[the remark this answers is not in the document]
dashpole: maybe in the future.  It would need to be moved out of our test cases, though.

Issue #675, opened 2023-07-17
Non-monotonic counters don't have the correct GMP prefix


dashpole:
Found when I added a case in GMP for non-monotonic counters: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/674

```json
{
  "metric": {
    "type": "prometheus.googleapis.com/system_memory_usage/counter",
    "labels": {
      "state": "used"
    }
  },
  "resource": {
    "type": "prometheus_target",
    "labels": {
      "cluster": "",
      "instance": "",
      "job": "",
      "location": "",
      "namespace": ""
    }
  },
  "metricKind": "GAUGE",
  "valueType": "INT64",
  "points": [
    {
      "interval": {
        "endTime": "1970-01-01T00:00:00Z"
      },
      "value": {
        "int64Value": "6641752448"
      }
    }
  ],
  "unit": "By"
}
```

Even though it is converted to a gauge, it gets the /counter suffix.

PR #676, opened 2023-07-17
non-monotonic sums have the /gauge suffix added in GMP exporter

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/675

### DIFF
--- a/exporter/collector/googlemanagedprometheus/naming.go
+++ b/exporter/collector/googlemanagedprometheus/naming.go
@@ -28,6 +28,10 @@ func GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
 	// Second, ad the GMP-specific suffix
 	switch metric.Type() {
 	case pmetric.MetricTypeSum:
+		if !metric.Sum().IsMonotonic() {
+			// Non-monotonic sums are converted to GCM gauges
+			return compliantName + "/gauge", nil
+		}
 		return compliantName + getUnknownMetricSuffix(metric.Sum().DataPoints(), "/counter", "counter"), nil
 	case pmetric.MetricTypeGauge:
 		return compliantName + getUnknownMetricSuffix(metric.Gauge().DataPoints(), "/gauge", ""), nil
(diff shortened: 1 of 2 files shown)


exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -28,7 +28,11 @@ func GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
 	// Second, ad the GMP-specific suffix
 	switch metric.Type() {
 	case pmetric.MetricTypeSum:
-		return compliantName + getUnknownMetricSuffix(metric.Sum().DataPoints(), "/counter", "counter"), nil
+		if metric.Sum().IsMonotonic() {
+			return compliantName + getUnknownMetricSuffix(metric.Sum().DataPoints(), "/counter", "counter"), nil
+		}
+		// Non-monotonic sums are converted to GCM gauges
+		return compliantName + "/gauge", nil
```
aabmass, line 35: nit i'd consider monotonic the happy path and would swap which case ends up in the indented if block

aabmass (approved):
Nit consider adding one of these to the GMP test fixture

exporter/collector/googlemanagedprometheus/naming.go
dashpole: done

dashpole:
I'm adding that case in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/674

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -28,6 +28,10 @@ func GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
 	// Second, ad the GMP-specific suffix
 	switch metric.Type() {
 	case pmetric.MetricTypeSum:
+		if !metric.Sum().IsMonotonic() {
+			// Non-monotonic sums are converted to GCM gauges
+			return compliantName + "/gauge", nil
```
aabmass, line 33: Sorry one more question, why doesn't this case use `getUnknownMetricSuffix`?

dashpole: Prometheus unknown is always an OTel gauge.  When we "double-write" this gauge, we copy the gauge into a monotonic counter as well, hence the handling above.  There isn't a way to get a non-monotonic sum from an unknown metric point.

damemi (approved):

PR #678, opened 2023-07-18
Remove _total from duplicate untyped Sums before export

### LINKED ISSUE #677: Untyped metrics are incorrectly named when metric name normalization is enabled.
Found while working on https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/674.

`prometheus.BuildPromCompliantName` builds a metric name based on the name, unit, and type.  For sums, it adds a `_total` suffix.

If you have an unknown gauge metric named `fake.untyped.metric`, it is renamed to `fake_untyped_metric_total` for the counter version.  Ideally, we should use the compliant name it would have if it was a gauge for the unknown metric.

Want:
```diff
[
  {
    "metric": {
      "type": "prometheus.googleapis.com/fake_untyped_metric/unknown",
      "labels": {
        "ex_com_lemons_untyped": "13"
      }
    },
    "resource": {
      "type": "prometheus_target",
      "labels": {
        "cluster": "",
        "instance": "",
        "job": "",
        "location": "",
        "namespace": ""
      }
    },
    "metricKind": "GAUGE",
    "valueType": "DOUBLE",
    "points": [
      {
        "interval": {
          "endTime": "1970-01-01T00:00:00Z"
        },
        "value": {
          "doubleValue": 3
        }
      }
    ]
  },
  {
    "metric": {
-      "type": "prometheus.googleapis.com/fake_untyped_metric_total/unknown:counter",
+      "type": "prometheus.googleapis.com/fake_untyped_metric/unknown:counter",
      "labels": {
        "ex_com_lemons_untyped": "13"
      }
    },
    "resource": {
      "type": "prometheus_target",
      "labels": {
        "cluster": "",
        "instance": "",
        "job": "",
        "location": "",
        "namespace": ""
      }
    },
    "metricKind": "CUMULATIVE",
    "valueType": "DOUBLE",
    "points": [
      {
        "interval": {
          "endTime": "1970-01-01T00:00:00Z",
          "startTime": "1970-01-01T00:00:00Z"
        },
        "value": {
          "doubleValue": 3
        }
      }
    ]
  }
]
```

damemi:
Fixes #677 

updates GMP exporter's custom `GetMetricName()` to remove `_total` from Sums which have the double-export untyped-metric attribute, if `total` was not present in the original metric name (before being normalized by OTel's [prometheus.BuildPromCompliantName()](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/pkg/translator/prometheus/v0.81.0/pkg/translator/prometheus/normalize_name.go#L89) function).

We need this because the [extra untyped Sum metric](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/191af029411ce4b54ed631733721a2ac58d798e2/exporter/collector/googlemanagedprometheus/extra_metrics.go#L42) is added at [the ExtraMetrics extension point](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/191af029411ce4b54ed631733721a2ac58d798e2/exporter/collector/metrics.go#L304-L307), which is before [name normalization happens](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/191af029411ce4b54ed631733721a2ac58d798e2/exporter/collector/metrics.go#L1150) for the individual metric types.

Alternatively, the extra metric could be added *after* name normalization (ie, copy the normalized Gauge metric name to a new Sum after [this line in gaugePointToTimeSeries](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/191af029411ce4b54ed631733721a2ac58d798e2/exporter/collector/metrics.go#L1204)). But that would require refactoring to add a new extension point, which is not worth the effort with a long term upstream plan in the works. The current method takes advantage of existing metric handling for Gauges+Sums as well, such as cumulative normalization.

### DIFF
--- a/exporter/collector/googlemanagedprometheus/naming.go
+++ b/exporter/collector/googlemanagedprometheus/naming.go
@@ -17,6 +17,7 @@ package googlemanagedprometheus
 import (
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheus"
 	"go.opentelemetry.io/collector/pdata/pmetric"
@@ -32,9 +33,9 @@ func GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
 			// Non-monotonic sums are converted to GCM gauges
 			return compliantName + "/gauge", nil
 		}
-		return compliantName + getUnknownMetricSuffix(metric.Sum().DataPoints(), "/counter", "counter"), nil
+		return getUnknownMetricName(metric.Sum().DataPoints(), "/counter", "counter", metric.Name(), compliantName), nil
 	case pmetric.MetricTypeGauge:
-		return compliantName + getUnknownMetricSuffix(metric.Gauge().DataPoints(), "/gauge", ""), nil
+		return getUnknownMetricName(metric.Gauge().DataPoints(), "/gauge", "", metric.Name(), compliantName), nil
 	case pmetric.MetricTypeSummary:
 		// summaries are sent as the following series:
 		// * Sum: prometheus.googleapis.com/<baseName>_sum/summary:counter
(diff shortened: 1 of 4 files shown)


exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -365,5 +365,36 @@ var MetricsTestCases = []TestCase{
 		// SDK exporter does not support CreateServiceTimeSeries
 		SkipForSDK: true,
 	},
+	{
+		// https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/677
```
dashpole, line 369: Skip this test for now.  I'll be adding it in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/674

dashpole:
unit test would be useful here

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -56,11 +57,23 @@ func GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
...
 // It is based on the untyped_prometheus_metric data point attribute, and behind a feature gate.
-func getUnknownMetricSuffix(points pmetric.NumberDataPointSlice, suffix, secondarySuffix string) string {
+func getUnknownMetricName(points pmetric.NumberDataPointSlice, suffix, secondarySuffix, originalName, compliantName string) string {
 	if !untypedDoubleExportFeatureGate.IsEnabled() {
-		return suffix
+		return compliantName + suffix
 	}
+
+	// de-normalize "_total" suffix for counters where not present on original metric name
+	nameTokens := strings.FieldsFunc(
```
dashpole, line 69: is strings.HasSuffix not sufficient?  Why this extra strings logic?

damemi: This is the same code to break it into tokens that the prom library uses, I think to cover cases like: `sample.metric.total`, `foosecondstotal`, `bar_requests_total`

exporter/collector/integrationtest/testcases/testcases_metrics.go
damemi: marked it skip, imo we should still merge it just so it's in the history and you can remove it in your PR

damemi:
> unit test would be useful here

added a unit test

dashpole:
One thing we will have to think about going forward is that we won't have access to the "unknown" flag from a counter.  The "unknown" flag will be gauge-specific.  That will make our current approach of looking at labels in the naming function no longer work.  We don't need to solve that here, but worth noting

dashpole (approved):

damemi:
fyi @ridwanmsharif: if you see double-exported metrics with `_total` added to the counter, this will fix that

dashpole:
@damemi it looks like this removes `_total` from _all_ counters, not just those that came from untyped.  We should only remove `_total` from counters that come from unknown metrics.

PR #682, opened 2023-07-19
Add option to Metrics SDK exporter to allow sending sum of squared deviation

dashpole:
This is useful for users, but also allows me to test sumofsquareddeviations in the fixture tests for the collector and SDK.

### DIFF
--- a/exporter/metric/metric.go
+++ b/exporter/metric/metric.go
@@ -491,7 +491,7 @@ func (me *metricExporter) recordToTspb(m metricdata.Metrics, mr *monitoredrespb.
 		}
 	case metricdata.Histogram[int64]:
 		for _, point := range a.DataPoints {
-			ts, err := histogramToTimeSeries(point, m, mr)
+			ts, err := histogramToTimeSeries(point, m, mr, me.o.enableSumOfSquaredDeviation)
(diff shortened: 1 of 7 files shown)


exporter/metric/metric.go
```diff
@@ -579,19 +579,28 @@ func sumToTimeSeries[N int64 | float64](point metricdata.DataPoint[N], metrics m
 	}, nil
 }
 
-func histogramToTimeSeries[N int64 | float64](point metricdata.HistogramDataPoint[N], metrics metricdata.Metrics, mr *monitoredrespb.MonitoredResource) (*monitoringpb.TimeSeries, error) {
+//nolint:revive
```
damemi, line 582: is this for the control-coupling lint check? Could you add a comment/todo for that? I did the same in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/a3fa73dadda89df41b0641e74e796145f134a347/exporter/collector/monitoredresource.go#L59-L61

damemi (approved):

exporter/metric/metric.go
dashpole: done

PR #683, opened 2023-07-21
Expand scope of kubernetes monitored resource mapping to include other platforms

### LINKED ISSUE #627: k8s_container monitored resource type requires `cloud.platform: gcp_kubernetes_engine` even when running outside of Google Cloud
In the past, we were able to send logs to Cloud Logging by setting the following resource attributes
- k8s.container.name
- k8s.pod.name
- k8s.namespace.name
- k8s.cluster.name

Recently, I noticed that my logs are no longer shown under the `k8s_container` [monitored resource type](https://cloud.google.com/monitoring/api/resources#tag_k8s_container), instead they appear as `generic_node`.

This is because the [internal/resourcemapping/resourcemapping.go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/internal/resourcemapping/resourcemapping.go#L171) is checking specifically for `cloud.platform: gcp_kubernetes_engine`.

We work with many Google customers who operate Kubernetes clusters outside of Google Cloud, therefore, the requirement for `cloud.platform: gcp_kubernetes_engine` is surprising, especially when using OpenTelemetry.

I can reproduce this with Minikube running on my workstation.

**With cloud.platform set**

Using the resource attribute processor, I can set the cloud.platform value to "trick" the exporter.

Screenshot from 2023-04-21 19-13-48

**without cloud.platform**

If I do not add the attribute, my logs come through but as `generic_node`.

Screenshot from 2023-04-21 19-13-22

damemi:
Fixes #627 
This broadens the exporter's monitored resource mapping to match anything with `k8s.cluster.name` resource attribute to a k8s monitored resource. Using `k8s.cluster.name` means that any cloud provider k8s platforms like GKE, EKS, or AKS should be identified, as well as non-cloud clusters like local k8s or minikube.

This is a change from the previous behavior, in which any non-GKE kubernetes resources were mapped to `generic_node`.

### DIFF
--- a/internal/resourcemapping/resourcemapping.go
+++ b/internal/resourcemapping/resourcemapping.go
@@ -168,12 +168,12 @@ type ReadOnlyAttributes interface {
 // E.g.
 // This may output `gce_instance` type with appropriate labels.
 func ResourceAttributesToMonitoredResource(attrs ReadOnlyAttributes) *GceResource {
-	cloudPlatform, _ := attrs.GetString(string(semconv.CloudPlatformKey))
+	cloudPlatform := getK8sClusterOrCloudPlatform(attrs)
 	var mr *GceResource
 	switch cloudPlatform {
 	case semconv.CloudPlatformGCPComputeEngine.Value.AsString():
 		mr = createMonitoredResource(gceInstance, attrs)
-	case semconv.CloudPlatformGCPKubernetesEngine.Value.AsString():
+	case k8sCluster:
 		// Try for most to least specific k8s_container, k8s_pod, etc
 		if _, ok := attrs.GetString(string(semconv.K8SContainerNameKey)); ok {
 			mr = createMonitoredResource(k8sContainer, attrs)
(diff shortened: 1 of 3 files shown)


internal/resourcemapping/resourcemapping.go
```diff
@@ -208,6 +208,17 @@ func ResourceAttributesToMonitoredResource(attrs ReadOnlyAttributes) *GceResourc
 	return mr
 }
 
+// getK8sClusterOrCloudPlatform returns "k8s_cluster" if K8SClusterNameKey is set
+// (which identifies cloud provider clusters like GKE, EKS, AKS, or non-cloud clusters like minikube)
+// Otherwise, it returns the cloud.platform value.
+func getK8sClusterOrCloudPlatform(attrs ReadOnlyAttributes) string {
+	if _, cluster := attrs.GetString(string(semconv.K8SClusterNameKey)); cluster {
+		return k8sCluster
```
dashpole, line 216: nit: kind of strange to mix platform semantic conventions and other strings.  Maybe return semconv.CloudPlatformGCPKubernetesEngine.Value.AsString() here instead?

dashpole (approved):

internal/resourcemapping/resourcemapping.go
damemi: Yeah, I thought about that too. I don't think it's perfect either way (felt weird to return the platform as GCP when it could be any k8s).

It's odd to me that there isn't an attribute to indicate generic K8S platform. The best you can do, as far as I can tell, is just infer if `k8s.cluster.name` is set

PR #690, opened 2023-07-28
Add TraceFlags support to logs exporter

### LINKED ISSUE #685: Support TraceFlags in logging exporter
The collector logs exporter should parse the [LogRecord `TraceFlags` field](https://github.com/open-telemetry/opentelemetry-specification/blob/dc78006c12d9767fd2e35b691706c7572a76fd43/specification/logs/data-model.md#field-traceflags) to indicate if the trace is sampled.

It currently uses [`gcp.trace_sampled`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/54bb068ff7508c90901312926eeb81298a11b0b4/exporter/collector/logs.go#L422) (https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/444) which maps to the GCP LogEntry [`traceSampled` field](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). This matches the spec we have, where all `gcp.*` attributes map to the matching GCP-specific log field: https://github.com/open-telemetry/opentelemetry-specification/blob/dc78006c12d9767fd2e35b691706c7572a76fd43/specification/logs/data-model-appendix.md#google-cloud-logging, but `TraceFlags` should also be respected and mapped to this.

Slack thread: https://cloud-native.slack.com/archives/C043XTR0JHF/p1689196581019039

damemi:
Fixes #685 
Adds support for TraceFlags to parse traceID sampling (alongside the existing `gcp.trace_sampled` attribute)

Right now, if either flag is set then TraceSampled will be true. We might want to shift to one taking precedence, but for backward compatibility this is the least breaking way to introduce the feature.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -418,9 +418,9 @@ func (l logMapper) logToSplitEntries(
 		delete(attrsMap, SourceLocationAttributeKey)
 	}
 
-	// parse TraceSampled boolean from OTel attribute
-	if traceSampled, ok := attrsMap[TraceSampledAttributeKey]; ok {
-		entry.TraceSampled = traceSampled.Bool()
+	// parse TraceSampled boolean from OTel attribute or IsSampled OTLP flag
+	if traceSampled, ok := attrsMap[TraceSampledAttributeKey]; ok || logRecord.Flags().IsSampled() {
+		entry.TraceSampled = (traceSampled.Bool() || logRecord.Flags().IsSampled())
 		delete(attrsMap, TraceSampledAttributeKey)
 	}
 
--- a/exporter/collector/integrationtest/testdata/fixtures/logs/logs_span_trace_id.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/logs/logs_span_trace_id.json
@@ -31,7 +31,7 @@
                             {
                               "key": "msg",
                               "value": {
-                                "stringValue": "hello world 0"
+                                "stringValue": "this log has IsSampled=0"
                               }
                             }
                           ]
@@ -41,6 +41,7 @@
                   ]
                 }
               },
+              "flags": 0,
               "attributes": [
                 {
                   "key": "file.name",
@@ -83,7 +84,7 @@
                             {
                               "key": "msg",
                               "value": {
-                                "stringValue": "hello world 1"
+                                "stringValue": "this log has IsSampled=1"
                               }
                             }
                           ]
@@ -93,6 +94,7 @@
                   ]
                 }
               },
+              "flags": 1,
               "attributes": [
                 {
                   "key": "file.name",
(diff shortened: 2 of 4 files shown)


exporter/collector/logs.go
```diff
@@ -418,9 +418,9 @@ func (l logMapper) logToSplitEntries(
 		delete(attrsMap, SourceLocationAttributeKey)
 	}
 
-	// parse TraceSampled boolean from OTel attribute
-	if traceSampled, ok := attrsMap[TraceSampledAttributeKey]; ok {
-		entry.TraceSampled = traceSampled.Bool()
+	// parse TraceSampled boolean from OTel attribute or IsSampled OTLP flag
+	if traceSampled, ok := attrsMap[TraceSampledAttributeKey]; ok || logRecord.Flags().IsSampled() {
```
dashpole, line 422: Just curious, why did we have a "special" attribute in the first place?

exporter/collector/logs.go
```diff
@@ -418,9 +418,9 @@ func (l logMapper) logToSplitEntries(
...
+		entry.TraceSampled = (traceSampled.Bool() || logRecord.Flags().IsSampled())
 		delete(attrsMap, TraceSampledAttributeKey)
```
dashpole, line 424: Will this fail if the attribute doesn't exist?

exporter/collector/logs.go
damemi: In the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/dc78006c12d9767fd2e35b691706c7572a76fd43/specification/logs/data-model-appendix.md#google-cloud-logging) we map all GCP LogEntry fields to a matching `gcp.*` attribute. [LogEntry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) has a field `traceSampled`, so we get `gcp.trace_sampled` as the matching attribute.

exporter/collector/logs.go
damemi: nope, `delete()` is a no-op if the element doesn't exist. Also added the tests the verify this

exporter/collector/logs.go
dashpole: Should we update the spec to say that we should use the log record sampled flag instead of this attribute?  We can keep it in the exporter for backwards-compatibility, but sampled isn't something I expect to get from attributes generally speaking.

damemi: Yup good point, either way if we're adding support for log Flags we should update that spec doc to show it. I'll open a PR for that too

dashpole (approved):

exporter/collector/logs.go
damemi: https://github.com/open-telemetry/opentelemetry-specification/pull/3629

PR #691, opened 2023-07-28
feat: Add invalid_json_byte_strings option to logs exporter

### LINKED ISSUE #689: Exporting log records with a non-JSON Byte body fails
If the pdata log record is Bytes, we see this error in our logs:
```
{"level":"warn","ts":"2023-07-27T14:02:52.373-0400","caller":"batchprocessor@v0.81.0/batch_processor.go:258","msg":"Sender failed","kind":"exporter","data_type":"logs","name":"googlecloud/gcloudtest","error":"logging: json.Unmarshal: json: cannot unmarshal string into Go value of type map[string]interface {}"}
```

It looks like here, the body is extracted as an []bytes:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/f2188d1e6cfb3c1be26021962ae8f03eb42b916b/exporter/collector/logs.go#L539-L540

Then the payload is put directly on the entry:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/f2188d1e6cfb3c1be26021962ae8f03eb42b916b/exporter/collector/logs.go#L530

However, the [docs for logging.Entry](https://pkg.go.dev/cloud.google.com/go/logging#Entry) states that payload must be string or marshal to a JSON object (not a raw value), which `[]byte` does not do.

damemi:
Fixes #689 

Cloud Logging requires payloads to be a string, or something that marshals to a JSON object. However, a Log body in OTLP can be anything, even bytes that don't necessarily represent valid JSON.

Currently, sending non-JSON bytes through the exporter results in a json.Unmarshal error. But, users may want to export these bytes as a byte string, not necessarily JSON.

You can workaround this by transforming the data to an object. This PR adds a feature to pass along the raw byte string as a string payload instead.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -183,6 +183,11 @@ type LogConfig struct {
 	// ErrorReportingType enables automatically parsing error logs to a json payload containing the
 	// type value for GCP Error Reporting. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text.
 	ErrorReportingType bool `mapstructure:"error_reporting_type"`
+	// InvalidJsonByteStrings exports any Log body of type byte as a raw byte string, when the
+	// byte body does not marshal to valid JSON. Cloud Logging requires byte payloads to marshal to
+	// valid JSON via the encoding/json package. When byte payloads do not translate to valid JSON,
+	// this allows them to still be passed to Cloud Logging as a raw string payload.
+	InvalidJSONByteStrings bool `mapstructure:"invalid_json_byte_strings"`
 }
 
 // Known metric domains. Note: This is now configurable for advanced usages.
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -495,7 +495,7 @@ func (l logMapper) logToSplitEntries(
 	overheadClone := proto.Clone(logOverhead)
 	overheadBytes := proto.Size(overheadClone)
 
-	payload, splits, err := parseEntryPayload(logRecord.Body(), l.maxEntrySize-overheadBytes)
+	payload, splits, err := l.parseEntryPayload(logRecord.Body(), l.maxEntrySize-overheadBytes)
 	if err != nil {
 		return []logging.Entry{entry}, err
 	}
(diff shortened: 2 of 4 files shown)


exporter/collector/config.go
```diff
@@ -183,6 +183,11 @@ type LogConfig struct {
 	// ErrorReportingType enables automatically parsing error logs to a json payload containing the
 	// type value for GCP Error Reporting. See https://cloud.google.com/error-reporting/docs/formatting-error-messages#log-text.
 	ErrorReportingType bool `mapstructure:"error_reporting_type"`
+	// InvalidJsonByteStrings exports any Log body of type byte as a raw byte string, when the
+	// byte body does not marshal to valid JSON. Cloud Logging requires byte payloads to marshal to
+	// valid JSON via the encoding/json package. When byte payloads do not translate to valid JSON,
+	// this allows them to still be passed to Cloud Logging as a raw string payload.
+	InvalidJSONByteStrings bool `mapstructure:"invalid_json_byte_strings"`
```
dashpole, line 190: Should this be the default behavior?  It seems like by default we should make sure that valid OTLP is delivered to cloud logging.

damemi: I was thinking users might want to be aware if their valid OTLP isn't valid as a LogEntry. But the collector warning here could be good enough for that. If anything, maybe this option should be inverted (ie, `dont_send_invalid_bytes_json_as_a_string`) but until someone requests that, we can go with this as the default. The expectation seems overwhelmingly to be that if it's valid OTLP, it should end up in cloud logging. Updated

exporter/collector/logs.go
```diff
@@ -531,15 +531,24 @@ func (l logMapper) logToSplitEntries(
...
 		return nil, 0, nil
 	}
+
+	// number of payloads calculation, used when returning a string payload
+	payloads := int(math.Ceil(float64(len([]byte(logBody.AsString()))) / float64(maxEntrySize)))
+
 	switch logBody.Type() {
 	case pcommon.ValueTypeBytes:
+		var verify map[string]interface{}
+		if err := json.Unmarshal(logBody.Bytes().AsRaw(), &verify); err != nil {
```
dashpole, line 545: If we always sent the log body as a string, would that break users?  Or would the logging API still display it correctly?

damemi: if it can be parsed as an object, we want to send it as an object (protoPayload/jsonPayload). I don't think we want to send everything as a string

aabmass, line 545: It feels like we are having to hack around the handwritten Cloud Logging client library since it expects an `interface{}` and makes some assumptions about how you want your data to be displayed based on type and format. If we used the "raw" protobuf based client, I think this code could be simplified and more explicit. And also we have to unmarshal twice (once here and once in the client library) as is.

Is there an escape-hatch to pass a LogEntry **protobuf** instead of `interace{}`?

exporter/collector/logs.go
```diff
@@ -531,15 +531,27 @@ func (l logMapper) logToSplitEntries(
...
 }
 
-func parseEntryPayload(logBody pcommon.Value, maxEntrySize int) (interface{}, int, error) {
+func (l logMapper) parseEntryPayload(logBody pcommon.Value, maxEntrySize int) (interface{}, int, error) {
 	if len(logBody.AsString()) == 0 {
 		return nil, 0, nil
 	}
+
+	// number of payloads calculation, used when returning a string payload
+	payloads := int(math.Ceil(float64(len([]byte(logBody.AsString()))) / float64(maxEntrySize)))
```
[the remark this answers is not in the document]
damemi: Good call, I pushed that change

exporter/collector/logs.go
damemi: You're right, I think we could switch this to use the protobuf directly. Opened a separate PR showing that as a poc in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/692, but I couldn't get it to work entirely.

PR #692, opened 2023-07-31
WIP Switch to using logging protos directly

damemi:
This switches the logging exporter to use the generated logging protos directly, rather than using the handwritten `logging.LogEntry` structs (which we were just converting to the generated protos before sending anyway)

The only thing I couldn't figure out so far is getting the httpRequest conversion to work

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
-	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


exporter/collector/logs.go
```diff
@@ -527,8 +501,54 @@ func (l logMapper) logToSplitEntries(
...
+		var verify map[string]interface{}
+		if err := json.Unmarshal(jb, &verify); err == nil {
+			entry.Payload = &logpb.LogEntry_JsonPayload{JsonPayload: jsonMapToProtoStruct(verify)}
+		}
+	}
+
+	return []logpb.LogEntry{entry}, nil
+}
+
+func jsonMapToProtoStruct(m map[string]interface{}) *structpb.Struct {
```
damemi, line 523: this is directly copied from how the logging client converts with `ToLogEntry`: https://github.com/googleapis/google-cloud-go/blob/69705144832c715cf23832602ad9338b911dff9a/logging/logging.go#L853

Issue #679, opened 2023-07-18
Update usages of `faas.id` resource attribute to `faas.instance`


damemi:
Our resource detectors and monitored resource mappings currently rely on semconv v1.18, which has [`faas.id`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.18.0/specification/resource/semantic_conventions/faas.md#faas-resource-attributes), which we use for the Generic Task (Cloud Run) resource.

As of semconv v1.19.0, [`faas.id` is removed](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.19.0/specification/resource/semantic_conventions/faas.md#faas-resource-attributes)(/replaced with [`cloud.resource_id`](https://github.com/open-telemetry/opentelemetry-specification/pull/3188)).

So, when we update to semconv v1.19, we'll need to update our usages of `faas.id` to `faas.instance`:

* Upstream resource detection processor: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/02f688dd0e96cd10bb23a030739887570533e61d/processor/resourcedetectionprocessor/internal/gcp/gcp.go#L89-L91
* [genericTask resource mapping](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/b6fc3f0d850c1512de3a2a22bcba9904843eb6bd/internal/resourcemapping/resourcemapping.go#L140)
  * [also parsed in the mapping switch statement](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/b6fc3f0d850c1512de3a2a22bcba9904843eb6bd/internal/resourcemapping/resourcemapping.go#L201)
* [FaasID detector function (for reference, won't need to be updated)](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/b6fc3f0d850c1512de3a2a22bcba9904843eb6bd/detectors/gcp/faas.go#L91)

This attribute is also used in our code samples for GMP-Cloud Run integration: https://github.com/GoogleCloudPlatform/golang-samples/blob/e4eddad4457dc10f0fabb1b6f4c1906bea46aedc/run/custom-metrics/collector/collector-config.yaml#L45, which will need to be updated and the matching docs pages manually redeployed.

dashpole:
Proposed Plan:

1. Update GMP exporter to map `faas.name` to `job`, and `faas.instance` to `instance`.
2. Update e2e test framework to expect the new attribute.
[...]

PR #695, opened 2023-08-04
Map faas.name and faas.instance to prometheus job and instance for GMP

dashpole:
Step 1 for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/679.

Since service.name has a default value, we need faas.name to take precedence over service.name.  But I still want to give users the option to override job, so i've opted to add `job` as the override to match other labels.

For instance, i've opted to follow the pattern set by name.

In theory, this could be breaking for users if they are setting service.name or service.instance.id on cloud run and expect it to show up.  But if we don't take this approach, this doesn't make the migration to the new value seamless for users that followed the example.

Thoughts? @punya @aabmass


aabmass (approved):
LGTM and existing tests are passing

PR #697, opened 2023-08-06
Second attempt: Switch to using log protos directly

dashpole:
Superceeds https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/692

This switches the logging exporter to use the generated logging protos directly, rather than using the handwritten logging.LogEntry structs (which we were just converting to the generated protos before sending anyway)

I ensured that this maintains the same behavior by starting by copying in the implementation of ToLogEntry and working from that point.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -306,7 +306,7 @@ func (l logMapper) createEntries(ld plog.Logs) (map[string][]*logpb.LogEntry, er
 					if _, ok := entries[projectMapKey]; !ok {
 						entries[projectMapKey] = make([]*logpb.LogEntry, 0)
 					}
-					entries[projectMapKey] = append(entries[projectMapKey], entry)
+					entries[projectMapKey] = append(entries[projectMapKey], &entry)
 				}
 			}
 		}
@@ -468,12 +468,12 @@ func (l logMapper) logToSplitEntries(
 	processTime time.Time,
 	logName string,
 	projectID string,
-) ([]*logpb.LogEntry, error) {
+) ([]logpb.LogEntry, error) {
 	// make a copy in case we mutate the record
 	logRecord := plog.NewLogRecord()
 	log.CopyTo(logRecord)
 
-	entry := &logpb.LogEntry{
+	entry := logpb.LogEntry{
 		Resource: mr,
 	}
 
@@ -583,16 +583,8 @@ func (l logMapper) logToSplitEntries(
 	}
 	entry.LogName = fmt.Sprintf("projects/%s/logs/%s", projectID, url.PathEscape(logName))
 
-	// Calculate the size of the internal log entry so this overhead can be accounted
-	// for when determining the need to split based on payload size
-	// TODO(damemi): Find an appropriate estimated buffer to account for the LogSplit struct as well
-	// make a copy so the proto initialization doesn't modify the original entry
-	// TODO(dashpole): proto clone is modifying the requests...
-	overheadClone := proto.Clone(entry)
-	overheadBytes := proto.Size(overheadClone)
-
 	if len(logRecord.Body().AsString()) == 0 {
-		return []*logpb.LogEntry{entry}, nil
+		return []logpb.LogEntry{entry}, nil
 	}
 
 	switch logRecord.Body().Type() {
@@ -609,9 +601,13 @@ func (l logMapper) logToSplitEntries(
 		}
 		entry.Payload = &logpb.LogEntry_JsonPayload{JsonPayload: s}
 	case pcommon.ValueTypeStr:
+		// Calculate the size of the internal log entry so this overhead can be accounted
+		// for when determining the need to split based on payload size
+		// TODO(damemi): Find an appropriate estimated buffer to account for the LogSplit struct as well
+		overheadBytes := proto.Size(&entry)
 		// Split log entries with a string payload into fewer entries
 		splits := int(math.Ceil(float64(len([]byte(logRecord.Body().AsString()))) / float64(l.maxEntrySize-overheadBytes)))
-		entries := make([]*logpb.LogEntry, splits)
+		entries := make([]logpb.LogEntry, splits)
 		payloadString := logRecord.Body().AsString()
 		// Start by assuming all splits will be even (this may not be the case)
 		startIndex := 0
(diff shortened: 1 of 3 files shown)


exporter/collector/logs.go
```diff
@@ -468,47 +444,52 @@ func (l logMapper) logToSplitEntries(
...
-
+	switch logRecord.Body().Type() {
+	case pcommon.ValueTypeBytes:
+		s, err := toProtoStruct(logRecord.Body().Bytes().AsRaw())
+		if err != nil {
+			return nil, err
+		}
+		entry.Payload = &logpb.LogEntry_JsonPayload{JsonPayload: s}
+	case pcommon.ValueTypeMap:
+		s, err := toProtoStruct(logRecord.Body().Map().AsRaw())
```
aabmass, line 470: This can now be done directly without the JSON marshaling/unmarshal cycle https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewStruct

exporter/collector/logs.go
```diff
@@ -562,54 +531,126 @@ type httpRequestLog struct {
...
+func jsonMapToProtoStruct(m map[string]interface{}) *structpb.Struct {
+	fields := map[string]*structpb.Value{}
+	for k, v := range m {
+		fields[k] = jsonValueToStructValue(v)
+	}
+	return &structpb.Struct{Fields: fields}
+}
+
+// Copied from https://github.com/googleapis/google-cloud-go/blob/69705144832c715cf23832602ad9338b911dff9a/logging/logging.go#L612
+func jsonValueToStructValue(v interface{}) *structpb.Value {
```
aabmass, line 613: I'm not sure if this code predates this function, but wouldn't [`structpb.NewStruct()`](https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NewStruct) work instead of this?

exporter/collector/logs_test.go
```diff
@@ -76,14 +79,26 @@ func TestLogMapping(t *testing.T) {
...
 				{
-					Payload:   "abc",
-					Timestamp: testObservedTime,
+					LogName:   logName,
+					Payload:   &logpb.LogEntry_TextPayload{TextPayload: "abc"},
+					Timestamp: timestamppb.New(testObservedTime),
+					Split: &logpb.LogSplit{
+						Uid:         fmt.Sprintf("default-log-%s", testObservedTime.String()),
+						Index:       int32(0),
+						TotalSplits: int32(2),
```
aabmass, line 90: Are the `int32()` casts needed?

exporter/collector/logs_test.go
```diff
@@ -125,7 +144,7 @@ func TestLogMapping(t *testing.T) {
 			name: "log with json and httpRequest, empty monitoredresource",
 			log: func() plog.LogRecord {
 				log := plog.NewLogRecord()
-				log.Body().SetEmptyBytes().FromRaw([]byte(`{"message": "hello!"}`))
+				log.Body().SetEmptyMap().PutStr("message", "hello!")
```
aabmass, line 147: This change seems to actually be testing something different, was it intentional?

exporter/collector/logs_test.go
```diff
@@ -146,17 +165,24 @@ func TestLogMapping(t *testing.T) {
...
+					HttpRequest: &logtypepb.HttpRequest{
+						RequestMethod:                  "GET",
+						UserAgent:                      "test",
+						Referer:                        "https://www.example2.com",
+						RequestUrl:                     "https://www.example.com",
+						Protocol:                       "HTTP/1.1",
 						RequestSize:                    1,
 						Status:                         200,
 						ResponseSize:                   1,
-						LocalIP:                        "192.168.0.2",
```
aabmass, line 158: Seems weird that there is no `LocalIP` in the plane protobufs.

exporter/collector/logs_test.go
dashpole: The original code doesn't work.  Basically, the previous test only checked that we generated the expected logging.LogEntry.  If you tried to translate that to the proto LogEntry, it produced an error. See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/698

dashpole: I even tried marshalling a map[string]string to bytes using json.Marshal, and that still didn't work for me...

exporter/collector/logs_test.go
dashpole: See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/697/files#diff-5b79131adcc4992f4a98bfaa91757bb5cc73c66f4412ed468d7e2fcd4fa7640eL600.  LocalIP == ServerIp

exporter/collector/logs.go
```diff
@@ -562,54 +531,126 @@ type httpRequestLog struct {
...
 	}
+	return pb, nil
+}
 
-	return httpRequest, nil
+// toProtoStruct converts v, which must marshal into a JSON object,
+// into a Google Struct proto.
+// Mostly copied from
+// https://github.com/googleapis/google-cloud-go/blob/69705144832c715cf23832602ad9338b911dff9a/logging/logging.go#L577
+func toProtoStruct(v interface{}) (*structpb.Struct, error) {
```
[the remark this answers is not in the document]
dashpole: done

exporter/collector/logs.go
dashpole: It looks like it works, so I changed it to use that.

exporter/collector/logs_test.go
dashpole: removed

exporter/collector/logs.go
dashpole: done

exporter/collector/logs_test.go
aabmass: Got it thanks

exporter/collector/logs_test.go
aabmass: Got it hanks

exporter/collector/logs.go
```diff
@@ -562,54 +531,126 @@ type httpRequestLog struct {
...
-		bytes = httpRequestAttr.Bytes().AsRaw()
+		httpBytes = httpRequestAttr.Bytes().AsRaw()
 	case pcommon.ValueTypeStr, pcommon.ValueTypeMap:
-		bytes = []byte(httpRequestAttr.AsString())
+		httpBytes = []byte(httpRequestAttr.AsString())
 	}
 
 	// TODO: Investigate doing this without the JSON unmarshal. Getting the attribute as a map
 	// instead of a slice of bytes could do, but would need a lot of type casting and checking
 	// assertions with it.
```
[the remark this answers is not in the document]
aabmass: Do you know we don't handle map right now? It looks like `httpBytes` would be `nil` in the map case today causing the JSON unmarshal to fail.

If it's a new feature we can save this for a follow up PR. Feel free to resolve this comment.

aabmass (approved):

exporter/collector/logs.go
```diff
@@ -468,47 +444,52 @@ func (l logMapper) logToSplitEntries(
...
+			return nil, err
+		}
+		entry.Payload = &logpb.LogEntry_JsonPayload{JsonPayload: s}
+	case pcommon.ValueTypeStr:
+		// Calculate the size of the internal log entry so this overhead can be accounted
+		// for when determining the need to split based on payload size
+		// TODO(damemi): Find an appropriate estimated buffer to account for the LogSplit struct as well
+		overheadBytes := proto.Size(entry)
+		// Split log entries with a string payload into fewer entries
+		payloadString := logRecord.Body().AsString()
```
aabmass, line 481: nit since we've already checked its string typed
```suggestion
		payloadString := logRecord.Body().Str()
```

exporter/collector/logs.go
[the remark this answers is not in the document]
dashpole: I'll file an issue.

PR #701, opened 2023-08-08
Always fall back to string payload if we can't handle structured logs

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/689

Supersedes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/689

This is slightly broader in scope than https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/689.  That PR only handled the case where we are unable to marshal the payload to JSON.  It could still fail to export for other reasons, such as map keys or values containing invalid UTF-8.  It could also fail if we didn't support the ValueType (e.g. ValueTypeSlice, or ValueTypeInt).

This PR changes the logs exporter to always fall back to attempting to send a raw string payload if we are unable to create a structpb.Struct from the body.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -434,15 +434,12 @@ func (l logMapper) logToSplitEntries(
 	}
 	entry.Severity = severityMapping[severityNumber]
 
-	// log.Body().AsString() can be expensive, and we use it several times below this, so
-	// do it once and save that as a variable.
-	logBodyString := logRecord.Body().AsString()
-
 	// Parse severityNumber > 17 (error) to a GCP Error Reporting entry if enabled
 	if severityNumber >= 17 && l.cfg.LogConfig.ErrorReportingType {
 		if logRecord.Body().Type() != pcommon.ValueTypeMap {
+			strValue := logRecord.Body().AsString()
 			logRecord.Body().SetEmptyMap()
-			logRecord.Body().Map().PutStr("message", logBodyString)
+			logRecord.Body().Map().PutStr("message", strValue)
 		}
 		logRecord.Body().Map().PutStr(GCPTypeKey, GCPErrorReportingTypeValue)
 	}
@@ -456,10 +453,6 @@ func (l logMapper) logToSplitEntries(
 		if _, ok := entry.Labels[k]; !ok {
 			entry.Labels[k] = v.AsString()
 		}
-	}
-
-	if len(logBodyString) == 0 {
-		return []*logpb.LogEntry{entry}, nil
 	}
 
 	// Handle map and bytes as JSON-structured logs if they are successfully converted.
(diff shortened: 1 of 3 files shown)


exporter/collector/logs.go
```diff
@@ -455,66 +458,66 @@ func (l logMapper) logToSplitEntries(
...
-				Uid:         fmt.Sprintf("%s-%s", logName, entry.Timestamp.AsTime().String()),
-				Index:       int32(i),
-				TotalSplits: int32(splits),
-			}
-			entries[i] = newEntry
-
-			// Update slice indices to the next chunk
-			startIndex = endIndex
-			endIndex = int(math.Floor((float64(i+2) / float64(splits)) * float64(len(payloadString))))
+		l.obs.log.Warn(fmt.Sprintf("bytes body cannot be converted to a json payload, exporting as raw string: %+v", err))
```
aabmass, line 480: nit
```suggestion
		l.obs.log.Warn(fmt.Sprintf("bytes body cannot be converted to a json payload, exporting as base64 string: %+v", err))
```

exporter/collector/logs.go
```diff
@@ -434,12 +434,15 @@ func (l logMapper) logToSplitEntries(
 	}
 	entry.Severity = severityMapping[severityNumber]
 
+	// log.Body().AsString() can be expensive, and we use it several times below this, so
+	// do it once and save that as a variable.
+	logBodyString := logRecord.Body().AsString()
```
aabmass, line 439: It looks like we could avoid this call altogether in the map case, I'm wondering if it's worth optimizing for this.

Is a map-typed body our recommended way of doing structured logging in OTel -> collector?

dashpole: moved it to after we handle maps and bytes.  This means we can send empty bytes as json, but that is probably fine

dashpole: I'm not sure about the second question.  I suspect maps are easiest, but i'm not sure what structured json would parse into later if you parsed a structured log.

exporter/collector/logs.go
dashpole: fixed

exporter/collector/logs.go
```diff
@@ -455,66 +455,69 @@ func (l logMapper) logToSplitEntries(
...
-				currentSplit = payloadString[startIndex:endIndex]
-			}
-			newEntry.Payload = &logpb.LogEntry_TextPayload{TextPayload: currentSplit}
-			newEntry.Split = &logpb.LogSplit{
-				Uid:         fmt.Sprintf("%s-%s", logName, entry.Timestamp.AsTime().String()),
-				Index:       int32(i),
-				TotalSplits: int32(splits),
-			}
-			entries[i] = newEntry
+		l.obs.log.Warn(fmt.Sprintf("bytes body cannot be converted to a json payload, exporting as base64 string: %+v", err))
```
aabmass, line 473: I'm wondering if this is really a warnable case. I could imagine someone just trying to log some big byte blob and getting spammed with warning. Wdyt?

aabmass (approved):

exporter/collector/logs.go
aabmass: The question was mostly aiming to understand if its worth optimizing. The change LGTM so resolving

exporter/collector/logs.go
dashpole: Maybe debug would be more appropriate.  Updated

Issue #696, opened 2023-08-06
Incorrect Log Resource Type for GAE Logs


Cooksauce:
Scenario:
- otel collector w/ GoogeCloud exporter
- app running in Google App Engine exporting OTEL data to otel collector

Desired Behavior: Correctly forward the GAE labels into GCP Cloud Logging.

---

## Error - **Blocks all log export**:
It appears the logs export is using the wrong `resource.type` value for App Engine. It using `gae_instance` when the correct value appears to be `gae_app`. 

#### Here is the error itself:
> error	exporterhelper/queued_retry.go:367	Exporting failed. Try enabling retry_on_failure config option to retry on retryable errors	{"kind": "exporter", "data_type": "logs", "name": "googlecloud", "error": "rpc error: code = InvalidArgument desc = Resource type not supported: gae_instance\nerror details: name = Unknown  desc = log_entry_errors:{key:0 value:{code:3 message:\"Resource type not supported: gae_instance\"}}"}

#### Logs Explorer prefilled label filter:
<img width="236" alt="image" src="https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/assets/31948306/da8b6f8b-6ac5-415e-86e2-cc9eb2584a2e">

---

Digging into the source, it looks like the issue might be this constant value [here - resourcemapping.go#L46](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ce6e82d0b6c9bb65f3564c372ea09b49ba610633/internal/resourcemapping/resourcemapping.go#L46)... which is then passed into the `createMonitoredResource(..)` function.

dashpole:
Thanks for reporting this.  I'll make sure it gets fixed.

Cooksauce:
@dashpole Do you have any guesstimate as to when that PR might get merged and released?

dashpole:
It should be fixed this week.  Its been 3 weeks since our last release, so it is probably appropriate to cut a release after that as well.

PR #702, opened 2023-08-08
Write logs to gae_app instead of gae_instance

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/696

I also simplified MR mapping by moving returning a *monitoredrespb.MonitoredResource from the internal package, and performing utf-8 sanitization in the internal package.


dashpole:
@aabmass can you take a look when you get the chance?

aabmass (approved):

dashpole:
Yes.  Cloud logging has a different MR for GAE than cloud monitoring does.

PR #703, opened 2023-08-09
Add an experimental googleclientauth extension

dashpole:
Similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/oauth2clientauthextension, but supports google application default credentials.

This extension:

* Gets application default credentials
* Uses project and quota project, if those are set in the credentials.
* Ensures the user has set a project

Things I didn't do, but could have:

* Add impersonation, similar to the [googlecloud collector exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/419d54a3d9f6bc8aec86332b117ef3cb91db1ae5/exporter/collector/config.go#L157)
* Support `X-Goog-Request-Reason`, or other [system parameters](https://cloud.google.com/apis/docs/system-parameters).
* Support other [CredentialsParams](https://pkg.go.dev/golang.org/x/oauth2/google#CredentialsParams) that the client library supports when getting ADC.

Testing: I don't have a way to test this right now, but it seems simple enough that it should work.


dashpole:
I think I've thought of a way to test this.  We can support custom authenticators on our exporter in client config, and run integration tests using this as a source of auth

dashpole:
I've decided not to do the integration test in this PR, so i've reverted that commit.  I can do it as a follow-up, if it is needed.

dashpole:
Through the experiment with testing, I have confirmed that the extension can be used with the cloud logging, monitoring, and trace APIs.

extension/googleclientauthextension/README.md
```diff
@@ -0,0 +1,53 @@
...
+  pipelines:
+    metrics:
+      receivers: [otlp]
+      processors: []
+      exporters: [otlp/withauth]
+```
+
+Following are the configuration fields:
+- **project** - The Google Cloud Project telemetry is sent to if the gcp.project.id resource attribute is not set. If unspecified, this is determined using application default credentials.
+- [**scopes**](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) - The resource server's token endpoint URLs.
```
aabmass, line 42: This description doesn't look right

extension/googleclientauthextension/grpc.go
```diff
@@ -0,0 +1,46 @@
...
+// https://github.com/googleapis/google-api-go-client/blob/113082d14d54f188d1b6c34c652e416592fc51b5/transport/grpc/dial.go#L224
+func (ca clientAuthenticator) GetRequestMetadata(ctx context.Context, uri ...string) (map[string]string, error) {
+	metadata, err := ca.TokenSource.GetRequestMetadata(ctx, uri...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Attach system parameters
+	if ca.config.QuotaProject != "" {
+		metadata["X-goog-user-project"] = ca.config.QuotaProject
```
aabmass, line 40: Should the `X` be lowercase as well https://cloud.google.com/apis/docs/system-parameters#:~:text=HTTP%20request%20headers%20with%20keys%20in%20lowercase ?

aabmass (approved):
LGTM

extension/googleclientauthextension/README.md
dashpole: fixed

Issue #704, opened 2023-08-14
Migrate from multierr to errors.Join


dashpole:
There is now support in the base language for joining errors.  https://pkg.go.dev/errors#Join.  We should use that instead to minimize dependencies.

PR #708, opened 2023-08-15
Migrate from multierr.Combine to errors.Join

dashpole:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/704

From the definition of [errors.Join](https://pkg.go.dev/errors#Join):

> Join returns an error that wraps the given errors. Any nil error values are discarded. Join returns nil if every value in errs is nil. The error formats as the concatenation of the strings obtained by calling the Error method of each element of errs, with a newline between each string.

> A non-nil error returned by Join implements the Unwrap() []error method.

### DIFF
--- a/e2e-test-server/go.mod
+++ b/e2e-test-server/go.mod
@@ -30,8 +30,6 @@ require (
 	github.com/googleapis/gax-go/v2 v2.11.0 // indirect
 	go.opencensus.io v0.24.0 // indirect
 	go.opentelemetry.io/otel/metric v1.16.0 // indirect
-	go.uber.org/atomic v1.10.0 // indirect
-	go.uber.org/multierr v1.8.0 // indirect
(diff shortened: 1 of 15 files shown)


aabmass (approved):
LGTM if you can get the tests to pass

anuraaga:
Hi @dashpole - it looks like this was merged and released with `go 1.21`. Given your last comment, wondering if this was supposed to be `go 1.20`?

dashpole:
@anuraaga I went ahead and updated to 1.21, since we support the latest 2 versions of go.  Did this cause issues for you?

anuraaga:
I saw a warning when compiling with 1.20, not sure if there was something special about my build setup. But AFAIK the intent of that flag  is for libraries to always set the version in go.mod to the minimum supported so the Go compiler prevents as much as possible incompatibilities when compiling with the older version.

PR #709, opened 2023-08-21
Update our OTel dependencies, and move config into GMP module

dashpole:
We need to switch from prometheus.BuildPromCompliantName to prometheus.BuildCompliantName, which takes an argument for enabling unit suffixes.  We would like to allow users to disable unit suffixes, as the (upstream) feature gate will not apply to unit suffixes anymore.

I decided to simplify our API surface in the GMP module to reduce the number of breaking changes we will need to make in the future.  To do this, I moved configuration related to naming, resource, and extra metrics into a config struct in this module.  **This will not result in any breaking changes upstream**, since it copies existing configuration from the GMP exporter for target and scope info.

This also simplifies up the ExtraMetrics function, since we don't need to have it return anything.


exporter/collector/googlemanagedprometheus/config.go
```diff
@@ -0,0 +1,47 @@
...
+// It is meant to be embedded in the googlemanagedprometheus configuration.
+type Config struct {
+	// AddMetricSuffixes controls whether suffixes are added to metric names. Defaults to true.
+	AddMetricSuffixes bool `mapstructure:"add_metric_suffixes"`
+	// ExtraMetricsConfig configures the target_info and otel_scope_info metrics.
+	ExtraMetricsConfig ExtraMetricsConfig `mapstructure:"extra_metrics_config"`
+}
+
+// ExtraMetricsConfig controls the inclusion of additional metrics.
+type ExtraMetricsConfig struct {
```
aabmass, line 27: Do you think it makes sense to  invert the meaning of these booleans so they are false by default i.e. `DisableTargetInfo` and `DisableScopeInfo`. Not sure if this matches the existing config for our `googlecloudexporter`, consistency would come before this concern.

Unless you think we will change the default at some point.

aabmass (approved):

exporter/collector/googlemanagedprometheus/config.go
dashpole: I don't think we will change the default at this point.  I generally prefer true == enabled in configs.

This matches the current config: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlemanagedprometheusexporter/config.go#L41, so i'd like to keep it that way

Issue #713, opened 2023-08-23
User Agent passed to SDK trace exporter is not respected.


dashpole:
The trace exporter always appends a fixed user agent to the trace exporter: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/858780dfcbce2ac418bf3fc5fe68b1cf860850d7/exporter/trace/trace.go#L45

This means if a user passes in a custom WithUserAgent option, it isn't respected.  This impacts the collector because it passes the user agent as an option to the trace exporter:  https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/858780dfcbce2ac418bf3fc5fe68b1cf860850d7/exporter/collector/config.go#L256

dashpole:
We should also add user-agent checks to fixture-based tests.

PR #714, opened 2023-08-23
Fix trace and logging user agent, and add tests

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/713.

I also discovered that the logging exporter was not overriding `{{version}}` in user agent.

This adds the user agent to the fixture tests, and adds a test to specifically check that the user_agent config is applied to outgoing requests.


internal/cloudmock/metrics.go
```diff
@@ -71,20 +82,26 @@ func (m *MetricsTestServer) CreateServiceTimeSeriesRequests() []*monitoringpb.Cr
...
 	defer m.mu.Unlock()
 	m.createMetricDescriptorReqs = append(m.createMetricDescriptorReqs, req)
 }
-func (m *MetricsTestServer) appendCreateTimeSeriesReq(req *monitoringpb.CreateTimeSeriesRequest) {
+func (m *MetricsTestServer) appendCreateTimeSeriesReq(ctx context.Context, req *monitoringpb.CreateTimeSeriesRequest) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.createTimeSeriesReqs = append(m.createTimeSeriesReqs, req)
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		m.userAgent = strings.Join(md.Get("User-Agent"), ";")
```
aabmass, line 95: Looks like this would overwrite UA across different requests. Should be OK, maybe we should raise an error if it would be overwritten with a new value

exporter/collector/integrationtest/testcases/testcases_metrics.go
```diff
@@ -317,6 +318,17 @@ var MetricsTestCases = []TestCase{
 		// SDK exporter does not support CreateServiceTimeSeries
 		SkipForSDK: true,
 	},
+	{
+		Name:                 "Custom User Agent",
+		OTLPInputFixturePath: "testdata/fixtures/metrics/counter.json",
+		ExpectFixturePath:    "testdata/fixtures/metrics/counter_user_agent_expect.json",
+		ConfigureCollector: func(cfg *collector.Config) {
+			cfg.UserAgent = "custom-user-agent"
```
aabmass, line 326: Should we test the `{{version}}` interpolation here as well?

internal/cloudmock/metrics.go
aabmass: Do we expect there to be multiple `User-Agent` MD values or just a single one pre-joined with `;`?

aabmass (approved):
💯

exporter/collector/integrationtest/testcases/testcases_metrics.go
dashpole: I initially had that, but it doesn't work with the SDK exporter.  I decided to remove it from this test, as it is already demonstrated in other tests.

internal/cloudmock/metrics.go
dashpole: We expect a single `User-Agent` MD value. I just wanted to make sure we would notice if we somehow got more than one.

I'll update it to append to the user-agent.  That way, we would just fail, and be able to see all the user-agents that were written

Issue #711, opened 2023-08-21
Remove g.co.* labels from trace export.


dashpole:
The new Cloud Trace UI uses OTel semantic conventions for service.name

We should remove our mapping for service.name to `g.co/gae/app/module`

dashpole:
Discussed with the team.  This isn't super-high priority, so lowering to p2.

dashpole:
This only impacts the go exporter: https://github.com/search?q=org%3AGoogleCloudPlatform%20g.co%2Fgae%2Fapp%2Fmodule&type=code

PR #715, opened 2023-08-23
remove service.name -> g.co/gae/app/module mapping from trace exporter's default mapping

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/711

 The UI now supports service.name.

This brings the go trace exporter in-line with trace exporters in other languages.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/traces/traces_basic_expected.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/traces/traces_basic_expected.json
@@ -47,11 +47,6 @@
                   "value": "opentelemetry-go 1.16.0; google-cloud-trace-exporter 1.19.1"
                 }
               },
-              "g.co/gae/app/module": {
-                "stringValue": {
-                  "value": "demo-server"
-                }
-              },
               "g.co/r/generic_node/location": {
                 "stringValue": {
                   "value": "global"
@@ -162,6 +157,11 @@
                   "value": "foo"
                 }
               },
+              "service.name": {
+                "stringValue": {
+                  "value": "demo-server"
+                }
+              },
               "telemetry.sdk.language": {
                 "stringValue": {
                   "value": "go"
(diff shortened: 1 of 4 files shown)


dashpole:
only service.name was added to the UI, so I need to revert the changes to http mappings.

dashpole:
Or maybe we don't need to do this at all...

aabmass (approved):
Is there any concern of this being a breaking change?

dashpole:
We expect users to prefer service.name, and don't expect many to be using g.co/gae/app/module.

AndreasBergmeier6176:
This actually IS a breaking change since `g.co/gae/app/module` still, at least until January 26th 2024, gets mapped to `Service` filter.
It would be ok, if the `Service` filter in GCP Trace would also filter `service.name`, but it does not!
Thus EVERY developer now needs to get taught to manually search by entering `service.name` because there is also no hinting for that name in the UI.
Maybe in the future it would be better to sync with the UI team to either improve UX or at least not to break it.

dashpole:
@AndreasBergmeier6176 sorry about that! I've raised it with the UI team, and will report back here when it is fixed. We did work with the UI team on this, but missed the issue you've encountered.  Please let me know if you find any other discrepancies caused by this change.

dashpole:
@AndreasBergmeier6176 the UI team got back to me, and it isn't an easy problem to fix because of limitations on the trace query language.  They said it will likely be fixed as part of broader changes planned for later this year.

In the meantime, you can revert back to the previous behavior with:

```golang
trace.New(
	trace.WithAttributeMapping(func(k attribute.Key) attribute.Key {
		switch k {
		case "http.path":
			return "/http/path"
		case "http.host":
			return "/http/host"
		case "http.method":
			return "/http/method"
		case "http.user_agent":
			return "/http/user_agent"
		case "http.status_code":
			return "/http/status_code"
		case "service.name":
			return "g.co/gae/app/module"
		}
		return k
	})
)
```

Again, very sorry about that.

PR #717, opened 2023-08-28
Revert to go 1.20 and prepare for v1.19.1/v0.43.1

dashpole:
See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/708#issuecomment-1695684872.  We should set the go version in our go.mod to the minimum version we support, which is go 1.20.

Also, prepare for a patch release.


dashpole:
cc @anuraaga

anuraaga:
Thanks for the quick fix! 🙏

punya (approved):

anuraaga:
Hi @dashpole - I randomly realized that opentelemetry-go-contrib targets 1.19 and depends on this library

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/go.mod

AFAIK, this isn't really allowed, basically if someone tries to compile that library with 1.19 it will fail due to the missing `errors.Join` I think. Another side effect I have found is that `go mod tidy` automatically raises to the ceiling of all dependencies, which is actually 1.21 currently since that is still pointing to the older release (I added `replace github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.19.0 => github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.19.1` to my project to fix that). 

My project is now on 1.20 so not personally affected but wanted to share as it looks like it could be a problem for other users of the opentelemetry-go-contrib library.

dashpole:
Ah, thanks.  I think we are technically OK, as the `detectors/gcp` module doesn't use `errors.Join`.  If it did, it would have failed presubmits when I bumped that in otel-go-contrib.

Contrib is using the patch release already, so you shouldn't need a replace: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/3ad5a2c545431824f851c62be1c9253a59427c07/detectors/gcp/go.mod#L7

PR #721, opened 2023-09-13
Update OTel dependencies

### LINKED ISSUE #719: Dependency conflict with latest metric SDK
When updating dependencies in my project, I found a conflict preventing bumping the OTel SDK to the latest due to a breaking change in the metrics SDK. This is somewhat expected since it's a 0.x library, but it's also nice if the code can be updated to resolve the conflict relatively quickly, but after digging at it, I found that due to the collector exporter's integration test, transitively this repo ends up being blocked on this

https://github.com/open-telemetry/opentelemetry-collector/pull/8299#issuecomment-1701892585

As it seems like it requires an opentelemetry-specification change, it seems like it could take a long time to resolve the dependency conflict. While no easy solution is obvious, I wanted to get visibility on the team in case there is any way to reduce the chance of a conflict being unresolved for a while. One idea is moving the code under `collector` into the opentelemetry-collector-contrib repo itself, naturally there will be thoughts on it re: code ownership, but it would allow the collector code to evolve at the pace of the collector, without affecting the SDK components in this repo which are integrated into user apps and have a larger impact surface for issues.

anuraaga:
Fixes #719


Makefile
```diff
@@ -49,6 +49,7 @@ tools: $(GOLANGCI_LINT) $(MISSPELL) $(GOCOVMERGE) $(STRINGER) $(GOJQ) $(FIELDALI
 
 PROTOBUF_VERSION = 3.19.0
 PROTOBUF_OS = linux
+UNAME_S := $(shell uname -s)
```
anuraaga, line 52: I noticed protoc wasn't working on my mac as this variable wasn't actually set to anything, I believe it was removed with some cleanup at some point. It could be nice to run the entire CI on multiple machines to ensure it continues to build on them, though the notable step is this one I think

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/.github/workflows/ci.yml#L41

exporter/collector/googlemanagedprometheus/naming_test.go
```diff
@@ -46,13 +46,25 @@ func TestGetMetricName(t *testing.T) {
...
 				sum := m.SetEmptySum()
 				sum.SetIsMonotonic(true)
 			},
+			expected: "foo_total/counter",
+		},
+		{
+			desc:     "sum without total is unaffected when normalization disabled",
+			baseName: "foo",
+			metric: func(m pmetric.Metric) {
+				//nolint:errcheck
```
anuraaga, line 62: The default for this feature gate changed, I tried to preserve tests with and without the change

exporter/collector/googlemanagedprometheus/naming_test.go
```diff
@@ -238,6 +278,10 @@ func TestGetMetricName(t *testing.T) {
 		},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
+			// Name translation uses global feature gate registry, set to defaults before each test.
```
anuraaga, line 281: This was missing before but seems to be important

exporter/collector/metrics_test.go
```diff
@@ -307,7 +307,7 @@ func TestHistogramPointToTimeSeries(t *testing.T) {
 	hdp := ts.Points[0].Value.GetDistributionValue()
 	assert.Equal(t, int64(15), hdp.Count)
 	assert.ElementsMatch(t, []int64{1, 2, 3, 4, 5}, hdp.BucketCounts)
-	assert.Equal(t, 12847.600000000002, hdp.SumOfSquaredDeviation)
+	assert.InDelta(t, 12847.6, hdp.SumOfSquaredDeviation, 0.0001)
```
anuraaga, line 310: Was failing for me on M1, I think this is an appropriate way of asserting on a floating point number that is going through rounding

exporter/metric/metric.go
```diff
@@ -155,7 +154,7 @@ func (me *metricExporter) Temporality(ik metric.InstrumentKind) metricdata.Tempo
 }
 
 // Aggregation returns the Aggregation to use for an instrument kind.
-func (me *metricExporter) Aggregation(ik metric.InstrumentKind) aggregation.Aggregation {
+func (me *metricExporter) Aggregation(ik metric.InstrumentKind) metric.Aggregation {
```
anuraaga, line 157: This is the only non-test change to work with the new version of the SDK

dashpole (approved):

Issue #761, opened 2023-10-31
Collector exporter does not log errors when requests fail


dashpole:
After removing retry_on_failure in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25900, we no longer get any log messages when we fail to send.

Discovered in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/760.

aabmass:
Looking at the original PR, its very weird that this removed logging. I would have expected the collector to log errors for you even without exporterhelper. Do you think there is an upstream collector bug here?

dashpole:
Opened https://github.com/open-telemetry/opentelemetry-collector/issues/8791 and https://github.com/open-telemetry/opentelemetry-collector/pull/8792.  We can hold off fixing it here until those are resolved.

dashpole:
Fixed by https://github.com/open-telemetry/opentelemetry-collector/pull/8792

PR #763, opened 2023-11-01
Fix issue where errors exporting from the collector are not logged

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/761

Rather than restore the retry_on_failure component, add logging in the library directly.  This has the added advantage of removing the "Try enabling retry_on_failure config option to retry on retryable errors" from the message.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -196,6 +196,17 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
 }
 
 func (l *LogsExporter) PushLogs(ctx context.Context, ld plog.Logs) error {
+	err := l.pushLogs(ctx, ld)
+	if err != nil {
+		l.obs.log.Error(
+			"Exporting logs failed",
+			zap.Error(err),
+		)
+	}
+	return err
+}
+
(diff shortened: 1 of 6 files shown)


exporter/collector/logs.go
```diff
@@ -196,6 +196,17 @@ func (l *LogsExporter) Shutdown(ctx context.Context) error {
 }
 
 func (l *LogsExporter) PushLogs(ctx context.Context, ld plog.Logs) error {
+	err := l.pushLogs(ctx, ld)
+	if err != nil {
+		l.obs.log.Error(
+			"Exporting logs failed",
+			zap.Error(err),
+		)
+	}
```
aabmass, line 205: Should we make a generic wrapper for this kind of thing? Similar to how it worked with retry helper before

dashpole: We could write a util function, but it would need access to each logger anyways.  Without it being built into the component framework (e.g. like exporterhelper), I don't think we can write a generic wrapper.

dashpole:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/8792

Issue #735, opened 2023-10-11
Map faas.* attributes to generic_task in resource mapping 


dashpole:
Implement https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/273 in the resource mapping in this repository.

dashpole:
You are correct.  That was definitely an oversight during https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/679.  I'll make sure that is fixed as well.  Thanks for pointing it out.

PR #764, opened 2023-11-01
Map faas.* attributes to generic task after service.name, but ignore unknown_service

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/735

Also, fixes the mapping logic to use faas.instance instead of faas.id


dashpole:
@aabmass

exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -69,9 +72,24 @@ func (c Config) MapToPrometheusTarget(res pcommon.Resource) *monitoredrespb.Moni
 // getStringOrEmpty returns the value of the first key found, or the empty string.
 func getStringOrEmpty(attributes pcommon.Map, keys ...string) string {
 	for _, k := range keys {
-		if val, ok := attributes.Get(k); ok {
+		if val, ok := attributes.Get(k); ok && !strings.HasPrefix(val.Str(), unknownServicePrefix) {
```
aabmass, line 75: nit can you add a comment explaining why we skip here?

aabmass (approved):

exporter/collector/googlemanagedprometheus/monitoredresource.go
dashpole: added

Issue #765, opened 2023-11-01
Collector Metric exporter will continue to attempt to send MetricDescriptors even if permissions are unavailable.


jsuereth:
Today, the [code to create metric descriptors](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/metrics.go#L671) is not resolving "recoverable" errors from "non-recoverable" errors.  If a service account does not have access to create metric descriptors, then the collector will attempt to create new metric descriptors every time that metric is encountered, not just once-per-instantiation.

Suggest the following:

- The err handling logic is updated so that any 4xx HTTP (or unrecoverable gRPC) error leads to updating the cache, preventing further CMD calls.
- If necessary, we update CreateMetricDescriptor configuration to allow configuring the setting for how to deal with failures.

PR #768, opened 2023-11-09
Metrics exporter cache requests that fail with an unrecoverable error

avilevy18:
Fixes [#765](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/765)

When the metrics exporter fails to create a new metric descriptor any non-recoverable grpc errors will update the cache, preventing further CMD calls.

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -661,6 +661,7 @@ func projectName(projectID string) string {
 	return fmt.Sprintf("projects/%s", projectID)
 }
 
+// isNotRecoverable returns true if the error is permanent.
 func isNotRecoverable(err error) bool {
 	s := status.Convert(err)
 	return !(s.Code() == codes.DeadlineExceeded || s.Code() == codes.Unavailable)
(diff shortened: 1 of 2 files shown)


punya (changes requested):
Please add some test coverage

dashpole:
@avilevy18 I'm planning to cut a release later today.  If you can add test coverage by then, then we can include it.  Let me know if we can help direct you.

dashpole:
you can run `make fieldalignment` to fix the error

exporter/collector/metrics.go
```diff
@@ -137,6 +137,13 @@ const (
 
 type labels map[string]string
 
+type monitoringClient interface {
```
dashpole, line 140:
```suggestion
// monitoringClient is the subset of monitoring.MetricClient this exporter uses,
// and allows us to mock the implementation for testing.
type monitoringClient interface {
```

exporter/collector/metrics.go
```diff
@@ -655,6 +661,11 @@ func projectName(projectID string) string {
 	return fmt.Sprintf("projects/%s", projectID)
 }
 
+func isNotRecoverable(err error) bool {
```
dashpole, line 664:
```suggestion
// isNotRecoverable returns true if the error is permanent.
func isNotRecoverable(err error) bool {
```

dashpole (approved):

PR #769, opened 2023-11-13
Rename untyped featuregate to comply with feature gate character set restrictions

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29158


dashpole:
cc @ridwanmsharif

aabmass (approved):

Issue #734, opened 2023-10-09
[security] audit repository tooling


dashpole:
Even though this repository isn't in the OTel github organization, we should still try to follow best practices used by the rest of the go community: https://github.com/open-telemetry/opentelemetry-go/issues/4459.

We should look at the list of recommendations, determine what we need to change, and make updates to this repository if possible.

- [ ] CodeQL enabled via [GitHub Actions](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/codeql-analysis.yml)
- [x] Static code analysis tool (the collector uses govulncheck [https://pkg.go.dev/golang.org/x/vuln] on every build)
- [ ] Repository security settings
    - [ ] Security Policy ✅
    - [x] Security advisories ✅
    - [x] Private vulnerability reporting ✅
    - [x] Dependabot alerts ✅
    - [x] Code scanning alerts ✅

aabmass:
David to look into this next sprint

dashpole:
codeql is not using github actions.  I need to look into that.

PR #772, opened 2023-11-21
Add security policy

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/734

Based on https://github.com/open-telemetry/opentelemetry-go/security/policy

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
```
aabmass, line 3: Also see the google OSS policy https://opensource.google/documentation/policies/library-breaking-change. There is a major version backport policy

aabmass (approved):
Otherwise LGTM

SECURITY.md
dashpole: updated

PR #775, opened 2023-11-23
Add support for CLOUD_SDK_ACCESS_TOKEN_PATH

alyshanjahani-crl:
This commit adds support for configuring the collector to authenticate with a token source that comes from a file defined by the CLOUD_SDK_ACCESS_TOKEN_PATH environment variable.

When the token expires, the token source will refresh by reading the file again. Optionally the user can set CLOUD_SDK_ACCESS_TOKEN_EARLY_EXPIRY to control when the refresh occurs.

The gcloud CLI has support for authenticating with user-supplied access tokens. See: https://issuetracker.google.com/issues/134539464#comment12

This is useful when credentials need to be generated dynamically from some centralized system that passes these scoped short-lived credentials to untrusted systems which use them to authenticate with GCP services (instead of having the untrusted system generate the credentals with some base/root credentials).

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -16,8 +16,10 @@ package collector
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


dashpole:
I don't think we should include GCP auth-related code in our OTel exporters.  I would recommend opening a feature request with https://github.com/googleapis/google-cloud-go.  If it is accepted, then we will support it once we update the client library.

PR #776, opened 2023-11-28
Provide useful cluster names for non k8s environments on managed prom…

jsuereth:
…etheus exporter.

- we use `__run__` on cloud run and `__gce__` on compute engine.
- This matches the documentation for cloud run and support in Ops Agent.
- This can be controlled via a configuration parameter.

cc @ridwanmsharif 

Note: This just modifies the defaults, happy to provide docs or transition period.

### DIFF
--- a/exporter/collector/googlemanagedprometheus/config.go
+++ b/exporter/collector/googlemanagedprometheus/config.go
@@ -21,9 +21,6 @@ type Config struct {
 	AddMetricSuffixes bool `mapstructure:"add_metric_suffixes"`
 	// ExtraMetricsConfig configures the target_info and otel_scope_info metrics.
 	ExtraMetricsConfig ExtraMetricsConfig `mapstructure:"extra_metrics_config"`
-	// UseEnvironmentClusterNames controls whether cluster names "__run__" and "__gce__" are used on
-	// these environments.
-	UseEnvironmentClusterNames bool `mapstructure:"use_environment_cluster_names"`
 }
 
 // ExtraMetricsConfig controls the inclusion of additional metrics.
@@ -41,7 +38,6 @@ func DefaultConfig() Config {
 			EnableTargetInfo: true,
 			EnableScopeInfo:  true,
 		},
-		UseEnvironmentClusterNames: true,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -48,6 +48,28 @@ var promTargetKeys = map[string][]string{
...
+
+	cloudPlatform := getStringOrEmpty(attrs, semconv.AttributeCloudPlatform)
+	switch cloudPlatform {
+	case semconv.AttributeCloudPlatformGCPComputeEngine:
+		return "__gce__"
+	case semconv.AttributeCloudPlatformGCPCloudRun:
+		return "__run__"
+	default:
+		// No special rules, look up cluster as normal.
+		return getStringOrEmpty(attrs, promTargetKeys[clusterLabel]...)
```
dashpole, line 69: can this be the primary lookup?  i.e. if a user has set the `cluster` resource attribute manually, it would still respected.  Then we can fall back to the platform-based cluster names if the primary keys are not found.

exporter/collector/googlemanagedprometheus/config.go
```diff
@@ -21,6 +21,9 @@ type Config struct {
 	AddMetricSuffixes bool `mapstructure:"add_metric_suffixes"`
 	// ExtraMetricsConfig configures the target_info and otel_scope_info metrics.
 	ExtraMetricsConfig ExtraMetricsConfig `mapstructure:"extra_metrics_config"`
+	// UseEnvironmentClusterNames controls whether cluster names "__run__" and "__gce__" are used on
+	// these environments.
+	UseEnvironmentClusterNames bool `mapstructure:"use_environment_cluster_names"`
```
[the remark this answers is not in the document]
jsuereth: Done.

exporter/collector/googlemanagedprometheus/monitoredresource.go
jsuereth: Done, did a slight refactor to the `getStringOr*` methods.

dashpole (approved):

PR #777, opened 2023-12-11
exporter/metric: Add support to export Distribution with Exemplars and Exponential Histograms.

franciscovalentecastro:
This PR adds the following functionality to the `exporter/metric` : 
- Added support to export `ExponentialHistogram`.
- Added support to export `TypedValue` [Distribution](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#distribution) (either `Histogram` and `ExponentialHistogram`) with exemplars.
- Adds `exponential_histogram` example with a `dashboard.json` to visualize it.

Details : 
- https://github.com/open-telemetry/opentelemetry-go/pull/3849 added the `Exemplar` data type and the `Exemplars` field to a `HistogramDataPoint`. 
- https://github.com/open-telemetry/opentelemetry-go/pull/4871 added experimental `Exemplar` sampling support.


dashpole:
The feature has been implemented in the OTel SDK: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/metric/internal/x/README.md#exemplars

Let me know if you are interested in contributing this.

exporter/metric/metric.go
```diff
@@ -643,8 +677,49 @@ func histToDistribution[N int64 | float64](hist metricdata.HistogramDataPoint[N]
...
+			},
+		},
+		Exemplars: toDistributionExemplar[N](hist.Exemplars),
+	}
+}
+
+func toDistributionExemplar[N int64 | float64](Exemplars []metricdata.Exemplar[N]) []*distribution.Distribution_Exemplar {
+	var exemplars []*distribution.Distribution_Exemplar
+	for _, e := range Exemplars {
+		exemplars = append(exemplars, &distribution.Distribution_Exemplar{Value: float64(e.Value), Timestamp: timestamppb.New(e.Time)})
```
franciscovalentecastro, line 717: Verified `Distribution_Exemplar` only supports `float64` values : https://pkg.go.dev/google.golang.org/genproto/googleapis/api/distribution#Distribution_Exemplar

exporter/metric/metric.go
```diff
@@ -643,8 +677,49 @@ func histToDistribution[N int64 | float64](hist metricdata.HistogramDataPoint[N]
...
+	}
+}
+
+func expHistToDistribution[N int64 | float64](hist metricdata.ExponentialHistogramDataPoint[N]) *distribution.Distribution {
+	numFiniteBuckets := len(hist.PositiveBucket.Counts)
+	counts := make([]int64, numFiniteBuckets)
+	positiveBucketCounts := 0
+	growthFactor := math.Exp2(math.Exp2(-float64(hist.Scale)))
+	for i, v := range hist.PositiveBucket.Counts {
+		counts[i] = int64(v)
```
franciscovalentecastro, line 690: [Opentelemetry Exponential Buckets](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exponential-buckets) support `Negative Indices` too, but AFAIU the [TypedValue Exponential Buckets](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#exponential) only support `Positive Indices`.

example/metric/exponential_histogram/example.go
```diff
@@ -0,0 +1,322 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+)
+
+func workloadMetricsPrefixFormatter(d metricdata.Metrics) string {
+	return fmt.Sprintf("custom.googleapis.com/%s", d.Name)
```
dashpole, line 42: does the example need to write to custom.googleapis.com?  Did workload.googleapis.com not work?

example/metric/exponential_histogram/example.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// Create a meter with which we will record metrics for our package.
+	meter := provider.Meter("github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric")
+
+	_, err = meter.Float64ObservableGauge(
+		"latency",
+		metric.WithDescription(
+			"Latency",
+		),
+		metric.WithUnit("ms"),
```
dashpole, line 219: nit: prefer using seconds for durations in samples

exporter/metric/metric.go
dashpole, line 690: I think this might need to be increased by 1.  See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/87124ac8635698cc7cdc9f8f2dfca94a8aa13606/exporter/collector/metrics.go#L998-L1001

dashpole: Thats correct.  All negative observations and zero observations should be in the underflow bucket.  See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/87124ac8635698cc7cdc9f8f2dfca94a8aa13606/exporter/collector/metrics.go#L989-L993

exporter/metric/metric.go
```diff
@@ -643,8 +677,49 @@ func histToDistribution[N int64 | float64](hist metricdata.HistogramDataPoint[N]
...
+	return &distribution.Distribution{
+		Count:        int64(positiveBucketCounts),
+		Mean:         mean,
+		BucketCounts: counts,
+		BucketOptions: &distribution.Distribution_BucketOptions{
+			Options: &distribution.Distribution_BucketOptions_ExponentialBuckets{
+				ExponentialBuckets: &distribution.Distribution_BucketOptions_Exponential{
+					Scale:            math.Pow(growthFactor, float64(hist.PositiveBucket.Offset)),
+					GrowthFactor:     growthFactor,
+					NumFiniteBuckets: int32(numFiniteBuckets),
```
dashpole, line 706: Looks like the collector exporter also sends an underflow and overflow buckets: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/87124ac8635698cc7cdc9f8f2dfca94a8aa13606/exporter/collector/metrics.go#L1022

example/metric/exponential_histogram/example.go
franciscovalentecastro: I don't think it "needs" to write to `custom.googleapis.com`. I remember trying to use `workload.googleapis.com` and then not seeing the metrics, but this probably was another bug during development. I never tried `workload.googleapis.com` again.

I will try again with `workload`, is there a preference ? I don't really know the main difference.

example/metric/exponential_histogram/example.go
franciscovalentecastro: Is there a functionality difference between using `s` or `ms` ? 

For this specific example, i chose `ms` since the generated data ("Latency" Log-Normal Distribution) has a mean of around ~ `30`. So, I think it seems more realistic `30ms` of server delay over `30s`. Though, this is purely an example, so at the end there is no real difference.

exporter/metric/metric.go
franciscovalentecastro: Fixed. I used as reference the implementation of `collector/metrics.go : exponentialHistogramPoint` to reimplement all the mentioned details of `expHistToDistribution` in this review, but using the correct types and methods.

exporter/metric/metric.go
franciscovalentecastro: Fixed. I used as reference the implementation of `collector/metrics.go : exponentialHistogramPoint` to reimplement all the mentioned details of `expHistToDistribution` in this review, but using the correct types and methods.

example/metric/exponential_histogram/example.go
franciscovalentecastro: Used `workload.googleapis.com` and it worked fine 👍 . Fixed.

exporter/metric/metric.go
franciscovalentecastro: Fixed. I used as reference the implementation of collector/metrics.go : exponentialHistogramPoint to reimplement all the mentioned details of expHistToDistribution in this review, but using the correct types and methods.

franciscovalentecastro:
@dashpole Update on last changes to PR :
- Added `dashboard.json` to visualize the exponential_histogram example. This may conflict/be confused with the other `dashboard.json` in the above. Is this the correct place and/or naming ? (I imported manually and i work correctly).
- Fixed most of the open comments.
- Added a `TODO` for the missing `[]attachments` to the exemplar exporting (AFAIU, this will add the span_id, etc, though i'm not very knowledgable about how to do this correctly).

example/metric/exponential_histogram/example.go
```diff
@@ -0,0 +1,322 @@
...
+		),
+	)
+	if err != nil {
+		log.Fatalf("resource.New: %v", err)
+	}
+
+	// Create exporter pipeline
+	exporter, err := mexporter.New(
+		mexporter.WithMetricDescriptorTypeFormatter(workloadMetricsPrefixFormatter),
+	)
```
dashpole, line 63:
```suggestion
	exporter, err := mexporter.New()
```
The workload metrics prefix is already the default.

example/metric/exponential_histogram/example.go
```diff
@@ -0,0 +1,322 @@
...
+			Name: "latency_multimodal_c",
+			Aggregation: sdkmetric.AggregationExplicitBucketHistogram{
+				Boundaries: defaultBuckets,
+			},
+		},
+	)
+
+	// initialize a MeterProvider with that periodically exports to the GCP exporter.
+	provider := sdkmetric.NewMeterProvider(
+		sdkmetric.WithView(viewLatencyA),
```
dashpole, line 188: nit: WithView can take more than 1 view.  You can pass all views in a single option

exporter/metric/metric.go
```diff
@@ -505,6 +506,16 @@ func (me *metricExporter) recordToTspb(m metricdata.Metrics, mr *monitoredrespb.
 			ts.Metric = me.recordToMpb(m, point.Attributes, library, extraLabels)
 			tss = append(tss, ts)
 		}
+	case metricdata.ExponentialHistogram[float64]:
```
dashpole, line 509: Based on https://github.com/open-telemetry/opentelemetry-go/blob/a7034da63185be247206ed9539d8ab93d982d7e3/exporters/otlp/otlpmetric/otlpmetricgrpc/internal/transform/metricdata.go#L218, it looks like the only difference between int64 and float64 exponential histograms is the sum.  Can we accept int64 exponential histograms as well, and just cast the sum to a float64?

dashpole: Actually, it looks like you already handle int64 exponential histograms below.  I think you just need to handle the `case metricdata.ExponentialHistogram[int64]:` here.

example/metric/exponential_histogram/example.go
franciscovalentecastro: Fixed

exporter/metric/metric.go
franciscovalentecastro: Added `case` for `int64` type. Fixed.

dashpole (approved):

dashpole:
Follow-up tasks required:

* README for the example
* Unit testing
* Exemplar attachments
* SumOfSquaredDeviations

@franciscovalentecastro can you add the README?  You are welcome to work on the other tasks as well if you have time.  Otherwise, I'll make sure they get done.  Just let me know which you'd like to tackle

PR #788, opened 2023-12-19
Fixes #778 - Exemplars should be sorted before sending to Cloud Ops.

### LINKED ISSUE #778: Exemplars should be sorted in Collector exporter
@franciscovalentecastro discovered in #777 that we need to make sure exemplars are sorted when sending to Cloud Monitoring or the points will be rejected.

We've been lucky so far, as most exemplars have come from explicit bucket histograms, where they are implicitly sorted.

jsuereth:
Fixes #778 

We discovered that using the fixed-sized reservoir does not guarantee exemplar order, but this is required in Cloud Operations for most distribution types.

cc @franciscovalentecastro

### DIFF
--- a/exporter/collector/metrics.go
+++ b/exporter/collector/metrics.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"path"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


damemi (approved):

damemi:
codecov failure is https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/790

PR #793, opened 2024-01-09
Update OTel dependencies for go workspaces

braydonk:
This PR updates all OTel dependencies to the latest version to allow this repo to leverage go workspaces without the ambiguous import problem with `github.com/knadh/koanf`.

To produce these changes I did the following:

```bash
go work init
go work use -r .
go mod edit -exclude=github.com/knadh/koanf@v1.5.0
go work sync
```

I then updated the OTel versions in the Makefile, and added a couple OTel dependencies that weren't listed in the `update-otel` target. Then ran `make update-otel`.

I also added a new make target that will generate the Go module workspace.


dashpole (approved):
is there any action required for developers to use go workspaces?

damemi:
Thanks for making the workspaces changes @braydonk, especially figuring out that `knadh/koanf` issue. Was that `go mod exclude` supposed to be committed or part of the `make go-work` command you added?

I looked a little into workspaces and it sounds like `go.work` files shouldn't be committed, basically because they would then silently override dependencies for everyone. But +1 to adding a makefile step to set it up more easily

Issue #792, opened 2024-01-05
exporter/collector/logs: segfault when setting `gcp.source_location`


quentinmit:
It looks like this is coming from

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/02b243a6fbedbdc6118f2e2dc34aced98fc70dba/exporter/collector/logs.go#L397

which calls `sourceLocation.Bytes().AsRaw()` without confirming that the attribute has a bytes value.

## Stacktrace
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xce2fbc]

goroutine 127 [running]:
go.opentelemetry.io/collector/pdata/pcommon.ByteSlice.AsRaw(...)
        /root/go/pkg/mod/go.opentelemetry.io/collector/pdata@v1.0.0-rcv0016/pcommon/generated_byteslice.go:37
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector.logMapper.logToSplitEntries({{0xc0019b22a0}, {{0xc00155a030, 0xa}, {0xc001d1a810, 0x26}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, ...}}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/!google!cloud!platform/opentelemetry-operations-go/exporter/collector@v0.44.0/logs.go:397 +0x47c
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector.logMapper.createEntries({{0xc0019b22a0}, {{0xc00155a030, 0xa}, {0xc001d1a810, 0x26}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, ...}}, ...}, ...}, ...)
        /root/go/pkg/mod/github.com/!google!cloud!platform/opentelemetry-operations-go/exporter/collector@v0.44.0/logs.go:282 +0x8d4
github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector.(*LogsExporter).PushLogs(0xc001eb4800, {0x5b60de0, 0xc001f9c930}, {0xc001b8e468?, 0xc001f9a2b0?})
        /root/go/pkg/mod/github.com/!google!cloud!platform/opentelemetry-operations-go/exporter/collector@v0.44.0/logs.go:199 +0xca
[...]

PR #794, opened 2024-01-09
Support gcp.source_location being set as other types

braydonk:
By copying the logic from parsing gcp.http_request, this PR adds support to also parse `gcp.source_location` from types other than Bytes.

This fixes the basic case of #792 such that the collector does not panic when setting `gcp.source_location` as something other than bytes.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -124,6 +124,27 @@ var otelSeverityForText = map[string]plog.SeverityNumber{
 	"fatal4": plog.SeverityNumberFatal4,
 }
 
+type AttributeProcessingError struct {
+	Err error
+	Key string
+}
+
+func (e *AttributeProcessingError) Error() string {
+	return fmt.Sprintf("could not process attribute %s: %s", e.Key, e.Err.Error())
+}
+
+func (e *AttributeProcessingError) Unwrap() error {
+	return e.Err
+}
+
+type UnsupportedValueTypeError struct {
+	ValueType pcommon.ValueType
+}
+
+func (e *UnsupportedValueTypeError) Error() string {
+	return fmt.Sprintf("unsupported value type %v", e.ValueType)
+}
+
 type LogsExporter struct {
 	obs           selfObservability
 	loggingClient *loggingv2.Client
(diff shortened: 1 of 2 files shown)


exporter/collector/logs.go
```diff
@@ -393,8 +393,15 @@ func (l logMapper) logToSplitEntries(
 
 	// parse LogEntrySourceLocation struct from OTel attribute
 	if sourceLocation, ok := attrsMap[SourceLocationAttributeKey]; ok {
+		var sourceLocationBytes []byte
```
quentinmit, line 396: Maybe worth refactoring this into a common function that's used by both call sites? (Especially since it's going to be needed a third time to support `operation`)

exporter/collector/logs_test.go
```diff
@@ -396,6 +395,103 @@ func TestLogMapping(t *testing.T) {
 			},
 			maxEntrySize: defaultMaxEntrySize,
 		},
+		{
+			name: "log with bad source location (bytes)",
```
quentinmit, line 399: What makes this test "bad"?

exporter/collector/logs.go
braydonk: Done in [a0f37b2](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/794/commits/a0f37b2a6901f8c2529db678a696f414e0a6bf5b)

exporter/collector/logs_test.go
braydonk: It's because the value provided for `gcp.source_location` can't actually be parsed, the json Unmarshal fails due to an invalid type in one of the fields. Is there a nicer way I should say that concisely in the test or should I add a comment?

exporter/collector/logs.go
```diff
@@ -124,6 +124,27 @@ var otelSeverityForText = map[string]plog.SeverityNumber{
 	"fatal4": plog.SeverityNumberFatal4,
 }
 
+type AttributeProcessingError struct {
```
dashpole, line 127: Does this need to be public?  Same question for other error below.

braydonk: I tend to make these public out of habit so that they can be used in test assertions, but since we don't use the `_test` package for these tests I can make these private.

braydonk: Done in [4b7052e](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/794/commits/4b7052e0d2962cb46c613a4f1e56d431b7ad894e)

dashpole (approved):

dashpole:
@quentinmit i'll wait for your approval to merge

exporter/collector/logs.go
```diff
@@ -631,3 +652,17 @@ func fixUTF8(s string) string {
 	}
 	return buf.String()
 }
+
+func bytesFromValue(v pcommon.Value) ([]byte, error) {
```
quentinmit, line 656: Why not put the `json.Unmarshal` in the function too? Then the signature can just be `unmarshalAttribute(v pcommon.Value, out any) error`

quentinmit (approved):
See comment, but otherwise this LGTM

exporter/collector/logs_test.go
quentinmit: `"log with wrong type in sourceLocation (bytes)"`, perhaps?

exporter/collector/logs.go
braydonk: Done in [992e990](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/794/commits/992e99041c6f46520f344b0da0cb006b98b972ad)

braydonk:
Let's see if the e2e tests will work today :pray:

braydonk:
I still don't seem to have permission to run these :(

Issue #789, opened 2023-12-20
Internal server error when location field not set


jsirianni:
This is a follow up for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/760. The issue with logs being silently dropped has been resolved and has exposed the real issue I am encountering.

When sending a `k8s_container` metric to Cloud Monitoring, I get an "internal server error". It is resolved by setting `cloud.availability_zone` (or the region resource attribute).

With logging, I can get by without setting the location field. It seems that this should be the case with metrics as well. My cluster is not part of a cloud and has no reasonable region / zone value. This is also true for the GCP customers that we (observiq) support. They have on premise clusters.

This is the error text, formatted to be easily readable:
```
One or more TimeSeries could not be written: Internal error encountered. 
Please retry after a few seconds. If internal errors persist, contact support at https://cloud.google.com/support/docs.: 

k8s_container{
    container_name:coredns,
    location:,
    pod_name:coredns-787d4945fb-r9fk5,
    namespace_name:kube-system,
    cluster_name:stage
} 

timeSeries[0]: workload.googleapis.com/core.log.count{
    k8s_pod_name:coredns-787d4945fb-r9fk5,
    k8s_node_name:minikube,
    k8s_cluster_name:stage,
    host_name:bindplane-gateway-agent-2,
    instrumentation_source:logcount,
    k8s_container_name:coredns,
    k8s_namespace_name:kube-system
}

error details: name = Unknown  desc = total_point_count:1 errors:{status:{code:13} point_count:1}
```

The error's metric descriptor does make it clear that the location field is unset.

This is the log from the collector (will the error value removed.
```json
{
  "level":"error",
  "ts":"2023-12-20T16:41:34.815Z",
  "caller":"exporterhelper/queue_sender.go:123",
  "msg":"Exporting failed. No more retries left. Dropping data.",
  "kind":"exporter",
  "data_type":"metrics",
  "name":"googlecloud/jsirianni-12202023-temp",
  "error":"<see error>",
  "dropped_items":1,
[...]

PR #795, opened 2024-01-10
Fall back to global for all monitored resources that accept it

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/789

Tested each monitored resource manually.  For each where I added a global fallback, the monitoring API will reject requests with an empty location, but accept requests with a "global" location.

### DIFF
--- a/internal/resourcemapping/resourcemapping.go
+++ b/internal/resourcemapping/resourcemapping.go
@@ -72,57 +72,77 @@ var (
 		otelKeys []string
 	}{
 		gceInstance: {
-			zone:       {otelKeys: []string{string(semconv.CloudAvailabilityZoneKey)}},
+			zone: {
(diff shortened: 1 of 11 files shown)


internal/resourcemapping/resourcemapping.go
aabmass, line 1: If it's not too much work, it would be good to add a metric fixture test for these fallbacks since it will be tested against the real Monitoring API. Maybe a minimal OTel Resource for each MR that uses as many fallbacks as possible

aabmass (approved):

dashpole:
Closing based on https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/789#issuecomment-1896633989

PR #800, opened 2024-02-07
Add gRPC examples with OTLP SDK

damemi:
Uses our basic examples with gRPC application default credential functions to send OTLP


punya (approved):

example/metric/otlpgrpc/example.go
```diff
@@ -0,0 +1,83 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	creds, err := oauth.NewApplicationDefault(ctx)
+	if err != nil {
+		panic(err)
+	}
+
+	res, err := resource.New(
```
dashpole, line 40: I would consider omitting resource to keep the sample as simple as possible, given the point is to validate that creds can be used properly.

example/trace/otlpgrpc/client/client.go
```diff
@@ -0,0 +1,125 @@
...
+	otel.SetTracerProvider(tp)
+	return func() {
+		err := tp.Shutdown(context.Background())
+		if err != nil {
+			fmt.Printf("error shutting down trace provider: %+v", err)
+		}
+	}, nil
+}
+
+func installPropagators() {
```
dashpole, line 69: consider omitting propagators for the same reason.

example/trace/otlpgrpc/client/client.go
```diff
@@ -0,0 +1,125 @@
...
+func main() {
+	installPropagators()
+	shutdown, err := initTracer()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer shutdown()
+	tr := otel.Tracer("cloudtrace/example/client")
+
+	client := http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
```
dashpole, line 89: Any reason why you are using otelhttp here, rather than just sending a simple span?

.gitignore
```diff
@@ -13,5 +13,6 @@ coverage.*
 /example/metric/collector/collector
 /example/metric/sdk/metrics
 /example/metric/sdk/sdk
+example/metric/otlpgrpc/otlpgrpc
```
dashpole, line 16: add the trace binary here as well

dashpole (commented):
any reason why the trace sample is in a `client` subdirectory, but the metric one isn't?

example/trace/otlpgrpc/client/client.go
damemi: was just copying from the existing example. I simplified it down now

dashpole (approved):

PR #803, opened 2024-02-16
Update OTLP auth example

damemi:
Add `x-goog-user-project` header and set `gcp.project_id` (not sure if that needs to be a span attribute or resource attribute, so I'm setting both for now)

Repro steps:
* `gcloud auth application-default login`
* `export OTEL_EXPORTER_OTLP_ENDPOINT=otlp://<endpoint>`
* `export GCLOUD_PROJECT=<project id>`
* `go run example.go`


aabmass (approved):

damemi:
Updated this to add an http sample too, going to merge if everything is green

example/trace/otlpgrpc/example.go
```diff
@@ -31,15 +34,18 @@ import (
...
 	creds, err := oauth.NewApplicationDefault(ctx)
 	if err != nil {
 		panic(err)
 	}
 
-	exporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithDialOption(grpc.WithPerRPCCredentials(creds)))
+	res := resource.NewWithAttributes(semconv.SchemaURL, attribute.String("gcp.project_id", projectID))
+
+	// set endpoint with OTEL_EXPORTER_OTLP_ENDPOINT=https://<endpoint>
+	exporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithDialOption(grpc.WithPerRPCCredentials(creds)), otlptracegrpc.WithHeaders(map[string]string{"x-goog-user-project": projectID}))
```
dashpole, line 48: Should we tell users to use OTEL_EXPORTER_OTLP_HEADERS instead?

example/trace/otlpgrpc/example.go
```diff
@@ -31,15 +34,18 @@ import (
...
+func initTracer(projectID string) (func(), error) {
 	ctx := context.Background()
 
 	creds, err := oauth.NewApplicationDefault(ctx)
 	if err != nil {
 		panic(err)
 	}
 
-	exporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithDialOption(grpc.WithPerRPCCredentials(creds)))
+	res := resource.NewWithAttributes(semconv.SchemaURL, attribute.String("gcp.project_id", projectID))
```
dashpole, line 45: Can we use OTEL_RESOURCE_ATTRIBUTES instead?

example/trace/otlpgrpc/example.go
damemi: This took some digging, but it looks like `WithHeaders` overwrites OTEL_EXPORTER_OTLP_HEADERS (couldn't find docs but I found a [test asserting it](https://github.com/open-telemetry/opentelemetry-go/blob/e6e186bfa485f679e35bb775cba63ca24029590d/exporters/otlp/otlpmetric/otlpmetrichttp/internal/oconf/options_test.go#L333-L341), and I couldn't get it to work with both)

So if we pass Authorization using WithHeaders, we also need the other headers that way too. Unless they can be merged somehow

example/trace/otlpgrpc/example.go
damemi: yup, updated

PR #810, opened 2024-03-06
Support exemplars from integer instruments

dashpole:
We weren't properly checking the value type of exemplars, meaning we were sending a value of 0 for all int exemplars.


ymotongpoo:
Thank you for this fix. I try how this works.

PR #814, opened 2024-03-06
Make client auth extension public and implement component.Component

damemi:
Working on contributing this upstream (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31518), we will need to export some of the functions like CreateExtension and CreateDefaultConfig. This also implements the `component.Component` interface so the extension can be created without calling `google.FindDefaultCredentials()`, which can result in an immediate error if the credentials are not found. This enables lifecycle tests upstream.

### DIFF
--- a/extension/googleclientauthextension/factory_test.go
+++ b/extension/googleclientauthextension/factory_test.go
@@ -41,3 +41,21 @@ func TestCreateExtension(t *testing.T) {
 	assert.NotNil(t, ext)
 	assert.NoError(t, err)
 }
+
+func TestStart(t *testing.T) {
+	t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", "testdata/fake_creds.json")
+	ext, err := NewFactory().CreateExtension(context.Background(), extension.CreateSettings{}, CreateDefaultConfig())
+	assert.NotNil(t, ext)
+	assert.NoError(t, err)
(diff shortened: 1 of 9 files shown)


dashpole:
Given how small this library is, should the entire thing just live upstream?

damemi:
@dashpole updated this like we talked about offline, ready for review now

extension/googleclientauthextension/factory_test.go
```diff
@@ -37,7 +37,25 @@ func TestNewFactory(t *testing.T) {
 
 func TestCreateExtension(t *testing.T) {
```
dashpole, line 38: Can you also test that we don't panic if the extension is used (e.g. PerRPCCredentials or RoundTrip is invoked) before the component is started?  Might also be good to check that Start can be invoked more than once.

damemi: Added checks for RoundTrip, RoundTripper, and PerRPCCredentials. I couldn't get a basic test working for GetRequestMetadata because that seems to need some more involved gRPC calls. Do you think we need a test for GetRequestMetadata if PerRPCCredentials is already tested?

dashpole: Thats fine

dashpole (approved):

Issue #804, opened 2024-03-04
README for exponential histogram example


dashpole:
Follow-up to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/777

Add a README to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/example/metric/exponential_histogram

It should include:

* Instructions to run the application, and create the dashboard
* An explanation of the differences between the various histograms created by the application

PR #821, opened 2024-03-20
example/metric : Add exponential_histogram README and reorganize folders.

franciscovalentecastro:
Closes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/804


damemi (commented):
@franciscovalentecastro I see this is a draft but just curious, why move the existing examples into a new `export` folder? Couldn't the `exponential_histogram` example folder just go under the existing examples directory with them?

franciscovalentecastro:
> @franciscovalentecastro I see this is a draft but just curious, why move the existing examples into a new `export` folder? Couldn't the `exponential_histogram` example folder just go under the existing examples directory with them?

Re @damemi, I was writing a README.md for the `exponential_histogram` with the realization that most steps were very similar to what the other examples (`collector`, `otlpgrpc`, `spk`) require to run. There were two options : 

1. Add all steps to run `exponential_histogram` within the already existing `example/metric/README` and be sure to clarify the different paths and details. For example, now we would have multiple `dashboard.json` files (one for `sdk` and one for `exponential_histogram`), so `create_dashboard.sh` needs a parameter and reorganize the files (rename). Also, reorganize the descriptions to have an holistic description (this seems best in the long term now giving more thought to it).

2. Create another folder `exponential_histogram` with a new self contained `README` and steps like i did in the current state of the PR.

I don't have any strong opinion between options 1. or 2. , i just chose 2. initially for simplicity. I could do 1. too. WDYT @damemi ?

dashpole:
I'll let @damemi do this review

damemi:
@franciscovalentecastro ah I see. What if we add a new `dashboards` folder in `example/metric` and put just the `dashboard.json` files in there.

Then copy `create_dashboard.sh` to each example, and modify the command to point to the right dashboard file (ie `../dashboards/export.json` or `../dashboards/exponential_histogram.json`). Then the README is the same for each example.

The only difference is the chart image, but you could add a small README to each example that just shows the final chart image.

My reasoning behind this is that the `sdk` and `collector` examples aren't really grouped as one (they're intentionally separate examples, they just share the same metric names). `otlpgrpc` is also its own example, and this one doesn't seem like it needs to be organized on its own either.

damemi:
@franciscovalentecastro could you please update the package in 
 [`exponential_histogram/go.mod`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ed7ca960a45de2039422f7051fc3e59f8da99029/example/metric/exponential_histogram/go.mod#L1)? it conflicts with the `sdk` path

franciscovalentecastro:
> What if we add a new dashboards folder in example/metric and put just the dashboard.json files in there.

Thanks for the review and guidance. I agree with all comments. I addressed all requested changes.

> could you please update the package in [`exponential_histogram/go.mod`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ed7ca960a45de2039422f7051fc3e59f8da99029/example/metric/exponential_histogram/go.mod#L1)? it conflicts with the `sdk` path

Fixed.

damemi (approved):
Thanks @franciscovalentecastro !

PR #823, opened 2024-03-21
Add additional fallbacks for GMP exporter

### LINKED ISSUE #799: Add addiitonal fallbacks for the GMP exporter
Currently, MapToPrometheusTarget  has the following fallbacks for job and instance:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fda999eac0b4566cd4d32e1792318b7eb41d456d/exporter/collector/googlemanagedprometheus/monitoredresource.go#L42-L49

We should do the following:

1. We should not use `service.name` if it begins with `unknown_service`, since that isn't as useful.
2. We should use k8s.pod.name as an additional fallback for `instance`
3. We should use: k8s.deployment.name, k8s.statefulset.name, k8s.daemonset.name, k8s.job.name, k8s.cronjob.name as additional fallbacks for `job`

damemi:
Fixes #799

### DIFF
--- a/exporter/collector/googlemanagedprometheus/monitoredresource.go
+++ b/exporter/collector/googlemanagedprometheus/monitoredresource.go
@@ -40,12 +40,38 @@ const (
 // promTargetKeys are attribute keys which are used in the prometheus_target monitored resource.
 // It is also used by GMP to exclude these keys from the target_info metric.
 var promTargetKeys = map[string][]string{
-	locationLabel:         {locationLabel, semconv.AttributeCloudAvailabilityZone, semconv.AttributeCloudRegion},
-	clusterLabel:          {clusterLabel, semconv.AttributeK8SClusterName},
-	namespaceLabel:        {namespaceLabel, semconv.AttributeK8SNamespaceName},
-	jobLabel:              {jobLabel, semconv.AttributeServiceName, semconv.AttributeFaaSName},
-	serviceNamespaceLabel: {semconv.AttributeServiceNamespace},
-	instanceLabel:         {instanceLabel, semconv.AttributeServiceInstanceID, semconv.AttributeFaaSInstance},
+	locationLabel: {
+		locationLabel,
+		semconv.AttributeCloudAvailabilityZone,
+		semconv.AttributeCloudRegion,
+	},
+	clusterLabel: {
+		clusterLabel,
+		semconv.AttributeK8SClusterName,
+	},
+	namespaceLabel: {
+		namespaceLabel,
+		semconv.AttributeK8SNamespaceName,
+	},
+	jobLabel: {
+		jobLabel,
+		semconv.AttributeServiceName,
+		semconv.AttributeFaaSName,
+		semconv.AttributeK8SDeploymentName,
+		semconv.AttributeK8SStatefulSetName,
+		semconv.AttributeK8SDaemonSetName,
+		semconv.AttributeK8SJobName,
+		semconv.AttributeK8SCronJobName,
+	},
+	serviceNamespaceLabel: {
+		semconv.AttributeServiceNamespace,
+	},
+	instanceLabel: {
+		instanceLabel,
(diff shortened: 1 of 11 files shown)


exporter/collector/googlemanagedprometheus/monitoredresource.go
```diff
@@ -94,12 +120,6 @@ func getStringOrDefault(attributes pcommon.Map, orElse string, keys ...string) s
 			return val.Str()
 		}
 	}
-	if contains(keys, string(semconv.AttributeServiceName)) {
```
dashpole, line 97: I think this is supposed to stay, right?

damemi: yup my mistake

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json
```diff
@@ -942,7 +942,6 @@
               "http_scheme": "http",
               "k8s_container_name": "rabbitmq",
               "k8s_node_name": "10.92.5.2",
-              "k8s_pod_name": "rabbitmq-server-0",
```
dashpole, line 945: Do you know why this gets removed?

damemi: yup now that `k8s.pod.name` gets parsed to `job` the extra metrics function ignores it: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/2be0414d564ce62759e2da63615b27103ab950d3/exporter/collector/googlemanagedprometheus/extra_metrics.go#L180-L185

dashpole: Thats unfortunate, since we may have only added the service.name to the monitored resource.  It would be nice if we only dropped labels that were _actually_ used in the monitored resource, but i'm not sure how hard that is to do.

damemi: No you're totally right, I assumed this was still being preserved but I see it's not. This shouldn't be dropped, I'll have to fix that logic

exporter/collector/googlemanagedprometheus/extra_metrics_test.go
```diff
@@ -66,6 +66,7 @@ func TestAddExtraMetrics(t *testing.T) {
 				metric := sm.Metrics().AppendEmpty()
 				metric.SetName("target_info")
 				metric.SetEmptyGauge().DataPoints().AppendEmpty().SetIntValue(1)
+				metric.Gauge().DataPoints().At(0).Attributes().PutStr(locationLabel, "us-east")
```
damemi, line 69: @dashpole when we only drop service.name/etc, do we want to drop GMP mr labels too? Like here, `location` will get added to the monitored resource *and* as a metric label... that won't work right?

exporter/collector/integrationtest/testdata/fixtures/metrics/google_managed_prometheus_expect.json
damemi: Updated in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/823/commits/1fbb3d532461dfcf4085e48900a2e44e8a92ead3 based on offline discussion

exporter/collector/googlemanagedprometheus/extra_metrics_test.go
dashpole: Yeah, that sounds right.  Drop it if it is one of the reserved labels ("location", "cluster", "namespace", etc), or if it is a service.name/service.namespace/service.instance.id label

dashpole (approved):

PR #828, opened 2024-04-05
Add support for BMS machines to the resource detection library

alex-basinov:
This PR adds support for BMS machines to the resource detection library.

I'm planning to upgrade the resource detection processor within the opentelemetry-collector-contrib repository to make use of this new BMS detector, aiming to enable native BMS support in the otel collector.

Due to the absence of a metadata server in the BMS environment, we rely on the presence of BMS_PROJECT_ID, BMS_REGION, and BMS_INSTANCE_ID  environment variables as an indication that the collector is running on BMS machines. Customers are expected to manually enter these variables into the Ops Agent's collector using the DefaultEnvironment parameter in the /etc/systemd/system.conf file.

### DIFF
--- a/detectors/gcp/bms.go
+++ b/detectors/gcp/bms.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/detectors/gcp/bms.go
+++ b/detectors/gcp/bms.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


detectors/gcp/bms.go
```diff
@@ -0,0 +1,53 @@
...
+package gcp
+
+const (
+	bmsProjectIDEnv  = "BMS_PROJECT_ID"
+	bmsRegionEnv     = "BMS_REGION"
+	bmsInstanceIDEnv = "BMS_INSTANCE_ID"
+)
+
+// Use BMS_PROJECT_ID, BMS_REGION and BMS_INSTANCE_ID env vars as an indication that we are running on BMS.
+func (d *Detector) onBMS() bool {
```
dashpole, line 24: Readers are unlikely to know what the BMS abbreviation means.  Can you write it out here and elsewhere (e.g. onBareMetalSolution), and add a link to documentation for BMS?

alex-basinov: Done.

PR #831, opened 2024-04-12
fix: flatten ValueTypeSlice to ValueTypeString

crossk3:
:wave: 

While working with the OpenTelemetry stack for Python as well as an OpenTelemetry Collector (Cloud Run Sidecar, per docs [here](https://github.com/GoogleCloudPlatform/opentelemetry-cloud-run)), I noticed a discrepancy in behavior.

GCP Cloud Trace doesn't support Slice/Array data types for Trace Attributes, which Python's `CloudTraceSpanExporter` handles by [flattening Slices to Strings](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py#L536-L540C). However, the golang Collector Exporter currently silently drops any `Slice[T]` attribute. This results in some oddities when choosing how to send traces to GCP Cloud Trace, where any time a collector exports to GCP Cloud Trace you silently lose attributes you would otherwise receive in Cloud Trace.

This PR attempts to unify this behavior, so that the same behavior is observed whether using the in-process exporter or the googlecloud Collector exporter.

### DIFF
--- a/exporter/collector/spandata.go
+++ b/exporter/collector/spandata.go
@@ -16,7 +16,6 @@ package collector
 
 import (
 	"context"
-	"strings"
 	"time"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -150,13 +149,8 @@ func pdataAttributesToOTAttributes(attrs pcommon.Map, resource pcommon.Resource)
 				otAttrs = append(otAttrs, attribute.Int64(k, v.Int()))
 			case pcommon.ValueTypeDouble:
 				otAttrs = append(otAttrs, attribute.Float64(k, v.Double()))
-			case pcommon.ValueTypeSlice:
-				slice := v.Slice()
-				stringArr := make([]string, 0, slice.Len())
-				for i := 0; i < slice.Len(); i++ {
-					stringArr = append(stringArr, slice.At(i).AsString())
-				}
-				otAttrs = append(otAttrs, attribute.String(k, strings.Join(stringArr, ",")))
+			default:
+				otAttrs = append(otAttrs, attribute.String(k, v.AsString()))
 			}
 			return true
 		})
--- a/exporter/collector/spandata.go
+++ b/exporter/collector/spandata.go
@@ -16,6 +16,7 @@ package collector
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -149,6 +150,13 @@ func pdataAttributesToOTAttributes(attrs pcommon.Map, resource pcommon.Resource)
 				otAttrs = append(otAttrs, attribute.Int64(k, v.Int()))
 			case pcommon.ValueTypeDouble:
 				otAttrs = append(otAttrs, attribute.Float64(k, v.Double()))
+			case pcommon.ValueTypeSlice:
+				slice := v.Slice()
+				stringArr := make([]string, 0, slice.Len())
+				for i := 0; i < slice.Len(); i++ {
+					stringArr = append(stringArr, slice.At(i).AsString())
+				}
+				otAttrs = append(otAttrs, attribute.String(k, strings.Join(stringArr, ",")))
 			}
 			return true
 		})
--- a/exporter/collector/spandata_test.go
+++ b/exporter/collector/spandata_test.go
@@ -15,6 +15,7 @@
 package collector
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


exporter/collector/spandata.go
```diff
@@ -149,6 +150,13 @@ func pdataAttributesToOTAttributes(attrs pcommon.Map, resource pcommon.Resource)
 				otAttrs = append(otAttrs, attribute.Int64(k, v.Int()))
 			case pcommon.ValueTypeDouble:
 				otAttrs = append(otAttrs, attribute.Float64(k, v.Double()))
+			case pcommon.ValueTypeSlice:
```
dashpole, line 153: Seems like we should use json-encoded strings for these: https://opentelemetry.io/docs/specs/otel/common/#attribute

> For protocols that do not natively support non-string values, non-string values SHOULD be represented as JSON-encoded strings.

Seems like as a fallback, we could just use `v.AsString()`, as that handles this case as well as Map and Bytes: https://github.com/open-telemetry/opentelemetry-collector/blob/15f10456afcdc57288a56376bc087d3594599104/pdata/pcommon/value.go#L367

dashpole:
Thanks for finding this @crossk3!

exporter/collector/spandata.go
crossk3: Good catch - I'll take a look at that!

crossk3: @dashpole taking a look at your comment in more depth, I originally implemented this with a simple `v.AsString()` since that `json.Marshal`s under the hood anyways as you pointed out. I changed approaches to this one after I realized the behavior would still be different, as the collector exporter would include extra characters `[`  `]` `"`.

I think having the primary case do `json.Marshal(v.At(i).AsRaw())` is a reasonable approach.

dashpole: Are the extra characters problematic?  It seems like it would be good to follow the specification unless we have good reason to deviate

crossk3: tl;dr: Better UX through consistent experience + more intuitive interface, and a different interpretation of spec.

----
@dashpole I had chosen to move to this approach for two main reasons:
1. match the other GCP Python exporter's behavior to provide a consistent experience within Cloud Trace
2. As a user, when I want to apply filters from within Cloud Trace for Slice attributes, manually entering the extra chars from JSON serialization seems unintuitive.

My use case in particular is scraping HTTP headers as attributes. Given a header `X-Custom-Header` with values `value1` and `value2`, I'd expect I could go into Trace Explorer and filter for `http.request.header.x_custom_header: value1,value2`. Having to filter for `http.request.header.x_custom_header: "[\"value1\", \"value2\"]"` feels unusual to me and would probably require further documentation.

I'm also not entirely convinced that part of spec applies in this case, as Cloud Trace does natively support non-string values as seen with the primitives. It just doesn't support the array of primitives, which is not the condition given in the original quoted part of spec.

>The attribute value is either:
>  - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
>  - An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types.

>For protocols that do not natively support non-string values, non-string values SHOULD be represented as JSON-encoded strings.

Ultimately its your package, so I'm happy to implement how you'd like. Just let me know and I'll make whatever changes necessary!

crossk3: also, apologies for the verbosity. This is my first PR to a widely adopted open-source project, so I'm trying to be thorough, not pedantic :sweat_smile:

dashpole: No worries.  I'll raise this with the rest of my team, and get back to you.

crossk3: sounds good, thanks for your help!

dashpole: We decided to use the JSON-encoding for slices

crossk3: @dashpole sorry, I missed this notification. Sounds good, json-encoded it is! I'll update the implementation shortly. Thanks for looking into it.

exporter/collector/spandata_test.go
```diff
@@ -82,9 +83,14 @@ func TestPDataResourceSpansToOTSpanData_endToEnd(t *testing.T) {
 	span.Attributes().PutInt("timeout_ns", 12e9)
 	span.Attributes().PutInt("ping_count", 25)
 	span.Attributes().PutStr("agent", "ocagent")
+	span.Attributes().PutEmptySlice("header")
+	strArr := []any{"value1", "value2"}
+	headerValue, _ := span.Attributes().Get("header")
+	headerValue.Slice().FromRaw(strArr)
```
dashpole, line 89:
```suggestion
	headerValue, found := span.Attributes().Get("header")
	assert.True(t, found)
	err := headerValue.Slice().FromRaw(strArr)
	assert.NoError(t, err)
```
Check return values

exporter/collector/spandata_test.go
```diff
@@ -82,9 +83,14 @@ func TestPDataResourceSpansToOTSpanData_endToEnd(t *testing.T) {
...
 
 	gotOTSpanData := pdataResourceSpansToOTSpanData(rs)
 
+	jsonStr, _ := json.Marshal(strArr)
```
dashpole, line 93:
```suggestion
	jsonStr, err := json.Marshal(strArr)
	assert.NoError(t, err)
```

dashpole:
Ah, you will need to run `make fixtures` to regenerate the test files.

crossk3:
> Ah, you will need to run `make fixtures` to regenerate the test files.

@dashpole done!

dashpole (approved):

PR #835, opened 2024-04-18
Update grpc example to allow long runs

psx95:
Updates the sample to add a `--keepRunning` flag that runs the sample program for 2 hours, generating a test span every 10 seconds. A successful run of this sample would verify that the per rpc credentials refresh is working as expected with the `otlpgrpctraceexporter`. 

### Testing
 - `go run example.go`
 - `go run example.go --keepRunning`

### DIFF
--- a/example/trace/otlpgrpc/example.go
+++ b/example/trace/otlpgrpc/example.go
@@ -17,6 +17,7 @@ package main
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"log"
 	"time"
@@ -30,6 +31,8 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/oauth"
 )
+
+var sustainedRun = flag.Bool("keepRunning", false, "Set to true for generating spans at a fixed rate for 2 hours. Default is false.")
 
 func initTracer() (func(), error) {
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


dashpole (approved):

example/trace/otlpgrpc/example.go
```diff
@@ -62,7 +65,34 @@ func initTracer() (func(), error) {
...
+
+func generateSpansAtFixedRate(ctx context.Context, tr trace.Tracer) {
+	fmt.Println("Generating 1 test span every 10 seconds for 2 hours")
+	totalDuration := 2 * time.Hour
+	endTime := time.Now().Add(totalDuration)
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+
+	for tick := range ticker.C {
+		if tick.After(endTime) {
```
damemi, line 85: any reason to set the limit? I think we could just make this run indefinitely

psx95: No, no technical reason, other than this is not written as a web service and that ADC expires after an hour

Why would we want to keep it indefinite though ?

damemi: I'd expect it to keep running until I kill it. Hardcoding a limit is kind of unexpected and it means you can only test 1 token refresh

psx95: Ok, that makes sense to me. I will update it to keep running.

As a side benefit the flag name would perfectly fit the description 😄

example/trace/otlpgrpc/example.go
```diff
@@ -31,6 +32,8 @@ import (
 	"google.golang.org/grpc/credentials/oauth"
 )
 
+var keepRunning = flag.Bool("keepRunning", false, "Set to true for generating spans at a fixed rate indefinitely. Default is false.")
```
aabmass, line 35: I believe the `--help` text will already print the default so it's not necessary to add it, but please double check if I'm wrong.

damemi (approved):

aabmass (approved):

example/trace/otlpgrpc/example.go
psx95: Looks like `./example --help` does not print the default value.

```
Usage of ./example:
  -keepRunning
    	Set to true for generating spans at a fixed rate indefinitely.
```

aabmass: Seems like it only prints the default if the default is `true`

Issue #834, opened 2024-04-18
GMP exporter should deduplicate resources when writing target_info


dashpole:
When using the hostmetrics receiver, each scraper seems to emit metrics under its own resource, which causes DuplicateTimeseries errors when using the GMP exporter, as each resource is turned into its own target_info metric.  Since metrics all come from the same host, they all collide.

The exporter should handle cases like these by deduplicating resource or target_info metrics.

dashpole:
To work around the Duplicate Timeseries error, use this config:

```yaml
      googlemanagedprometheus:
        metric:
          extra_metrics_config:
            enable_target_info: false
```

PR #838, opened 2024-04-23
Deduplicate target_info and scope_info

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/834

It ensures there is only one target_info for each unique combination of service.* resource attributes.  This ensures that GMP metrics can always be joined with target_info on job + instance.

It ensures there is only one otel_scope_info for each scope name + version within a resource.

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -105,6 +105,12 @@ func addUntypedMetrics(m pmetric.Metrics) {
 	}
 }
 
+// resourceID identifies a resource. We only send one target info for each
+// resourceID.
+type resourceID struct {
+	serviceName, serviceInstanceID, serviceNamespace string
+}
+
 // addTargetInfoMetric inserts target_info for each resource.
 // First, it extracts the target_info metric from each ResourceMetric associated with the input pmetric.Metrics
 // and inserts it into a new ScopeMetric for that resource, as specified in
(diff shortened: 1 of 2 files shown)


punya (approved):

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -113,10 +119,17 @@ func (c Config) addTargetInfoMetric(m pmetric.Metrics) {
 	if !c.ExtraMetricsConfig.EnableTargetInfo {
 		return
 	}
+	ids := make(map[resourceID]bool)
```
damemi, line 122: do you have any preference for `map[resourceId]bool` vs `map[resourceId]struct{}`?

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -201,33 +225,37 @@ func (c Config) addTargetInfoMetric(m pmetric.Metrics) {
...
+}
+
 // addScopeInfoMetric adds the otel_scope_info metric to a Metrics slice as specified in
 // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.16.0/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1
 // It also updates all other metrics with the corresponding scope_name and scope_version attributes, if they are present.
 func (c Config) addScopeInfoMetric(m pmetric.Metrics) {
 	if !c.ExtraMetricsConfig.EnableScopeInfo {
 		return
 	}
+	ids := make(map[scopeID]bool)
```
damemi, line 241: same comment, noting here too

damemi (approved):

exporter/collector/googlemanagedprometheus/extra_metrics.go
dashpole: ah, looks like struct{} is smaller

dashpole: fixed

exporter/collector/googlemanagedprometheus/extra_metrics.go
dashpole: fixed

exporter/collector/googlemanagedprometheus/extra_metrics.go
damemi: yeah it's a tiny difference but I thought it could be helpful here

PR #839, opened 2024-04-23
Use the prometheus.type metric metadata to support double-writing unknown-typed metrics

dashpole:
Depends on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32605

This removes the `gcp.untypedDoubleExport` feature gate, as it is no longer needed.

## Specification

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#gauges-1

> An [OpenTelemetry Gauge](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#gauge) MUST be converted to a Prometheus Unknown-typed metric if the prometheus.type key of [metric.metadata](https://github.com/open-telemetry/opentelemetry-proto/blob/c451441d7b73f702d1647574c730daf7786f188c/opentelemetry/proto/metrics/v1/metrics.proto#L199) is unknown. Otherwise, it MUST be converted to a Prometheus Gauge.

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics.go
@@ -17,26 +17,13 @@ package googlemanagedprometheus
 import (
 	"time"
 
-	"go.opentelemetry.io/collector/featuregate"
+	"github.com/prometheus/common/model"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
 )
 
-const (
-	// Special attribute key used by Ops Agent prometheus receiver to denote untyped
-	// prometheus metric. Internal use only.
-	GCPOpsAgentUntypedMetricKey = "prometheus.googleapis.com/internal/untyped_metric"
-
-	gcpUntypedDoubleExportGateKey = "gcp.untypedDoubleExport"
-)
-
-var untypedDoubleExportFeatureGate = featuregate.GlobalRegistry().MustRegister(
-	gcpUntypedDoubleExportGateKey,
-	featuregate.StageAlpha,
-	featuregate.WithRegisterFromVersion("v0.77.0"),
-	featuregate.WithRegisterDescription("Enable automatically exporting untyped Prometheus metrics as both gauge and cumulative to GCP."),
-	featuregate.WithRegisterReferenceURL("https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/668"))
+const prometheusMetricMetadataTypeKey = "prometheus.type"
 
 func (c Config) ExtraMetrics(m pmetric.Metrics) {
 	addUntypedMetrics(m)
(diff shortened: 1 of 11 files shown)


dashpole:
cc @ridwanmsharif

dashpole:
Blocked by https://github.com/open-telemetry/opentelemetry-collector/pull/10026

dashpole:
I pinned the pdata dependency only in the integration test module (so it doesn't affect the released modules) to include https://github.com/open-telemetry/opentelemetry-collector/pull/10026.  This makes sure the fixture-based tests continue to work correctly.

dashpole:
This is ready for review now

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -25,6 +25,7 @@ import (
 
 // GetMetricName returns the metric name with GMP-specific suffixes. The.
 func (c Config) GetMetricName(baseName string, metric pmetric.Metric) (string, error) {
+	unknown := isUnknown(metric)
```
damemi, line 28: you could pass `isUnknown()` directly to the functions instead of making a new variable

exporter/collector/googlemanagedprometheus/naming.go
```diff
@@ -60,35 +61,24 @@ func (c Config) GetMetricName(baseName string, metric pmetric.Metric) (string, e
...
 // the original metric name before calling prometheus.BuildCompliantName(), which is hacky.
-// It is based on the untyped_prometheus_metric data point attribute, and behind a feature gate.
-func getUnknownMetricName(points pmetric.NumberDataPointSlice, suffix, secondarySuffix, originalName, compliantName string) string {
-	if !untypedDoubleExportFeatureGate.IsEnabled() {
-		return compliantName + suffix
-	}
-
+//
+//nolint:revive
+func getUnknownMetricName(points pmetric.NumberDataPointSlice, unknown bool, suffix, secondarySuffix, originalName, compliantName string) string {
```
damemi, line 66: it would be nice if there was a clean way to just pass `metric` as a param (either instead of or along the `NumberDataPointSlice`), so then you could just call `isUnknown` inside the function. I'm assuming the bool param is what failed revive (the check is called something like conditional control..?)

exporter/collector/googlemanagedprometheus/extra_metrics.go
```diff
@@ -17,26 +17,13 @@ package googlemanagedprometheus
...
+	"github.com/prometheus/common/model"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
 )
 
-const (
-	// Special attribute key used by Ops Agent prometheus receiver to denote untyped
-	// prometheus metric. Internal use only.
-	GCPOpsAgentUntypedMetricKey = "prometheus.googleapis.com/internal/untyped_metric"
```
damemi, line 29: will this break consistency/compatibility with the ops agent? Are they supporting this semconv yet?

damemi (approved):

exporter/collector/googlemanagedprometheus/extra_metrics.go
dashpole: As long as they update their prometheus receiver when they update this exporter, they won't break.  Thats partially why I want Ridwan to review.

exporter/collector/googlemanagedprometheus/naming.go
dashpole: refactored to avoid this

exporter/collector/googlemanagedprometheus/naming.go
dashpole: passed in metric directly

exporter/collector/googlemanagedprometheus/extra_metrics.go
ridwanmsharif: +1 to what David said, this shouldn't break the Ops Agent, we plan on unforking all prometheus components at once

ridwanmsharif (approved):
Woooo! So happy we can unfork now

Issue #843, opened 2024-04-30
GMP: target_info and otel_scope_info metrics are INT-typed


dashpole:
Prometheus metrics scraped from a prometheus endpoint are always double-typed, so if a user is using GMP's managed collection, and has a target_info or otel_scope_info metric, that metric will have a double value type.

This means that using the collector's GMP exporter and managed GMP collection will result in one of the two having write errors.

This is similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/798, in that we want to write all GMP metrics as double to avoid this scenario.

Issue #798, opened 2024-01-24
GMP exporter should convert Int values to Double


dashpole:
~~GMP only accepts Double values.  Sending it an int will cause the metric to be rejected.~~
This is not correct.  GMP accepts int and double values.

This can be done with the transform processor today, although I think this needs to be filtered so it only applies to int valued datapoints:

```
      transform/inttodouble:
        metric_statements:
          - context: datapoint
            statements:
              # GMP metrics are expected to be double
              - set(value_double, Double(value_int))
```

Ideally, the exporter should handle this, rather than requiring users to add processors.  I encountered this when using the spanmetrics exporter, which produces int-valued metrics.

dashpole:
This is tricky.  If we do this in the exporter, existing users will encounter errors when they upgrade, since their existing metric descriptors are the wrong type.  We would likely need to make this opt-in behavior.

aabmass:
If GMP only supports doubles, how could their descriptors have the wrong type on upgrade?

dashpole:
I think I must have been mistaken.  There was a user that was able to write int metrics, and had problems switching to double.

dashpole:
Confirmed that GMP accepts both int and double metrics

dashpole:
In that case, we should probably not change the value type.  If users have the same metric in both INT and Double, they can use the above workaround to merge them together.

olekgo:
Commenting here since it seems very relevant but happy to open a new issue. 

> If users have the same metric in both INT and Double, they can use the above workaround to merge them together.

Is it possible to change the value type of a metric in GMP after it has been created?

I've tried deleting the metric descriptor via the API ([doc](https://cloud.google.com/monitoring/docs/samples/monitoring-delete-metric)) and then pushing the metric with a new type. This approach worked partially. I can now write the metric (previously, I was getting a `Value type for metric prometheus.googleapis.com/container_memory_rss/gauge conflicts with the existing value type (INT64).` error). However, I now encounter the following error when trying to query the metric:

```
An error occurred requesting data. bad_data: invalid parameter "query": substitute queries failed: convert vector selector failed: mismatching value types between inputs to the union: [DOUBLE INT64]

```

This has broken the metric in the whole Metric Scope.

dashpole:
@olekgo i'm no expert in metrics scopes, but I suspect it is still an INT in a different project in the metrics scope?  If you are able, I would open a bug/ticket with the standard GCP support channel.

[...]

PR #844, opened 2024-04-30
Add feature gate to convert int metrics to double

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/798

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/843

A user that turns this on will need to manually delete metric descriptors that are failing to write with mismatched INT vs DOUBLE errors.

### DIFF
--- a/exporter/collector/googlemanagedprometheus/extra_metrics_test.go
+++ b/exporter/collector/googlemanagedprometheus/extra_metrics_test.go
@@ -19,6 +19,8 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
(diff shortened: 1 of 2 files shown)


exporter/collector/googlemanagedprometheus/extra_metrics_test.go
```diff
@@ -239,6 +242,52 @@ func TestAddExtraMetrics(t *testing.T) {
...
+			expected: func() pmetric.ResourceMetricsSlice {
+				metrics := testMetric(timestamp).ResourceMetrics()
+				scopeMetrics := metrics.At(0).ScopeMetrics()
+
+				// Insert a new, empty ScopeMetricsSlice for this resource that will hold target_info
+				sm := scopeMetrics.AppendEmpty()
+				metric := sm.Metrics().AppendEmpty()
+				metric.SetName("target_info")
+				// This changes the value to double because of the feature gate.
+				metric.SetEmptyGauge().DataPoints().AppendEmpty().SetDoubleValue(1)
```
damemi, line 262: should probably add a test case for Sum too, and one for the default/continue case

damemi (approved):

PR #850, opened 2024-05-08
feat: WithMonitoredResourceAttributes and WithMonitoredResourceDescription

bhshkh:
Context: https://github.com/googleapis/google-cloud-go/pull/10046

I'm working on exporting Bigtable client metrics to Google Cloud Monitoring.
I have tried using the exporter in its current form:

CBT client side metrics
In step 15 in the above diagram, the monitored resource in CreateServiceTimeSeries request has 2 fields:
1. **Type**: OpenTelemetry resource with attributes gets created in step 5. The exporter uses the “cloud.platform” resource attribute value to determine the monitored resource type. Complete logic can be seen [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fda999eac0b4566cd4d32e1792318b7eb41d456d/internal/resourcemapping/resourcemapping.go#L188-L200). 
**Problem 1**: The CBT client side metrics need to be written to ‘bigtable_client_raw’ monitored resource type but the exporter does not allow writing to resource types other than the ones mentioned [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fda999eac0b4566cd4d32e1792318b7eb41d456d/internal/resourcemapping/resourcemapping.go#L66) 
2. **Labels**: OpenTelemetry resource with attributes gets created in step 5. The exporter adds these resource attributes as labels to the monitored resource. The ‘bigtable_client_raw’ monitored resource needs to written with labels: project, instance, table, cluster and zone
**Problem 2**: Only project and instance values are known at step 5 point in time. Table is not known until data operation such as ReadRows is invoked. 
A Bigtable instance is a container for user's data. Instances have one or more clusters, located in different zones. When library sends requests to a Bigtable instance, those requests are handled by one of the clusters in the instance. The cluster is not known while sending the request. Cluster and zone need to be read from response headers of data operation. Thus, the existing metric exporter cannot add these labels to the monitored resource.

To resolve problem 1, solution similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/346 is being included in this PR.

How to include  monitored resource labels?
1. While recording the metrics, add the required labels to the metrics
2. Exporter will use the function provided in the `WithMonitoredResourceAttributes` to extract the resource labels from the metric labels and add them to monitored resource. Example usage:

https://github.com/bhshkh/google-cloud-go/blob/bc1d79ae1a71a774b96ab94b875b8a29c0046501/bigtable/metric_monitoring_exporter.go#L72-L75


exporter/metric/option.go
```diff
@@ -43,10 +44,20 @@ type options struct {
...
+	monitoredResourceCreator func(*metricdata.ResourceMetrics) (*monitoredrespb.MonitoredResource, error)
+
 	// resourceAttributeFilter determinies which resource attributes to
 	// add to metrics as metric labels. By default, it adds service.name,
 	// service.namespace, and service.instance.id.
 	resourceAttributeFilter attribute.Filter
+
+	// metricAttributesFilter determinies which metric attributes to
+	// add to metrics as metric labels
+	metricAttributesFilter attribute.Filter
```
dashpole, line 59: You shouldn't need this config in the exporter.  Use https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#WithView to pass views to the metrics SDK, which allow filtering attributes on metrics.

dashpole:
cc @psx95

psx95:
> cc @psx95

This looks similar to the recent request we had for the Java exporter, I will take a look into this PR.

psx95:
A similar change to the Java exporter was just approved and merged in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/346. We would likely want to follow similar logic here.

exporter/metric/option.go
bhshkh: with views, the metric attributes do not reach the exporter at all. The attributes need to reach the exporter. Then, the exporter can add them to resource

bhshkh:
> Similar functionality has been added in #854.

The functionality added in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/854 solves only one of the issues this PR was trying to address

Hey

When using github.com/GoogleCloudPlatform/opentelemetry-operations-go, user needs to define a resource with attributes and then these attributes get added to monitored resource label while exporting to GCM

For e.g.

```

// create the exporter with the Monitored Resource Description
exporter, err := mexporter.New(mexporter.WithProjectID(os.Getenv("GOOGLE_PROJECT_ID")), mexporter.WithMonitoredResourceDescription("dataflow_job", []string{"job_name", "project_id", "region"}))
    if err != nil {
        log.Fatalf("Failed to create exporter: %v", err)
    }

...

// Make sure the OTel resource has the `gcp.resoure_type` along with the required labels configured.
 res, err := resource.New(
        ctx,
        // Use the GCP resource detector to detect information about the GCP platform
        resource.WithDetectors(gcp.NewDetector()),
        // Keep the default detectors
        resource.WithTelemetrySDK(),
        // Add attributes from environment variables
        resource.WithFromEnv(),
        // Add your own custom attributes to identify your application
        resource.WithAttributes(
            semconv.ServiceNameKey.String("example-application"),
            attribute.KeyValue{Key: "gcp.resource_type", Value: attribute.StringValue("dataflow_job")},
            attribute.KeyValue{Key: "job_name", Value: attribute.StringValue("some-job")},
            attribute.KeyValue{Key: "region", Value: attribute.StringValue("us1")},
        ),
    )

// Construct the exporter using the above config
provider := sdkmetric.NewMeterProvider(
        sdkmetric.WithReader(sdkmetric.NewPeriodicReader(exporter)),
        sdkmetric.WithResource(res),
    )
```

Here, the attributes "job_name", "region", etc would get added as monitored resource labels in CreateServiceTimeSeries requests.

This setup works well for users when all the resource attributes are known before setting up the meter provider or while creating the resource. 

But in Bigtable, a monitored resource needs labels cluster, instance, project_id, label and zone. Out of these attributes, cluster and zone are not known when the meter provider is created i.e. it is unknown while client creation. 
The values for these labels are returned by Bigtable service in each RPC response header/trailer. Thus, these can be different per RPC.

PR #852, opened 2024-05-22
Add host.id as a fallback for instance

dashpole:
This will prevent users from getting a blank `instance` label for GMP when running on GCE.

FYI @quentinmit @ridwanmsharif @jefferbrecht

### DIFF
--- a/exporter/collector/googlemanagedprometheus/monitoredresource_test.go
+++ b/exporter/collector/googlemanagedprometheus/monitoredresource_test.go
@@ -336,6 +336,26 @@ func TestMapToPrometheusTarget(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "Attributes from GCE with instance ID override",
+			resourceLabels: map[string]string{
+				"cloud.platform":      "gcp_compute_engine",
+				"cloud.region":        "us-central1",
+				"service.name":        "service-name",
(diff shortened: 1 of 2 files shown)


exporter/collector/googlemanagedprometheus/monitoredresource_test.go
```diff
@@ -320,10 +320,10 @@ func TestMapToPrometheusTarget(t *testing.T) {
 		{
 			desc: "Attributes from GCE with environment label",
 			resourceLabels: map[string]string{
-				"cloud.platform":      "gcp_compute_engine",
-				"cloud.region":        "us-central1",
-				"service.name":        "service-name",
-				"service.instance.id": "1234759430923053489543203",
```
damemi, line 326: should we add a test for if `service.instance.id` and `host.id` are both set? Or is that not possible?

damemi (approved):

exporter/collector/googlemanagedprometheus/monitoredresource_test.go
dashpole: It is possible.  I can test for that

dashpole: added

Issue #841, opened 2024-04-26
Export Exponential Histogram to Managed Prometheus doesn't work due to unimplemented naming convertion


alexvanboxel:
The Google Managed Prometheus naming doesn't support the `pmetric.MetricTypeExponentialHistogram`; this makes the export of Exponential Histograms fail.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/21602689459a185ee389b4ea871b392de5a010ef/exporter/collector/googlemanagedprometheus/naming.go#L54

I think adding the following will fix the issue. But I don't know what the correct Suffix would be.

```
	case pmetric.MetricTypeExponentialHistogram:
		return compliantName + "/???", nil
```

dashpole:
Thanks @alexvanboxel.  We are aware of this, and are coordinating with the Google Managed Prometheus team on adding support for exponential histograms.

alexvanboxel:
I got it working locally; I've added the above code with `/distribution` and it works... at least for so far as I see (I can see something in Metrics Explorer. Need more exploration.

dashpole:
It should work just fine with the regular metrics viewer.  I don't believe it will work with promQL yet.

dashpole:
the GMP team has said that they plan to use the `/histogram` suffix we already use for regular histograms

PR #853, opened 2024-05-30
Add support for exponential histograms in the GMP exporter

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/841

This does not currently include support for PromQL on exponential histograms written to GMP, which should come later.  Exponential histograms sent with the GMP exporter will need to be visualized using the query builder or MQL.

### DIFF
--- a/exporter/collector/integrationtest/testdata/fixtures/metrics/exponential_histogram_gmp_expect.json
+++ b/exporter/collector/integrationtest/testdata/fixtures/metrics/exponential_histogram_gmp_expect.json
@@ -5,15 +5,220 @@
       "timeSeries": [
         {
           "metric": {
-            "type": "prometheus.googleapis.com/target_info/gauge"
+            "type": "prometheus.googleapis.com/durationhist_seconds/histogram"
           },
           "resource": {
             "type": "prometheus_target",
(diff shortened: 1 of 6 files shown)


exporter/collector/integrationtest/testdata/fixtures/metrics/exponential_histogram_gmp_expect.json
```diff
@@ -410,5 +615,6 @@
         }
       }
     ]
-  }
+  },
+  "userAgent": "opentelemetry-collector-contrib latest grpc-go/1.63.2"
```
damemi, line 619: just curious do you know why this is new?

damemi (approved):

exporter/collector/integrationtest/testdata/fixtures/metrics/exponential_histogram_gmp_expect.json
dashpole: the file: exponential_histogram_gmp_expect.json wasn't actually used in any tests until now, so it likely had an old version of the fixture from before we included the userAgent in the fixture.

PR #863, opened 2024-06-20
fix(googleclientauth/idtoken): use metadata server when credential json is not available

rinx:
`googleclientauth` extension with `id_token` mode works correctly when credential.json is specified.
However, the current implementation doesn't work correctly with the metadata server.

This PR fixes its behavior.
I confirmed that it works correctly on Cloud Run.

### DIFF
--- a/extension/googleclientauthextension/factory.go
+++ b/extension/googleclientauthextension/factory.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/idtoken"
+	"google.golang.org/api/option"
 	"google.golang.org/grpc/credentials/oauth"
 )
 
@@ -67,10 +68,16 @@ func (ca *clientAuthenticator) Start(ctx context.Context, _ component.Host) erro
 func (ca *clientAuthenticator) newTokenSource(ctx context.Context, creds *google.Credentials) (oauth2.TokenSource, error) {
 	switch ca.config.TokenType {
 	case idToken:
+		opts := []option.ClientOption{}
+
+		if len(creds.JSON) > 0 {
+			opts = append(opts, idtoken.WithCredentialsJSON(creds.JSON))
+		}
+
 		return idtoken.NewTokenSource(
 			ctx,
 			ca.config.Audience,
-			idtoken.WithCredentialsJSON(creds.JSON),
+			opts...,
 		)
 	default:
 		return creds.TokenSource, nil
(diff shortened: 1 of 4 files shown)


damemi (changes requested):
Reading the code for [NewTokenSource](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/idtoken/idtoken.go#L83), it looks like [`idtoken.WithCredentialsJSON`](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/idtoken/idtoken.go#L242) returns an option function that [just copies the JSON (`creds.json`) to a field](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/option/option.go#L61-L70).

[NewTokenSource](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/idtoken/idtoken.go#L83) then calls [`internal.Creds()`](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/idtoken/idtoken.go#L123-L137) which calls [baseCreds()](https://github.com/googleapis/google-api-go-client/blob/20ffdd800072046285b0900af690d05e2a6f9524/internal/creds.go#L31-L43) which [reads from the JSON if it's not `nil`](https://github.com/googleapis/google-api-go-client/blob/20ffdd800072046285b0900af690d05e2a6f9524/internal/creds.go#L130-L132). Otherwise, [newTokenSource() falls back to the metadata server by default](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/idtoken/idtoken.go#L131-L135).

I'm guessing that the bug here comes from [the current line always setting `idtoken.WithCredentialsJSON(creds.JSON)`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/8ffe2564d48ba8085950552371de0915cfdd9b6f/extension/googleclientauthextension/factory.go#L73) even if `creds.JSON` is empty. 

My confusion though, is that I would expect the option function I linked above to [make an empty array](https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/option/option.go#L67-L70), but I guess that's different from a `nil` array? And that's the bug, because `baseCreds()` is [checking for `creds != nil`](https://github.com/googleapis/google-api-go-client/blob/20ffdd800072046285b0900af690d05e2a6f9524/internal/creds.go#L130-L132) when maybe it should be checking for `len(creds) > 0`.

Either way, @rinx is there any way we can add a test for this? Verifying manually is good but we should have something reproducible

rinx:
@damemi Your point is correct.
I've added several test cases to verify the changes are correct.

damemi (approved):

damemi (approved):

Issue #797, opened 2024-01-23
Migrate collector exporter self-observability from OpenCensus to OpenTelemetry


dashpole:
Using OpenTelemetry for self-observability metrics is now stable in the collector: https://github.com/open-telemetry/opentelemetry-collector/pull/9102.  OpenCensus metrics are still supported on the prometheus endpoint: https://github.com/open-telemetry/opentelemetry-collector/blob/7ade1016cf138965685e533f60ec83892d26abc6/service/internal/proctelemetry/config.go#L212, but the OpenCensus bridge is not used if you are exporting with OTLP, which we might want to support in the future.

Our OpenCensus usage:

We define `googlecloudmonitoring/point_count` and `googlecloudmonitoring/exemplar_attachments_dropped` in [exporter/collector/observability.go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fda999eac0b4566cd4d32e1792318b7eb41d456d/exporter/collector/observability.go) using OpenCensus.  Migrating this to OpenTelemetry should be straightforward.

We also use [ocgrpc](go.opencensus.io/plugin/ocgrpc) in the collector exporter: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fda999eac0b4566cd4d32e1792318b7eb41d456d/exporter/collector/config.go#L263.  Migrating this to OTel may be tricky, as we would need to either migrate to otelgrpc, or to gRPC's upcoming self-observability metrics: https://github.com/grpc/proposal/blob/master/A66-otel-stats.md.  This is likely to be a breaking change, which we will need to migrate with a featuregate.

dashpole:
cc @braydonk @quentinmit

dashpole:
The OC bridge has now been removed: https://github.com/open-telemetry/opentelemetry-collector/pull/10406

This means we have now lost any opencensus-based metrics, and recent versions will no longer have these.

We don't need a feature gate, since the old metrics have already been removed (and there is no way for us to bring them back).

dashpole:
Self-observability metrics have been missing since v0.104.0, and can be re-enabled by setting the service.disableOpenCensusBridge feature gate to false.  Starting in v0.106.0, it was updated to stable, and OC metrics can no longer be re-enabled

PR #885, opened 2024-08-26
Migrate collector self-observability from OpenCensus to OpenTelemetry

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/797

## Changes

This migrates two uses of OpenCensus:

* go.opencensus.io/plugin/ocgrpc -> google.golang.org/grpc/stats/opentelemetry
* Custom instrumentation:
    * googlecloudmonitoring/point_count
    * googlecloudmonitoring/exemplar_attachments_dropped
    
This includes:

* Update gRPC v1.64.1 -> v1.65.0
* [breaking] Accepting a MeterProvider in our New* functions, and plumbing this MeterProvider to our instrumentation
* Change fixture generation and validation to use our Go in-process exporter to export self-observability metrics to GCM.
* Update fixture normalization to normalize new grpc metric names, and to normalize a label value to ensure they don't change during each run of fixture tests.

This PR additionally includes extending self-observability fixtures to the log and trace exporters for additional coverage.

### Notes to reviewers

It is probably easiest to review each commit.  Most of the diff is from generated fixture changes.

### Alternatives considered:

We could have used otelgrpc instead (go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc).  This PR uses the grpc instrumentation because we expect the native instrumentation to eventually replace the opentelemetry-provided library.


exporter/collector/config.go
```diff
@@ -266,7 +266,7 @@ func generateClientOptions(ctx context.Context, clientCfg *ClientConfig, cfg *Co
 			// option.WithGRPCConn option takes precedent over all other supplied options so the
 			// following user agent will be used by both exporters if we reach this branch
 			dialOpts := []grpc.DialOption{
-				grpc.WithStatsHandler(&ocgrpc.ClientHandler{}),
+				// TODO: use meterProvider for gRPC instrumentation
```
dashpole, line 269: @aabmass @damemi The only thing left for this PR is to choose to use either the gRPC or go-contrib library here.  I tried to use the gRPC version, but it hasn't actually been released yet: https://pkg.go.dev/google.golang.org/grpc/stats/opentelemetry (notice the tag is just a commit).

@zasweq do you know if the grpc OpenTelemetry metrics plugin is ready to be used by other projects, or when you expect the first release to be tagged?

exporter/collector/metrics.go
```diff
@@ -115,12 +116,15 @@ type metricMapper struct {
 	normalizer normalization.Normalizer
 	obs        selfObservability
 	cfg        Config
+	// self-observability metrics
+	exemplarAttachmentDropCount metricapi.Int64Counter
```
damemi, line 118: would this fit better under `obs` than its own field in the metricMapper?

exporter/collector/metrics.go
```diff
@@ -93,6 +92,8 @@ type MetricsExporter struct {
 	// goroutines tracks the currently running child tasks
 	goroutines sync.WaitGroup
 	timeout    time.Duration
+	// self-observability metrics
+	pointsExportedCounter metricapi.Int64Counter
```
damemi: would this also go better under `obs`?

exporter/collector/metrics.go
```diff
@@ -455,13 +472,64 @@ func (me *MetricsExporter) export(ctx context.Context, req *monitoringpb.CreateT
...
+		me.pointsExportedCounter.Add(
+			ctx,
+			int64(failedPoints),
+			metricapi.WithAttributes(attribute.String("status", st)),
+		)
 	}
 	return err
 }
 
+func statusCodeToString(s *status.Status) string {
```
damemi, line 490: could you use [`status.Code().String()`](https://github.com/grpc/grpc-go/blob/2da976983bbb33feb3e25b7daaa8f60b9769adb5/codes/code_string.go#L31) instead of redeclaring the function here?

damemi (approved):
looks good, couple nits

exporter/collector/metrics.go
dashpole: obs is shared by the metrics, logs, and trace exporters, so it didn't seem like the right place to put it.

exporter/collector/metrics.go
dashpole: same as above.

exporter/collector/metrics.go
dashpole: That would change the codes from e.g. `PERMISSION_DENIED` to `PermissionDenied`.  The capitalized version appears to be what the gRPC instrumentation will give us, so I suspect it was done to make the labels the same between this metric and the grpc metric.  I would prefer to leave it as-is for this PR so we don't make any breaking changes here.

dashpole: feel free to open a follow-up issue if you think it is worth doing.

PR #886, opened 2024-08-27
Switch to uint64 hash for identifier

dashpole:
```
goos: linux
goarch: amd64
pkg: github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/internal/datapointstorage
cpu: AMD EPYC 7B12
             │ bench2.txt  │             bench3.txt             │
             │   sec/op    │   sec/op     vs base               │
Identifier-2   4.805µ ± 4%   2.539µ ± 0%  -47.16% (p=0.002 n=6)

             │ bench2.txt  │            bench3.txt             │
             │    B/op     │    B/op     vs base               │
Identifier-2   2208.0 ± 0%   632.0 ± 0%  -71.38% (p=0.002 n=6)

             │ bench2.txt │            bench3.txt            │
             │ allocs/op  │ allocs/op   vs base              │
Identifier-2   47.00 ± 0%   44.00 ± 0%  -6.38% (p=0.002 n=6)
```
I'm still working on testing edge cases to make sure identifiers don't collide.

### DIFF
--- a/exporter/collector/internal/datapointstorage/datapointcache.go
+++ b/exporter/collector/internal/datapointstorage/datapointcache.go
@@ -211,9 +211,11 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
 	return true
 }
 
+// Use the same constants as Prometheus uses for hashing:
+// https://github.com/prometheus/prometheus/blob/282fb1632ad62a82401a230f486538a72384faf0/model/labels/labels_common.go#L32
 var (
 	itemSep = []byte{'\xfe'} // Used between identifiers
-	KVsep   = []byte{'\xff'} // Used between map keys and values
+	kvSep   = []byte{'\xff'} // Used between map keys and values
 )
 
 // Identifier returns the unique string identifier for a metric.
@@ -244,8 +246,8 @@ func hashOfMap(h hash.Hash64, m map[string]string) {
 	sort.Strings(keys)
 	for _, key := range keys {
 		h.Write([]byte(key))
-		h.Write(KVsep)
+		h.Write(kvSep)
 		h.Write([]byte(m[key]))
-		h.Write(KVsep)
+		h.Write(kvSep)
 	}
 }
--- a/exporter/collector/internal/datapointstorage/datapointcache.go
+++ b/exporter/collector/internal/datapointstorage/datapointcache.go
@@ -212,57 +212,29 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
 }
 
 // Identifier returns the unique string identifier for a metric.
-func Identifier(resource *monitoredrespb.MonitoredResource, extraLabels map[string]string, metric pmetric.Metric, attributes pcommon.Map) (uint64, error) {
-	var err error
+func Identifier(resource *monitoredrespb.MonitoredResource, extraLabels map[string]string, metric pmetric.Metric, attributes pcommon.Map) uint64 {
 	h := fnv.New64()
-
-	_, err = h.Write([]byte(resource.GetType()))
-	if err != nil {
-		return 0, err
-	}
-	_, err = h.Write([]byte(metric.Name()))
-	if err != nil {
-		return 0, err
-	}
-
+	h.Write([]byte(resource.GetType()))
+	h.Write([]byte(metric.Name()))
 	attrs := make(map[string]string)
 	attributes.Range(func(k string, v pcommon.Value) bool {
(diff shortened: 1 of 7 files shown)


bwplotka:
FYI: Checkout amazing https://pkg.go.dev/golang.org/x/perf/cmd/benchstat tool for easier comparision of those results 🤗

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -212,27 +212,57 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
...
-	// Instrumentation library labels and additional resource labels
-	fmt.Fprintf(&b, " - %v", extraLabels)
-
-	// Metric identifiers
-	fmt.Fprintf(&b, " - %s -", metric.Name())
-	attrsIds := make([]string, 0, attributes.Len())
+	attrs := make(map[string]string)
 	attributes.Range(func(k string, v pcommon.Value) bool {
-		attrsIds = append(attrsIds, k+"="+v.AsString())
+		attrs[k] = v.AsString()
```
bwplotka, line 230: Why writing to hash directly here?

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -212,27 +212,57 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
...
+}
+
+func hashOfMap(h hash.Hash64, m map[string]string) error {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		_, err := h.Write([]byte(key))
```
bwplotka, line 258: I would ignore all those errors explicitly `_ = ` for hash you use it's noop I believe (https://github.com/prometheus/prometheus/blob/main/model/labels/labels_dedupelabels.go#L191)

bwplotka (approved):
This is a solid start 👍🏽  Good work!

exporter/collector/internal/datapointstorage/datapointcache.go
dashpole: yeah, they aren't guaranteed to be sorted, so we are sorting first.

exporter/collector/internal/datapointstorage/datapointcache.go
dashpole: updated to ignore errors

dashpole:
added delimiters to make sure we don't get collisions between the different maps

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -211,28 +211,41 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
...
-	// Resource identifiers
-	if resource != nil {
-		fmt.Fprintf(&b, "%v", resource.GetLabels())
-	}
-
-	// Instrumentation library labels and additional resource labels
-	fmt.Fprintf(&b, " - %v", extraLabels)
+var (
+	itemSep = []byte{'\xfe'} // Used between identifiers
+	KVsep   = []byte{'\xff'} // Used between map keys and values
```
aabmass, line 216: Does this need to be public?

aabmass (approved):

exporter/collector/internal/datapointstorage/datapointcache.go
```diff
@@ -211,28 +211,41 @@ func (c *Cache) gc(shutdown <-chan struct{}, tickerCh <-chan time.Time) bool {
...
-	var b strings.Builder
-
-	// Resource identifiers
-	if resource != nil {
-		fmt.Fprintf(&b, "%v", resource.GetLabels())
-	}
-
-	// Instrumentation library labels and additional resource labels
-	fmt.Fprintf(&b, " - %v", extraLabels)
+var (
```
aabmass, line 214: It might be nice to leave a comment on how you chose these separators. Is there a guarantee that the individual strings wouldn't contain these bytes?

exporter/collector/internal/datapointstorage/datapointcache.go
dashpole: nope.  Made non-public

exporter/collector/internal/datapointstorage/datapointcache.go
dashpole: added a comment.  They are the delimiters used by prometheus/prometheus for similar hashing.  I added a link above the constants.

PR #889, opened 2024-09-09
exporter/collector: avoid using req.String() to check for empty request

ridwanmsharif:
This change is motivated by some CPU profiling that showed that its responsible for over 20% of CPU time in my environment (running avalanche, OTel with Prom receiver and the gmp exporter). Sending this change because its logically equivalent to before but avoids the serialization to the string.

Flamegraph: pprof/?id=f2daddf0bdf127bfa136618996b263c2&tab=flame&path=1ygnz740h91y6o


exporter/collector/metrics.go
```diff
@@ -434,7 +434,7 @@ func (me *MetricsExporter) exportToTimeSeries(ctx context.Context, req *monitori
 func (me *MetricsExporter) export(ctx context.Context, req *monitoringpb.CreateTimeSeriesRequest) error {
 	// if this is an empty request, skip it
 	// empty requests are used by the WAL to signal the end of pending data
-	if len(req.String()) == 0 {
+	if req == nil || req.TimeSeries == nil || len(req.TimeSeries) == 0 {
```
quentinmit, line 437: The `req.TimeSeries == nil` check is redundant; `nil` slices have a `len` of 0.

ridwanmsharif: simplified the check

exporter/collector/metrics.go
```diff
@@ -434,7 +434,7 @@ func (me *MetricsExporter) exportToTimeSeries(ctx context.Context, req *monitori
 func (me *MetricsExporter) export(ctx context.Context, req *monitoringpb.CreateTimeSeriesRequest) error {
 	// if this is an empty request, skip it
 	// empty requests are used by the WAL to signal the end of pending data
-	if len(req.String()) == 0 {
+	if req == nil || (req.TimeSeries == nil && req.Name == "") {
```
bwplotka, line 437: Idea for readability: define a helper function `isReqEmpty` -- compiler will likely inline it anyway.

exporter/collector/metrics.go
```diff
@@ -504,7 +504,7 @@ func (me *MetricsExporter) readWALAndExport(ctx context.Context) error {
 		// If we are at the last index, and this last index is not an empty request
 		// (we use empty requests to fill out the end of a log, and if we didn't check for them
 		// this would loop constantly adding empty requests onto the end)
-		if readIndex == writeIndex && len(req.String()) > 0 {
+		if readIndex == writeIndex && (req != nil && (req.Name != "" || req.TimeSeries != nil)) {
```
bwplotka, line 507: Same here, I guess here it would be `!isReqEmpty(req)` then?

bwplotka (approved):
LGTM, amazing work! Sounds like our benchmarks already has been proven to be useful.

The best optimizations are one-liners 💪🏽 

I guess it would be nice to run some microbenchmark here to double check, but in theory this be faster vs stringifying

ridwanmsharif:
Added some microbenchmarks, here are some savings:

```
➜  collector git:(ridwanmsharif/check-empty) ✗ go test -bench . -benchmem -count=10 > orig.txt
➜  collector git:(ridwanmsharif/check-empty) ✗ go test -bench . -benchmem -count=10 > fix.txt 
➜  collector git:(ridwanmsharif/check-empty) ✗ benchstat orig.txt fix.txt            
goos: linux
goarch: amd64
pkg: github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
                         │  orig.txt   │               fix.txt               │
                         │   sec/op    │   sec/op     vs base                │
_TestExport-16             3.194µ ± 7%   1.910µ ± 7%  -40.21% (p=0.000 n=10)
_TestReadWALAndExport-16   7.585m ± 3%   6.797m ± 3%  -10.39% (p=0.000 n=10)
geomean                    155.6µ        113.9µ       -26.80%

                         │   orig.txt   │               fix.txt                │
                         │     B/op     │     B/op      vs base                │
_TestExport-16               953.0 ± 0%     656.0 ± 0%  -31.16% (p=0.000 n=10)
_TestReadWALAndExport-16   8.060Ki ± 1%   7.466Ki ± 0%   -7.36% (p=0.000 n=10)
geomean                    2.739Ki        2.187Ki       -20.14%

                         │  orig.txt  │              fix.txt               │
                         │ allocs/op  │ allocs/op   vs base                │
_TestExport-16             20.00 ± 0%   10.00 ± 0%  -50.00% (p=0.000 n=10)
_TestReadWALAndExport-16   128.0 ± 0%   108.0 ± 0%  -15.62% (p=0.000 n=10)
geomean                    50.60        32.86       -35.05%
```

When WAL is disabled (default) the savings are higher, but with WAL this still has significant savings.

Issue #888, opened 2024-09-09
Investigate performance improvements for datapoint normalization


dashpole:
From a customer's profiler of their production workload, NormalizeXXX functions in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/internal/normalization/standard_normalizer.go account for a good chunk of the memory usage of the collector exporter.

There are benchmarks for these functions in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/internal/normalization/benchmark_test.go

Most of the memory usage comes from our use of CopyTo for data points.  We do that because our exporter is marked as non-mutating.  Maybe there is an alternative way of storing what we need so we can keep less data around.

PR #890, opened 2024-09-09
Optimize normalization caching and copying

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/888

### Changes

This avoids creating new cached elements, updates existing cached datapoints if they are already there, and only copies the required fields into the cache.  In particular, not storing attributes lowers memory usage significantly. The benchmark difference is:

```
goos: linux
goarch: amd64
pkg: github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/internal/normalization
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
                                              │   main.txt   │            mincache.txt            │
                                              │    sec/op    │   sec/op     vs base               │
NormalizeNumberDataPoint-2                       606.5n ± 7%   585.2n ± 3%        ~ (p=0.132 n=6)
NormalizeHistogramDataPoint-2                    727.2n ± 3%   684.4n ± 7%   -5.89% (p=0.041 n=6)
NormalizeExopnentialHistogramDataPoint-2         776.1n ± 4%   801.5n ± 3%   +3.28% (p=0.009 n=6)
NormalizeSummaryDataPoint-2                      552.4n ± 5%   331.3n ± 2%  -40.02% (p=0.002 n=6)
ResetNormalizeNumberDataPoint-2                  904.4n ± 2%   591.8n ± 3%  -34.57% (p=0.002 n=6)
ResetNormalizeHistogramDataPoint-2              1198.5n ± 4%   868.4n ± 6%  -27.54% (p=0.002 n=6)
ResetNormalizeExponentialHistogramDataPoint-2    949.9n ± 4%   651.4n ± 3%  -31.43% (p=0.002 n=6)
ResetNormalizeSummaryDataPoint-2                 859.9n ± 4%   383.6n ± 3%  -55.39% (p=0.002 n=6)
geomean                                          799.8n        584.7n       -26.89%

                                              │  main.txt   │              mincache.txt              │
                                              │    B/op     │    B/op     vs base                    │
NormalizeNumberDataPoint-2                       16.00 ± 0%   16.00 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeHistogramDataPoint-2                    32.00 ± 0%   32.00 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeExopnentialHistogramDataPoint-2         48.00 ± 0%   48.00 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeSummaryDataPoint-2                      16.00 ± 0%    0.00 ± 0%  -100.00% (p=0.002 n=6)
ResetNormalizeNumberDataPoint-2                 128.00 ± 0%   16.00 ± 0%   -87.50% (p=0.002 n=6)
ResetNormalizeHistogramDataPoint-2              256.00 ± 0%   64.00 ± 0%   -75.00% (p=0.002 n=6)
ResetNormalizeExponentialHistogramDataPoint-2   256.00 ± 0%   16.00 ± 0%   -93.75% (p=0.002 n=6)
ResetNormalizeSummaryDataPoint-2                 128.0 ± 0%     0.0 ± 0%  -100.00% (p=0.002 n=6)
geomean                                          67.33                    ?                      ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                                              │  main.txt  │              mincache.txt              │
                                              │ allocs/op  │ allocs/op   vs base                    │
NormalizeNumberDataPoint-2                      2.000 ± 0%   2.000 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeHistogramDataPoint-2                   3.000 ± 0%   3.000 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeExopnentialHistogramDataPoint-2        4.000 ± 0%   4.000 ± 0%         ~ (p=1.000 n=6) ¹
NormalizeSummaryDataPoint-2                     1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.002 n=6)
ResetNormalizeNumberDataPoint-2                 4.000 ± 0%   1.000 ± 0%   -75.00% (p=0.002 n=6)
ResetNormalizeHistogramDataPoint-2              7.000 ± 0%   4.000 ± 0%   -42.86% (p=0.002 n=6)
ResetNormalizeExponentialHistogramDataPoint-2   4.000 ± 0%   1.000 ± 0%   -75.00% (p=0.002 n=6)
ResetNormalizeSummaryDataPoint-2                4.000 ± 0%   0.000 ± 0%  -100.00% (p=0.002 n=6)
geomean                                         3.191                    ?                      ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean
```

Note that this primarily improves the "reset" benchmarks.  This is because we only store things in the cache when a reset happens.  This is primarily capturing the fact that we are re-using data points, rather than making new ones.

### DIFF
--- a/exporter/collector/internal/datapointstorage/datapointcache.go
+++ b/exporter/collector/internal/datapointstorage/datapointcache.go
@@ -93,7 +93,14 @@ func (c *Cache) GetNumberDataPoint(identifier uint64) (pmetric.NumberDataPoint, 
 func (c *Cache) SetNumberDataPoint(identifier uint64, point pmetric.NumberDataPoint) {
 	c.numberLock.Lock()
 	defer c.numberLock.Unlock()
-	c.numberCache[identifier] = usedNumberPoint{point, atomic.NewBool(true)}
+	if existing, ok := c.numberCache[identifier]; ok {
+		existing.used.Store(true)
+		minimalNumberDataPointCopyTo(point, existing.point)
+		return
+	}
+	newPoint := pmetric.NewNumberDataPoint()
+	minimalNumberDataPointCopyTo(point, newPoint)
+	c.numberCache[identifier] = usedNumberPoint{newPoint, atomic.NewBool(true)}
 }
 
 // GetSummaryDataPoint retrieves the point associated with the identifier, and whether
@@ -112,7 +119,14 @@ func (c *Cache) GetSummaryDataPoint(identifier uint64) (pmetric.SummaryDataPoint
 func (c *Cache) SetSummaryDataPoint(identifier uint64, point pmetric.SummaryDataPoint) {
 	c.summaryLock.Lock()
 	defer c.summaryLock.Unlock()
-	c.summaryCache[identifier] = usedSummaryPoint{point, atomic.NewBool(true)}
+	if existing, ok := c.summaryCache[identifier]; ok {
+		existing.used.Store(true)
+		minimalSummaryDataPointCopyTo(point, existing.point)
+		return
+	}
+	newPoint := pmetric.NewSummaryDataPoint()
+	minimalSummaryDataPointCopyTo(point, newPoint)
+	c.summaryCache[identifier] = usedSummaryPoint{newPoint, atomic.NewBool(true)}
 }
 
 // GetHistogramDataPoint retrieves the point associated with the identifier, and whether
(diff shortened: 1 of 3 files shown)


dashpole:
Ready for review

aabmass:
> Since our exporter is marked non-mutating, we can't change the data point that is passed to us

Would the code be significantly simpler if we could mutate? And are there any negative consequences of marking the exporter as mutating?

dashpole:
> Would the code be significantly simpler if we could mutate? And are there any negative consequences of marking the exporter as mutating?

That might be a good idea to explore as well.  It looks like it only matters if there are multiple exporters in the config: https://github.com/open-telemetry/opentelemetry-collector/blob/2166b11137ec473b01b6cc5a0b9bf932cb752cf2/internal/fanoutconsumer/metrics.go#L21.  If there is only a single one, there is no additional clone involved.

We already are doing a full clone for GMP anyways: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/a6eab1eadf29840b84f651dfca699765aa086290/exporter/collector/metrics.go#L342.  Given that, it seems like it would be better to just mark our exporter as mutating, and simplify those things.

> Could we wrap the pdata struct to hide this complexity to the caller?

We probably could, but that is a lot of boilerplate.

dashpole:
Lets try https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/892 first, and then see how much more of a benefit we get from some of the optimizations here.

damemi (approved):

PR #892, opened 2024-09-12
Make the collector exporter mutate data, and remove unnecessary CopyTo

dashpole:
Looking at https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/890#issuecomment-2344718704

This change makes the collector exporter now mutate data.  We were doing multiple layers additional copying in our exporter to prevent mutating data, but it is likely better for us to just be a mutating exporter, and let the framework handle it.

This keeps the collector code clean, and should have better CPU/Memory usage (even after the additional copy) since we are copying fewer times.

Benchmarks (including those added in https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/893):

```
goos: linux
goarch: amd64
pkg: github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector/internal/normalization
cpu: AMD EPYC 7B12
                                              │   main.txt    │             mutating.txt              │
                                              │    sec/op     │    sec/op     vs base                 │
NormalizeNumberDataPoint-2                      1845.0n ± ∞ ¹   532.6n ± ∞ ¹        ~ (p=1.000 n=1) ²
NormalizeHistogramDataPoint-2                   2314.0n ± ∞ ¹   615.7n ± ∞ ¹        ~ (p=1.000 n=1) ²
NormalizeExopnentialHistogramDataPoint-2        2071.0n ± ∞ ¹   673.7n ± ∞ ¹        ~ (p=1.000 n=1) ²
NormalizeSummaryDataPoint-2                     1718.0n ± ∞ ¹   465.9n ± ∞ ¹        ~ (p=1.000 n=1) ²
ResetNormalizeNumberDataPoint-2                 2094.0n ± ∞ ¹   826.3n ± ∞ ¹        ~ (p=1.000 n=1) ²
ResetNormalizeHistogramDataPoint-2               2.565µ ± ∞ ¹   1.043µ ± ∞ ¹        ~ (p=1.000 n=1) ²
ResetNormalizeExponentialHistogramDataPoint-2   2148.0n ± ∞ ¹   869.4n ± ∞ ¹        ~ (p=1.000 n=1) ²
ResetNormalizeSummaryDataPoint-2                1775.0n ± ∞ ¹   836.2n ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                          2.049µ         709.9n        -65.36%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

[...]

PR #893, opened 2024-09-12
Add benchmark for normalization with reset points

dashpole:
Split from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/890

Required for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/892

### DIFF
--- a/exporter/collector/internal/normalization/benchmark_test.go
+++ b/exporter/collector/internal/normalization/benchmark_test.go
@@ -28,21 +28,18 @@ func BenchmarkNormalizeNumberDataPoint(b *testing.B) {
 	shutdown := make(chan struct{})
 	defer close(shutdown)
 	normalizer := NewStandardNormalizer(shutdown, zap.NewNop())
-	startPoint := pmetric.NewNumberDataPoint()
-	startPoint.SetTimestamp(pcommon.NewTimestampFromTime(time.Now()))
-	startPoint.SetIntValue(12)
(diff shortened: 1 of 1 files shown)


exporter/collector/internal/normalization/benchmark_test.go
```diff
@@ -110,27 +91,171 @@ func BenchmarkNormalizeSummaryDataPoint(b *testing.B) {
...
+		assert.True(b, ok)
+	}
+}
+
+var (
+	now   = pcommon.NewTimestampFromTime(time.Now())
+	start = pcommon.NewTimestampFromTime(time.Now().Add(-time.Minute))
+)
+
+func testNumberDataPoint() pmetric.NumberDataPoint {
```
aabmass, line 212: You mentioned that these functions are pretty slow relative to normalization which was throwing off the benchmarks. Was it a problem or are we OK with the shorter benchmark time workaround?

aabmass (approved):

exporter/collector/internal/normalization/benchmark_test.go
dashpole: it seemed to slightly increase the variability in CPU benchmarks.  Memory benchmarks were completely unchanged.  0.01 seconds is what the otel-go uses upstream, so it shouldn't be an issue.

PR #895, opened 2024-09-19
Prepare 1.24.2 and 0.48.2

damemi:


dashpole:
I found time to finish the last optimization: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/890

If it is OK, it would be nice to release after that is merged.

dashpole (approved):

PR #897, opened 2024-09-24
Detect the managed instance group that created a VM

quentinmit:
In preparation for new resource attributes for the MIG.

### DIFF
--- a/detectors/gcp/gce.go
+++ b/detectors/gcp/gce.go
@@ -16,12 +16,18 @@ package gcp
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
+
+	"cloud.google.com/go/compute/metadata"
 )
 
 // See the available GCE instance metadata:
-// https://cloud.google.com/compute/docs/metadata/default-metadata-values#vm_instance_metadata
+// https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys#instance-metadata
 const machineTypeMetadataAttr = "instance/machine-type"
+
+// https://cloud.google.com/compute/docs/instance-groups/getting-info-about-migs#checking_if_a_vm_instance_is_part_of_a_mig
+const createdByMetadataAttr = "instance/created-by"
 
 func (d *Detector) onGCE() bool {
 	_, err := d.metadata.Get(machineTypeMetadataAttr)
@@ -72,4 +78,37 @@ func (d *Detector) GCEAvailabilityZoneAndRegion() (string, string, error) {
 		return "", "", fmt.Errorf("zone was not in the expected format: country-region-zone.  Got %v", zone)
 	}
 	return zone, strings.Join(splitZone[0:2], "-"), nil
+}
+
+type ManagedInstanceGroup struct {
+	Name     string
+	Location string
+	Type     LocationType
+}
+
+var createdByMIGRE = regexp.MustCompile(`^projects/[^/]+/(zones|regions)/([^/]+)/instanceGroupManagers/([^/]+)$`)
+
+func (d *Detector) GCEManagedInstanceGroup() (ManagedInstanceGroup, error) {
+	createdBy, err := d.metadata.Get(createdByMetadataAttr)
+	if _, ok := err.(metadata.NotDefinedError); ok {
+		return ManagedInstanceGroup{}, nil
+	} else if err != nil {
+		return ManagedInstanceGroup{}, err
+	}
+	matches := createdByMIGRE.FindStringSubmatch(createdBy)
+	if matches == nil {
+		return ManagedInstanceGroup{}, nil
+	}
+
+	mig := ManagedInstanceGroup{
+		Name:     matches[3],
+		Location: matches[2],
+	}
+	switch matches[1] {
+	case "zones":
+		mig.Type = Zone
+	case "regions":
+		mig.Type = Region
+	}
+	return mig, nil
 }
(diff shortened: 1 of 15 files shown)


detectors/gcp/gce.go
```diff
@@ -73,3 +79,36 @@ func (d *Detector) GCEAvailabilityZoneAndRegion() (string, string, error) {
...
+func (d *Detector) GCEManagedInstanceGroup() (ManagedInstanceGroup, error) {
+	createdBy, err := d.metadata.Get(createdByMetadataAttr)
+	if _, ok := err.(metadata.NotDefinedError); ok {
+		return ManagedInstanceGroup{}, nil
+	} else if err != nil {
+		return ManagedInstanceGroup{}, err
+	}
+	matches := createdByMIGRE.FindStringSubmatch(createdBy)
+	if matches == nil {
+		return ManagedInstanceGroup{}, nil
```
dashpole, line 100: Should we return an error here? Seems like if it is in the wrong format we should let the user know, right?

quentinmit: The key is just `created-by`; my assumption is that other tools will put their own strings indicating that they created the VM, and that's not an error, it just means the VM wasn't created by a MIG.

dashpole: please add a comment :)

detectors/gcp/utils_test.go
```diff
@@ -14,62 +14,87 @@
...
+	fakeProjectNumber       = "1234567890"
+	fakeInstanceID          = "1234567891"
+	fakeInstanceName        = "my-instance"
+	fakeInstanceHostname    = fakeInstanceName + "." + fakeZone + ".c." + fakeProjectID + ".internal"
+	fakeInstanceMachineType = "n1-standard1"
+)
+
+// fmp constructs a FakeMetadatProvider that responds with the basic GCE metadata fields.
+// To add more metadata, pass key-value pairs as arguments.
+func fmp(keyValues ...string) *FakeMetadataProvider {
```
dashpole, line 82: nit: i'd prefer newFakeMetadataProvider to the abbreviation for readability.

detectors/gcp/utils_test.go
```diff
@@ -14,62 +14,87 @@
...
 }
 func (f *FakeMetadataProvider) Zone() (string, error) {
-	if f.Err != nil {
-		return "", f.Err
+	value, err := f.Get("instance/zone")
+	if err != nil {
+		return value, err
 	}
-	return f.FakeZone, nil
+	// https://github.com/googleapis/google-cloud-go/blob/35d1f813f755c88d6869d993796c5ddf8ac3eea3/compute/metadata/metadata.go#L663
```
dashpole, line 62: This link seems irrelevant?

detectors/gcp/utils_test.go
```diff
@@ -14,62 +14,87 @@
...
+	return value[strings.LastIndex(value, "/")+1:], nil
```
dashpole, line 63: I'm not sure this is helpful for testing our library, even if it is closer to how the metadata library works.

detectors/gcp/gce.go
```diff
@@ -73,3 +79,38 @@ func (d *Detector) GCEAvailabilityZoneAndRegion() (string, string, error) {
 	}
 	return zone, strings.Join(splitZone[0:2], "-"), nil
 }
+
+type ManagedInstanceGroup struct {
+	Name     string
+	Location string
```
dashpole, line 85: Do you think we should omit the location + location type until we are sure we actually need the location attributes in the conventions?

quentinmit: We definitely need a location attribute; there's otherwise no way to distinguish a zonal MIG named `foo` and a regional MIG named `foo`. I guess it's an open question how exactly that information is encoded into attributes, but we do need it in some form.

detectors/gcp/utils_test.go
quentinmit: Yeah, I was torn on whether to do this or not, but I came down on the side of wanting `Get("instance/zone")` and `Zone()` to always return matching values, which means copying this logic.

Now that I've done this simplification we could also get rid of `FakeMetadatProvider` entirely if we want and just use the upstream `metadata.Client` with a mock HTTP handler. Then we'd use upstream's split logic here. WDYT?

dashpole: sounds good.  lets do it

detectors/gcp/gce.go
dashpole: My question is: is the location of the instance (not the MIG) enough, or do we need both?  The location of the instance is already provided via GCEAvailabilityZoneAndRegion

detectors/gcp/utils_test.go
quentinmit: Done

detectors/gcp/gce.go
quentinmit: The location of the instance is always a zone. The location of the MIG can be a zone or region, so no, the location of the instance is not enough by itself. I *think* it might be the case that the location of the MIG is always *either* the zone or the region of the instance, so it could be modeled as *either* `mig.type: regional` (using only what I've called `Type` here) or `mig.region: us-central1` (using both `Location` and `Type`). I think OTel generally doesn't have `type` attributes like that, so I lean towards the latter.

But I can't find any documentation that actually asserts that MIGs will always be in the same region or zone as the instance, so I'm hedging my bets on the API here and including both.

dashpole: ok.  worst case, we have extra API surface we don't use

dashpole (approved):

dashpole:
Code changes LGTM.  Please document the steps you took to manually test this PR in the description before merging.

PR #900, opened 2024-10-09
Fix bug where grpc self-observability metrics were not added

dashpole:
In https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/885, I only added grpc observability metrics when insecure = true.  This passed our tests that use a local, fake grpc server, but didn't work for the general case.

It turns out that the metrics were _actually_ coming from the client library, which enables opencensus metrics by default.  After the OC bridge was removed, those metrics disappeared.

This PR adds grpc metrics for the non-insecure case.

TESTED=checked that the e2e test project contains grpc self-obs metrics

### DIFF
--- a/exporter/collector/integrationtest/integration_selfobs_test.go
+++ b/exporter/collector/integrationtest/integration_selfobs_test.go
@@ -1,0 +1,64 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


dashpole:
@ridwanmsharif

aabmass (approved):
Any way to test this?

dashpole:
The integration test now exports self-observability metrics using our SDK exporter.  I checked in the e2e test project to see that the metrics were present.

exporter/collector/integrationtest/integration_selfobs_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	res, err := resource.New(
+		ctx,
+		resource.WithTelemetrySDK(),
+		resource.WithFromEnv(),
+		resource.WithAttributes(
+			semconv.ServiceNameKey.String("integrationtest"),
+			semconv.ServiceInstanceIDKey.String(version4Uuid.String()),
+		),
+	)
```
ridwanmsharif, line 53: need to add an err check here?:

```
if err != nil {
	log.Fatalf("Failed to create new resource: %v", err)
}
```

dashpole: thanks.  fixed

Issue #871, opened 2024-07-17
Recommended span batch write size


jameshartig:
The trace [Readme.md](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/trace/README.md) doesn't call `WithMaxExportBatchSize` which means the export batch size is the default, 512. This means that every 512 spans the batch span processor will synchronously [1] export the spans during which new spans will be queued up to the `MaxQueueSize` which defaults to 2048. If you are regularly doing 1000s of spans per second you will end up calling `batchWriteSpans` several times a second needlessly using up the 4,800/minute quota [2] and if any of those RPC calls takes more than a couple hundred milliseconds you will end up dropping spans.

The quotas page [2] says that the "Maximum number of spans per `PatchTraces` call" is 25,000 (assuming that `PatchTraces` is the same as `batchWriteSpans`). Does that mean we can set the batch export size to 25,000? I know that many of Google's RPC methods have internal request size limits and I'm concerned that setting it at 25,000 might trip the limit. Is there a recommended value?

There was a discussion [3] a few years ago where it seemed like it was settled on 6k for that particular case mostly because higher batch sizes were slower. Is that still the case now? I also saw that the batching implementation was replaced since then which means there isn't a need for increasing the grpc pool size since there isn't more than 1 outstanding RPC at a time when using the batch span processor, correct?

We have replaced the batch span processor internally with our own implementation to help improve the situation but I still believe the recommendations here are relevant for either one.

[1] https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/batch_span_processor.go#L317
[2] https://cloud.google.com/trace/docs/quotas
[3] #456

jameshartig:
We've gone with a limit of 20,000 for now. I'll report back if we have issues.

aabmass:
Thanks for raising the issue. We'll try to figure out good defaults and update the documentation.

damemi:
`PatchTraces` is the v1 cloud trace API while `BatchWriteSpans` is v2 which our exporter uses. There doesn't actually seem to be any limit on BatchWriteSpans. I'll work on a summary of @dashpole's performance tuning findings from #456 for the readme

Issue #456, opened 2022-07-18
Collector exporter span throughput is lower than OTLP


dashpole:
This is an issue reported by a customer.  May be related to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/388.

Exporter configuration:

```
exporters:
  googlecloud:
    project: <project-id>
    timeout: 10s
  otelhttp:
    endpoint: <endpoint>
```

Using self-observability metrics, the otelhttp exporter shows ~600k spans/s, but the googlecloud exporter seems to stay at 2.6k spans/s.  The number of received spans matches the otelhttp exporter at ~600k spans/s.

dashpole:
We did some benchmarking with the logging exporter, which is probably relevant here.  The configuration that worked the best for them was using a batch processor, and setting the number of consumers and grpc pool size (new in the latest version of the collector):

```yaml
  batch:
    send_batch_max_size: 6000
    send_batch_size: 6000
  googlecloud:
    grpc_pool_size: 20
    sending_queue:
      num_consumers: 40
```

Even with those optimizations, it isn't quite as high-throughput as the OTLP exporter.  We suspect this is because of the latency of cloud monitoring endpoints.  They were able to achieve ~180k logs/s throughput.  I would expect a similar throughput for spans.

dashpole:
With the standard settings:

```yaml
  batch:
    send_batch_max_size: 200
    send_batch_size: 200
  googlecloud:
```
I was able to get ~13k spans/second in a collector.

With this config, I was able to get ~130k spans/second:
```yaml
  batch:
    send_batch_max_size: 6000
    send_batch_size: 6000
  googlecloud:
    sending_queue:
      num_consumers: 40
```

With `grpc_pool_size: 20`, i'd expect to be able to do even better, but I think I need to wait for a collector release for that.

Hopefully that helps.

emmaCullen:
Hi @dashpole , thank you for the tips here, we tried it this morning and saw a huge improvement.

We are still seeing some dropped spans however, and was wondering if you had any suggestions?

**Dropped spans**:
<img width="1006" alt="image (11)" src="https://user-images.githubusercontent.com/61304599/182741683-d6057566-1fc0-4b9e-b0d3-eb74cb432009.png">


**Config used**:
```
  batch:
    send_batch_max_size: 6000
    send_batch_size: 6000
  googlecloud:
    sending_queue:
      num_consumers: 40
```

[...]

PR #903, opened 2024-10-21
Add performance tuning notes to exporter readme

damemi:
Adds notes from #456 to the readme for performance tuning options.

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/871

### DIFF
--- a/exporter/collector/README.md
+++ b/exporter/collector/README.md
@@ -297,6 +297,36 @@ underscores (eg, `service.name`), but these will still collide once all
 attributes are normalized to metric labels. In this case you will need to update
 the collector config above appropriately.
 
+### Performance tuning
+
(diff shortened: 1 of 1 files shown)


exporter/collector/README.md
```diff
@@ -297,6 +297,36 @@ underscores (eg, `service.name`), but these will still collide once all
 attributes are normalized to metric labels. In this case you will need to update
 the collector config above appropriately.
 
+### Performance tuning
```
dashpole, line 300: Are these numbers specific to trace?

dashpole:
I've been thinking we should remove the readme from this repo, and redirect to the upstream README to have a single source of truth.  WDYT?

dashpole (approved):

PR #908, opened 2024-10-31
Remove dependency on prometheus translation package

dashpole:
We can move this logic into collector-contrib until the translation package is stabilized.

This unblocks https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35904

See exporter/collector/integrationtest/testcases/testcases_metrics.go for the change we would need to make in contrib after this.

cc @ArthurSens


aabmass:
> We can move this logic into collector-contrib until the translation package is stabilized.

Is there a separate PR for this?

aabmass (approved):

dashpole:
Yeah, I will have to cut a release and then make the change in contrib

ArthurSens:
thank you ❤️

PR #910, opened 2024-11-01
Enable renovate

dashpole:
This is based on the renovate config used for otel-go and for collector-contrib.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,40 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+      "config:recommended"
+    ],
+    "ignorePaths": [],
+    "labels": ["Skip Changelog", "dependencies"],
+    "postUpdateOptions" : [
+      "gomodTidy"
+    ],
+    "postUpdateTask" : {
+        "commands": ["make", "fixtures"]
+    },
+    "packageRules": [
+      {
+        "matchManagers": ["gomod"],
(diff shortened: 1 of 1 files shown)


dashpole:
It seems hard to test this locally. I'm inclined to just merge it, and see what it generates.

renovate.json
```diff
@@ -0,0 +1,40 @@
...
+      "config:recommended"
+    ],
+    "ignorePaths": [],
+    "labels": ["Skip Changelog", "dependencies"],
+    "postUpdateOptions" : [
+      "gomodTidy"
+    ],
+    "postUpdateTask" : {
+        "commands": ["make", "fixtures"]
+    },
```
aabmass, line 13: FYI this won't work unless we use self-hosted renovate https://docs.renovatebot.com/configuration-options/#postupgradetasks:~:text=Post%2Dupgrade%20tasks%20can%20only%20be%20used%20on%20self%2Dhosted%20Renovate%20instances.

aabmass (approved):
Are we moving from dependabot to renovate then?

renovate.json
dashpole: Got it.  Lets skip for now.

dashpole:
I think we should align with whatever the upstream version of our repository does.  We can lean on work done upstream to keep this config working (e.g. when new versions break things).  Since OTel-go uses renovate, it seems good to do the same here.

renovate.json
dashpole: removed

PR #945, opened 2024-11-26
Update all deps

dashpole:
This should help speed up the process of getting renovate up-to-date, and includes updates that require fixture updates.

This also removes tooling for pinning opentelemetry dependencies.


dashpole:
The toolchain directive seems to cause GAE tests to fail (probably because it doesn't have the right version of the Go toolchain).  I've removed it following https://github.com/open-telemetry/opentelemetry-go/issues/5660

e2e-test-server/cloud_functions/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/e2e-test-server/cloudfunctions
 
-go 1.22
-
-toolchain go1.22.0
+go 1.22.7
```
dashpole, line 3: This seems unintended, and was because of a bump in gRPC that has been reverted.  After the next gRPC release, we should be able to do this update without bumping the go version here.

dashpole:
We should not submit this until we update to the next grpc release

dashpole:
The next grpc release is here, but it looks like the minimum version update has already propagated across the ecosystem.  I think we should move forward with increasing the minimum version to 1.22.7

aabmass (approved):

PR #965, opened 2025-01-10
Ungroup the otel collector and otel SDK dependency updates

dashpole:
These currently are unlikely to succeed because of the size.  Split them to make this more reasonable.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -34,8 +34,13 @@
       },
       {
         "matchManagers": ["gomod"],
-        "matchPackageNames": ["go.opentelemetry.io/**"],
-        "groupName": "go.opentelemetry.io"
+        "matchPackageNames": ["go.opentelemetry.io/collector/**", "github.com/open-telemetry/opentelemetry-collector-contrib/**"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -34,8 +34,13 @@
...
+        "matchPackageNames": ["go.opentelemetry.io/collector/**", "github.com/open-telemetry/opentelemetry-collector-contrib/**"],
+        "groupName": "go.opentelemetry.io/collector"
+      },
+      {
+        "matchManagers": ["gomod"],
+        "matchPackageNames": ["go.opentelemetry.io/otel/**", "go.opentelemetry.io/contrib/**"],
+        "groupName": "go.opentelemetry.io/otel"
       }
     ]
   }
```
psx95, line 46: Do we need to add EOF newline?

psx95 (approved):

renovate.json
dashpole: doesn't seem to matter.  i'll skip if thats ok

PR #970, opened 2025-01-10
Restore trailing slashes to renovate config

dashpole:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/965 seems to have broken grouped updates

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -29,17 +29,17 @@
         "groupName": "golang.org/x"
       },
       {
-        "matchPackageNames": ["cloud.google.com/go**"],
-        "groupName": "cloud.google.com/go"
+        "matchPackageNames": ["cloud.google.com/**"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -29,17 +29,17 @@
         "groupName": "golang.org/x"
       },
       {
-        "matchPackageNames": ["cloud.google.com/go**"],
-        "groupName": "cloud.google.com/go"
+        "matchPackageNames": ["cloud.google.com/**"],
```
psx95, line 32: Was removing `go` intentional ?

psx95 (approved):

renovate.json
dashpole: yes.  It will group any update under cloud.google.com.  The previous match didn't work for cloud.google.com/go/longrunning, for example

PR #989, opened 2025-01-13
Manually update blocked dependency updates

dashpole:
Supersedes https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/979, https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/978, https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/987


detectors/gcp/go.mod
```diff
@@ -1,17 +1,19 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp
 
-go 1.21
+go 1.22
```
psx95, line 3: Is upgrading the minimum go version intentional ?

dashpole: It is ok.  I just haven't gotten around to doing it yet.  Let me do that separately

psx95 (approved):
LGTM. 

Upgrades min go version to 1.22.
Upgrades other dependency version bumps.

PR #992, opened 2025-01-13
Remove remaining testing and usage of go 1.21

dashpole:
Almost everything was using go 1.22 already as the minimum version.  This drops 1.21 tests and updates go mod compatibility versions.


psx95 (approved):
LGTM. 

Support for go 1.21 dropped.

PR #1005, opened 2025-01-22
Add readme for otlp metric sample

dashpole:
`export OTEL_EXPORTER_OTLP_HEADERS=X-Goog-User-Project=$PROJECT_ID` seemed to be required for me, even though I didn't need to do that for python or JS.  Not sure why there is a difference

### DIFF
--- a/example/metric/otlpgrpc/README.md
+++ b/example/metric/otlpgrpc/README.md
@@ -1,0 +1,24 @@
+# OTLP Metric with Google Auth Example
+
+Run this sample to connect to an endpoint that is protected by GCP authentication.
+
+#### Prerequisites
+
+Get Google credentials on your machine:
+
+```sh
+gcloud auth application-default login
+```
+
+#### Run the Sample
+
+```sh
+# export necessary OTEL environment variables
+export PROJECT_ID=<project-id>
+export OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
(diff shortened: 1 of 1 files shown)


psx95:
> `export OTEL_EXPORTER_OTLP_HEADERS=X-Goog-User-Project=$PROJECT_ID` seemed to be required for me, even though I didn't need to do that for python or JS. Not sure why there is a difference

One possible cause might be that ADC implementation in Go is not able to figure out the quota project ID.

psx95 (approved):

example/metric/otlpgrpc/README.md
```diff
@@ -0,0 +1,24 @@
...
+gcloud auth application-default login
+```
+
+#### Run the Sample
+
+```sh
+# export necessary OTEL environment variables
+export PROJECT_ID=<project-id>
+export OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
+export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=$PROJECT_ID,service.name=otlp-sample,service.instance.id=1"
```
psx95, line 19: Could be enhanced with resource detectors ?

dashpole: Not sure what you mean.  Other than service.instance.id, you can't detect these attributes

PR #1007, opened 2025-01-27
Prepare for v1.26.0/v0.50.0

dashpole:


psx95 (approved):
LGTM. 

The PR contains only version bumps across all updated files.

PR #1018, opened 2025-02-20
collector: use buildinfo to make user agent

braydonk:
This PR adjusts the `NewGoogleCloud*Exporter` functions to take a `component.BuildInfo`, and to format that build info into the UserAgent instead of defaulting to `opentelemetry-collector-contrib`.


exporter/collector/config_test.go
```diff
@@ -15,12 +15,19 @@
...
 import (
+	"fmt"
 	"testing"
 
+	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/confmap"
 	"gopkg.in/yaml.v3"
 )
 
+var testBuildInfo = component.BuildInfo{
```
dashpole, line 26: nit: consider moving this into the test function since it is only used there.  It makes it easier to read inputs and outputs when they are closer in the file.

exporter/collector/logs.go
```diff
@@ -163,10 +163,10 @@ func NewGoogleCloudLogsExporter(
 	cfg Config,
 	log *zap.Logger,
 	meterProvider metric.MeterProvider,
-	version string,
+	buildInfo component.BuildInfo,
```
dashpole, line 166: if we are going to make a breaking API change, should we switch to accepting exporter.Settings?  That way we can remove the logger argument, the meterprovider argument, and the version string argument.  Same comment for other new() functions

dashpole: Also note that we will need to coordinate the release of this repo with a dependency update in opentelemetry-collector-contrib

dashpole (commented):
I don't see any changes in the googlemanagedprometheus exporter code.  I thought that was where the bug was?

exporter/collector/logs.go
braydonk: I have the coordinated change queued up locally and ready to go once there is a new tag to refer to

braydonk:
> I don't see any changes in the googlemanagedprometheus exporter code. I thought that was where the bug was?

No changes needed by coincidence. The bug in there is that the `DefaultConfig` function did not provide a user agent string which all other code was expecting to exist. However I've changed that expectation in this PR to be that the string being empty means the user did not provide a user agent and to provide the fallback at that time. If the `googlemanagedprometheus.DefaultConfig` function was providing what it should have been originally, the change would have been to remove the `UserAgent` field set in that function, but that is already how it was.

exporter/collector/config_test.go
braydonk: It was originally used in the other files (kind of hard to tell) but since I changed things to `exporter.Settings` that's not the case anymore so I'll fix this

example/metric/collector/go.mod
```diff
@@ -1,30 +1,30 @@
...
 
-go 1.22.0
+go 1.22.7
 
-toolchain go1.23.4
+toolchain go1.24.0
 
 require (
-	go.opentelemetry.io/otel v1.31.0
+	go.opentelemetry.io/otel v1.30.0
```
dashpole, line 8: looks like some dependencies have been downgraded?  Can you revert that?

braydonk: Thanks I didn't notice that. Seems like an OTel upgrade happened without the `update-otel` make target at some point so the versions in the Makefile were out of sync.

example/trace/otlphttp/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/trace/otlphttp
 
-go 1.22.0
+go 1.22.7
```
dashpole, line 3: looks like this updates the minimum go version in a bunch of modules.  I don't think we want to do that yet

example/metric/sdk/go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/example/metric/sdk
 
-go 1.22
+go 1.22.7
 
-toolchain go1.22.0
+toolchain go1.24.0
```
dashpole, line 5: I don't think we test on 1.24.0 yet.  That is probably OK?

example/trace/otlpgrpc/go.mod
```diff
@@ -1,15 +1,15 @@
...
-toolchain go1.23.4
+toolchain go1.24.0
 
 require (
 	go.opentelemetry.io/otel v1.31.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.23.1
 	go.opentelemetry.io/otel/sdk v1.31.0
 	go.opentelemetry.io/otel/trace v1.31.0
-	google.golang.org/grpc v1.69.4
+	google.golang.org/grpc v1.69.0-dev
```
dashpole, line 12: we probably want to avoid using development releases of grpc

example/trace/otlphttp/go.mod
braydonk: The collector [requires go1.23.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/go.mod#L11) which is why a lot of those got updated; I won't be able to put them back and remain on this OTel version

example/trace/otlpgrpc/go.mod
braydonk: Not intentional. I think this came as a result of the `update-otel` make target somehow.

braydonk: This is driving me up the wall. I tried to manually fix this and it made this happen:
```
braydonk@bk:~/Git/opentelemetry-operations-go/example/metric/collector$ go get google.golang.org/grpc@v1.70.0
go: upgraded go.opentelemetry.io/otel v1.31.0 => v1.32.0
go: upgraded go.opentelemetry.io/otel/metric v1.31.0 => v1.32.0
go: upgraded go.opentelemetry.io/otel/sdk v1.31.0 => v1.32.0
go: upgraded go.opentelemetry.io/otel/sdk/metric v1.31.0 => v1.32.0
go: upgraded go.opentelemetry.io/otel/trace v1.31.0 => v1.32.0
go: upgraded google.golang.org/grpc v1.69.0-dev => v1.70.0
```

dashpole:
maybe we should do the dependency update by itself first...  What do you need updated to fix this?

example/trace/otlphttp/go.mod
dashpole: If possible, we should drop support for go1.22 after the release happens

braydonk:
I don't remember what broke, but I was incapable of building before doing an OTel update, there was an error in an underlying package that blocked builds and I couldn't figure it out. So I thought I'd an OTel update would fix it, and then disaster struck

dashpole:
lets do https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/1019 first

braydonk:
I'm going to recreate this PR rebased on main now that #1019 is merged. The conflict resolution will be too much of a mess

PR #1022, opened 2025-02-21
collector: use buildinfo to make user agent

braydonk:
This PR adjusts the `NewGoogleCloud*Exporter` functions to take a
`component.BuildInfo`, and to format that build info into the UserAgent
instead of defaulting to `opentelemetry-collector-contrib`.


dashpole (approved):

exporter/collector/integrationtest/protos/fixtures.pb.go
```diff
@@ -14,7 +14,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.2
+// 	protoc-gen-go v1.36.5
```
dashpole, line 17: You might need to update the version of protoc-gen-go in the tools directory to match this.  Is there a reason you want this updated?

braydonk: `protoc-gen-go` is installed by the Makefile, but does not appear to be in the `tools/go.mod` file. Shall I add it there? And should I just choose this version?

braydonk: Oh actually I misunderstood, it's just a sub-package of the `protobuf` package. I didn't actually want to do this update, so I can't tell why the `check-clean-work-tree` target is failing.

braydonk: Updating protobuf did end up being the solution. I'm not sure why this was necessary, don't fully understand how all these bits fit together

braydonk:
Opened a draft PR with the changes I made and tested with `go replace` of this branch. When the new release is cut here, I will make this PR point to the release and open it for review.

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38116

PR #1031, opened 2025-04-08
exporter/collector: Add mapping for LogSeverity values set as SeverityText to correct SeverityNumber.

franciscovalentecastro:
Fix for https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1030.


exporter/collector/logs.go
```diff
@@ -93,6 +93,21 @@ var severityMapping = []logtypepb.LogSeverity{
 	logtypepb.LogSeverity_EMERGENCY, // 24 -> Emergency
 }
 
+// otelSeverityForGoogleCloudLoggingFormatSeverityText maps Google Cloud Logging LogSeverity string
+// values set in SeverityText field to SeverityNumbers to be properly exported as a LogEntry.
+// (https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity)
+var otelSeverityForGoogleCloudLoggingFormatSeverityText = map[string]plog.SeverityNumber{
```
ridwanmsharif, line 99: Why not just add these to the map below? `otelSeverityForText` seems to be serving exactly this purpose?

ridwanmsharif (commented):
Can we add some tests for this as well?

exporter/collector/logs.go
franciscovalentecastro: > Why not just add these to the map below?

The other map "otelSeverityForText" [^1] is used by first transforming all `SeverityText` string values to lowercase and then using the map. There are a two elements (`DEBUG`, `INFO`) in the map that would collision with the `LogSeverity` formatted fields (probably without an issue) by doing the `toLowerCase` transformation. 

Also, i wanted to make sure only the exact "uppercase" [^2] `LogSeverity` strings were mapped. For example, i don't to want  to map the string `notice` (lowercase) at all.

I think adding to "otelSeverityForText" would work, but wanted to make sure the above two things were true. What do you think ?

[^1]: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/logs.go#L475-L477

[^2]: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverit

jsuereth: What's the rationale for being case sensitive and `notice` not working?  Are we concerned about a real world conflict where this shouldn't map?

franciscovalentecastro: > What's the rationale for being case sensitive and notice not working?

I wanted to add the least amount of possible "severity string" cases so there were only two mutually exclusive maps : Opentelemetry style severity strings and Google Cloud Logging LogSeverity strings. I was trying to be the least intrusive in this addition.

> Are we concerned about a real world conflict where this shouldn't map?

No, i can't think of any real world conflict that may arise.

After thinking this through, i'll just add the new values (`debug`, `notice`, ... ) to the `otelSeverityForText` map as `lowercase`. It's not worth it to have it separate and it would provide more value in all the logs exported from the `googlecloudexporter`.

franciscovalentecastro: Done!

franciscovalentecastro:
@ridwanmsharif I added tests for the Google Cloud Logging `LogSeverity` string values and addressed the other comment.

ridwanmsharif (approved):

PR #1040, opened 2025-05-28
Allow overriding Project ID using env var

psx95:
Allows the users to override project ID by setting the `GOOGLE_CLOUD_PROJECT` environment variable.

#### Testing
 - Added unit tests.


extension/googleclientauthextension/factory.go
```diff
@@ -27,6 +28,10 @@ import (
 	"google.golang.org/grpc/credentials/oauth"
 )
 
+const (
+	envProjectID = "GOOGLE_CLOUD_PROJECT"
```
dashpole, line 32: Can you add a comment linking to documentation, or code that also uses this?  I.e. "Why did we choose this env var?"

dashpole (approved):

PR #1048, opened 2025-06-03
go.mod: use go 1.23

braydonk:
We don't want our packages to force a Go version bump upstream. This PR changes our modules to just use `go 1.23` with no patch version.

### DIFF
--- a/detectors/gcp/go.mod
+++ b/detectors/gcp/go.mod
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp
 
-go 1.23.8
+go 1.23
 
 require (
 	cloud.google.com/go/compute/metadata v0.7.0
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/tools
 
-go 1.23.8
+go 1.23
 
 require (
 	github.com/client9/misspell v0.3.4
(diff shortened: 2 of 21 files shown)


braydonk:
Upstream issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40436

psx95 (commented):
Shouldn't the [CI](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/d0b1fbf6ade5554c75256090b7d701028d62d679/.github/workflows/ci.yml#L9) be updated as well to match this ?

Or are we still ok to run tests explicitly against 1.23.8 ?

dashpole:
We should run tests explicitly against a patch version to satisfy the go vuln checker

detectors/gcp/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp
 
-go 1.23.8
+go 1.23.0
+
+toolchain go1.24.2
```
psx95, line 5: Would this mean that at least go1.24.2 is required to build and compile this project? I'm ok with this, but is this intentional ?

dashpole:
@braydonk can you fix the lint so we can get this merged?  This is impacting other projects, so we should get a patch release out with this soon

braydonk:
I've solved the lint problem. I'm not sure how to handle that final e2e test failure.

tools/go.mod
```diff
@@ -1,36 +1,38 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/tools
 
-go 1.23.8
+go 1.23.0
+
+
```
dashpole, line 5:
```suggestion
go 1.23.0
```

dashpole (approved):

PR #1088, opened 2025-12-20
fix: Support OTLP event_name field in logs exporter

### LINKED ISSUE #1078: Collector logs exporter drops the OTLP `event_name`
`LogEntry.event_name` was added a few months ago https://github.com/open-telemetry/opentelemetry-proto/pull/600 and we don't do anything with it today. Before it was added to the proto, event name was just a convention to set `event.name` attribute so it was being captured.

putcho01:
## Summary
fix: #1078

## Problem

The `event_name` field was added to the OTLP LogRecord proto in [open-telemetry/opentelemetry-proto#600](https://github.com/open-telemetry/opentelemetry-proto/pull/600), but the logs exporter was not handling this field. Previously, event names were set as `event.name` attributes and were captured in labels. With the new dedicated proto field, this information was being dropped.

## Solution

Map the OTLP `event_name` field to the Cloud Logging LogEntry label `"event.name"`. This approach:
- Maintains backward compatibility with the previous convention of using `event.name` as an attribute
- Works for all log payload types (text, JSON, etc.)
- Allows easy querying via `labels."event.name"` in Cloud Logging


dashpole:
/gcbrun

@DylanRussell how did you handle Event name in the python SDK log exporter?  We should probably try to be consistent here.

dashpole (approved):
I checked, and this is consistent with they python implementation

PR #1103, opened 2026-02-12
Improve unit tests for gcp auth extension

dashpole:
This mocks the metadata server to return auth tokens, and verifies that they are properly attached to http requests.


dashpole:
I wanted to write a better test for https://github.com/grafana/alloy/pull/5526, and figured putting it in our repo as well would be good.

PR #1106, opened 2026-02-23
Add OTLP integration tests

dashpole:
This tests the same OTLP payloads against the `telemetry.googleapis.com` OTLP endpoint that we use to test the googlecloud and googlemanagedprometheus exporters.


dashpole:
Moving to draft until the test can pass

PR #1112, opened 2026-03-09
Add support for cloud run worker pools

dashpole:
https://docs.cloud.google.com/run/docs/container-contract#worker-pools-env-vars

* CLOUD_RUN_WORKER_POOL -> faas.name
* CLOUD_RUN_REVISION -> faas.version
* instance id from metadata server -> faas.instance
* region from metadata server -> cloud.region

This is similar to the standard cloud run service, except CLOUD_RUN_WORKER_POOL instead of K_SERVICE and CLOUD_RUN_REVISION instead of K_REVISION.

### Verification
Deployed a test workload to Cloud Run worker pools and verified the environment variables:
* `CLOUD_RUN_WORKER_POOL` is set to the worker pool name.
* `CLOUD_RUN_REVISION` is set to the revision name.
* `CLOUD_RUN_WORKER_POOL_REVISION` is NOT set.

### DIFF
--- a/detectors/gcp/faas.go
+++ b/detectors/gcp/faas.go
@@ -28,14 +28,19 @@ const (
 	//
 	// Cloud Run jobs env vars:
 	// https://cloud.google.com/run/docs/container-contract#jobs-env-vars
-	cloudFunctionsTargetEnv  = "FUNCTION_TARGET"
-	cloudRunConfigurationEnv = "K_CONFIGURATION"
-	cloudRunJobsEnv          = "CLOUD_RUN_JOB"
-	faasServiceEnv           = "K_SERVICE"
-	faasRevisionEnv          = "K_REVISION"
-	cloudRunJobExecutionEnv  = "CLOUD_RUN_EXECUTION"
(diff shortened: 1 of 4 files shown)


detectors/gcp/faas.go
```diff
@@ -69,6 +82,9 @@ func (d *Detector) FaaSVersion() (string, error) {
 	if version, found := d.os.LookupEnv(faasRevisionEnv); found {
 		return version, nil
 	}
+	if version, found := d.os.LookupEnv(cloudRunWorkerPoolRevision); found {
+		return version, nil
+	}
```
avilevy18:
```suggestion
	if version, found := d.os.LookupEnv(cloudRunWorkerPoolRevision); found {
		return version, nil
	}
	if version, found := d.os.LookupEnv(faasRevisionEnv); found {
		return version, nil
	}
```
Nit suggestion: if we switch the order then even if both are set once Cloud Run fixes it, we will prefer the worker pool.

avilevy18 (approved):
LGTM

PR #1141, opened 2026-04-14
exporter/trace: sanitize invalid UTF-8 in span fields

### LINKED ISSUE #901: Trace Exporter sanitizes invalid UTF-8
Forked from https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35723

> Our collector exporter uses the SDK exporter under the hood, but that shouldn't actually matter. We probably want to make the trace client remove invalid UTF-8 similar to what we do for metrics: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/16bba4f4e879814de7d3354ef83bcfd597e44b15/exporter/metric/metric.go#L583. It probably needs to be done somewhere around here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/16bba4f4e879814de7d3354ef83bcfd597e44b15/exporter/trace/trace_proto.go#L168

alliasgher:
## Summary

Fixes #901.

The metric exporter already sanitizes invalid UTF-8 by calling `strings.ToValidUTF8` on attribute values (see `metric.go:sanitizeUTF8`). The trace exporter had no equivalent, causing Cloud Trace to reject spans whose attribute values, span names, event names, or status descriptions contained invalid byte sequences.

Add a `sanitizeUTF8` helper to `trace_proto.go` (matching the one in `metric.go`) and apply it to:
- span `DisplayName`
- status `Description`
- string `AttributeValue`s
- event annotation `Description`s

## Testing

Added `TestSanitizeUTF8` regression test that feeds a span with `\xff\xfe` bytes in all four fields and asserts the exported proto contains no invalid bytes. All existing tests continue to pass.


dashpole:
Hey, thanks for the contribution!  I would strongly recommend migrating to the OTLP exporter: https://docs.cloud.google.com/stackdriver/docs/instrumentation/migrate-to-otlp-endpoints#telemetry_add_dependencies-go

I can still take a look at this, but we are being extremely careful about making changes because we want to keep this exporter's behavior similar to the new OTLP endpoint.

Did you individually confirm that the trace API rejects invalid UTF-8 for each of fields that you are sanitizing?

alliasgher:
Understood — thanks for the heads up on the OTLP migration. Happy to have this closed if it doesn't fit the maintenance posture.

dashpole (approved):

Issue #1062, opened 2025-07-30
[exporter/collector/logs] Exporter doesn't accept `httpRequest.status` integer value.


franciscovalentecastro:
Currently the `exporter/collector` doesn't export correctly Opentelemetry Logs with an `httpRequest.status` field that has been explicitly set to `integer` type (`Int()` in OTTL) . The whole `httpRequest` map is dropped since the tags `json:"status,string"` can only expect a string, throwing an error if the JSON field shows up like `{ "httpRequest" : { "status": 200 } } ` :
```
Error unmarshaling integer JSON: json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int32
```

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/48e79e1a69952f45aad4afe134bfeafb570e4db1/exporter/collector/logs.go#L574-L593

This can be easily reproduced : https://go.dev/play/p/sSuGklt_TMV . Notice there is no simple "update tag" fix (remove `,string`) due to how the json Unmarshall works.
 
Related issue : b/435212354

Edit : The following fields can also present the same issue : 
- `httpRequest.responseSize`
- `httpRequest.requestSize`
-  `httpRequest.cacheFillBytes`

quentinmit:
This looks like an oversight in the exporter. Request, response, and cache fill size *should* be strings (because they are int64s and JSON cannot represent int64) but status should be an integer. See: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest

Ideally we'd support both integer and string for all the fields though, for maximum backwards and forwards compatibility.

PR #1142, opened 2026-04-14
exporter/collector/logs: accept string-typed line in sourceLocation

### LINKED ISSUE #1119: [exporter/collector/logs] The exporter rejects it when `sourceLocation.line` comes in as a string
when `sourceLocation` attribute contains an string typed line number (in a given log line):

```{"logging.googleapis.com/severity": "WARNING", "logging.googleapis.com/labels": {"label1":"value1", "label2":"value2"}, "logging.googleapis.com/operation": {"id": "id", "producer": "producer", "first": true, "last": true}, "logging.googleapis.com/sourceLocation": {"file": "file", "line": "1", "function": "function"}, "logging.googleapis.com/trace":"projects/my-project/traces/deadcafedeadc0defacefeedb0bacafe", "logging.googleapis.com/spanId":"facefeedb0bacafe", "normal_field": "value"}```

google cloud exporter rejects it with the error: `could not process attribute gcp.source_location: json: cannot unmarshal string into Go struct field LogEntrySourceLocation.line of type int64`

Here is where it fails in google cloud exporter:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fbd3fd5fc47478a6ab4586e1d3a319ce944ceeeb/exporter/collector/logs.go#L451 -> https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fbd3fd5fc47478a6ab4586e1d3a319ce944ceeeb/exporter/collector/logs.go#L740

We can fix this the same way as in issue #1062 by changing the type of `sourceLocation.line` to `Int64OrString`.

alliasgher:
Fixes #1119. Adds a `sourceLocationLog` intermediate struct (using `Int64OrString` for the `line` field, same pattern as #1062) to accept both numeric and quoted-string line numbers. Updated the test that previously expected an error to now expect success.


dashpole:
Looks like you mistakenly included the trace commit in this pr

dashpole:
Similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/1141#issuecomment-4244708940, we are encouraging users to migrate to the new GCP OTLP endpoint.

See https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/v1.logs (better docs to come in the near future, but you can follow the trace docs).

I'll still take a look, but we may choose not to merge this to avoid making changes to the exporter.

alliasgher:
Fixed — cleaned up the branch to only contain the sourceLocation commit.

alliasgher:
Understood re: OTLP migration. Happy to close if this doesn't fit.

dashpole (approved):

PR #1143, opened 2026-04-14
exporter/collector/logs: use parsed httpRequest.protocol instead of hardcoded HTTP/1.1

### LINKED ISSUE #1120: [exporter/collector/logs] the Exporter hardcodes `httpRequest.protocol` to `http/1.1` and should not
See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fbd3fd5fc47478a6ab4586e1d3a319ce944ceeeb/exporter/collector/logs.go#L671

The exporter should use whatever protocol value parsed from the input.

alliasgher:
Fixes #1120. The httpRequestLog struct already captured the protocol field from the input, but parseHTTPRequest always set the Cloud Logging proto field to the literal string "HTTP/1.1". Wire parsedHTTPRequest.Protocol through and update four test cases that were sending HTTP/2 but expecting HTTP/1.1 back.


dashpole:
Looks like you mistakenly included the other commits here.  Same comment about recommending migrating to OTLP

alliasgher:
Fixed — cleaned up the branch to only contain the httpRequest.protocol commit.

alliasgher:
Understood re: OTLP migration. Happy to close if this doesn't fit.

XuechunHou:
I think this PR, along with #1142, should still be reviewed and merged. The Google Cloud Exporter is still depended on and required by collectors like the Ops Agent. This is especially important during the interim while the migration to the UTR endpoint is in progress, and for universes where the UTR endpoint is not yet available.

dashpole (approved):

PR #1154, opened 2026-05-05
Auto comment /gcbrun for dependency updates

dashpole:
Lets see if this helps make dependency updates easier


aabmass:
Were you able to test this? I think the user who leaves the comment has to have admin access on the repo. In JS we have the dpe bot: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/882#issuecomment-4331480073

dashpole:
Do you think I should try setting up the DPE bot?  How hard was that?

aabmass:
It was easy https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/.github/trusted-contribution.yml, but IIRC you said it wasn't working?

If the Github default token comment works I'm fine to shift to that.

dashpole:
Looks like that functionality is deprecated: https://github.com/googleapis/repo-automation-bots/tree/main/packages/trusted-contribution#%EF%B8%8F-deprecated--trusted-contribution

aabmass:
Ya, happy to migrate to whatever you find that works. I also haven't kept up on Cloud Build features generally so there might be a more elegant solution

dashpole:
Note that pull_request_target only respects the configuration from `main`.  I'm not sure if github actions counts as a contributor for cloud build purposes. If its ok, i'll just merge it and find out

dashpole:
I switched to github.actor.  Both that and the previous field were secure (can't be spoofed), but actor means dependabot/renovate actually need to have done the push for it to comment (someone can't push to their branch to trigger it).

aabmass (approved):

PR #1169, opened 2026-05-26
Migrate impersonate to use cloud.google.com/go/auth

### LINKED ISSUE #1168: Migrate impersonate to use cloud.google.com/go/auth
The [impersonate logic](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4884c69c180d1426accaca97134107980be70eae/exporter/collector/config.go#L31) is still utilizing the old auth flow (which does not support universe domains).

This should be migrated to using https://pkg.go.dev/cloud.google.com/go/auth package, which does correctly exchange and construct the endpoint matching the [informed universe domain](https://github.com/googleapis/google-cloud-go/blob/main/auth/credentials/impersonate/impersonate.go#L255)

dashpole:
## Summary
Migrates the impersonation logic in `exporter/collector` to use `cloud.google.com/go/auth/credentials/impersonate`. This supports universe domains correctly.
Fixes #1168

## Testing
Manual verification was performed as follows:
1. Created a service account with restricted permissions (e.g., `Monitoring Metric Writer`).
2. Granted `Service Account Token Creator` role to the calling user on that service account.
3. Created a test that calls `generateClientOptions` with impersonation config.
4. Verified that a call made with the resulting client failed with `PermissionDenied` for a read operation (e.g., `ListMetricDescriptors`), confirming the restricted identity was used.
Unit tests and integration tests in `exporter/collector` were also run and passed.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -28,7 +28,7 @@ import (
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/otel/metric"
 	"golang.org/x/oauth2/google"
-	"google.golang.org/api/impersonate"
+	"cloud.google.com/go/auth/credentials/impersonate"
 	"google.golang.org/api/option"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


psx95 (approved):
Vulncheck was failing, one dependency requires updating - seems unrelated to the changes in this PR.

PR #1177, opened 2026-07-06
Add GKEHostType

### LINKED ISSUE #1176: Add GKE host type detection to the GCP detector
While working on GKE resource detection, I found that the GCP detector can identify that it is running on GKE and can read cluster metadata such as the cluster name and location, but it does not expose the machine type of the underlying GCE instance. That leaves downstream integrations without a straightforward way to populate `host.type` for workloads running on GKE, even though the information exists in Google Cloud.

I would like to propose adding `GKEHostType()` to the detector. The implementation would call the Compute Engine `instances.get` API for the current instance, read the `machineType` field, and normalize it to the short machine type name. For example, if the API returns a full URI such as: `https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-a/machineTypes/e2-standard-4`

iblancasa:
Fixes #1176


iblancasa:
Should the dependencies be bumped?
```

Vulnerability #1: GO-2026-5158
    Opentelemetry-go's baggage parsing no longer caps raw header length in
    go.opentelemetry.io/otel
  More info: https://pkg.go.dev/vuln/GO-2026-5158
  Module: go.opentelemetry.io/otel
    Found in: go.opentelemetry.io/otel@v1.43.0
    Fixed in: go.opentelemetry.io/otel@v1.44.0
    Example traces found:
Error:       #1: client/client.go:88:23: client.main calls baggage.New
Error:       #2: server/server.go:89:27: server.main calls http.ListenAndServe, which eventually calls baggage.Parse
Error:       #3: server/server.go:89:27: server.main calls http.ListenAndServe, which eventually calls propagation.Baggage.Extract
```

dashpole:
You can ignore that. It is non-blocking

dashpole (approved):

PR #1178, opened 2026-07-10
Prepare release v1.34.0/v0.58.0

braydonk:


braydonk:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/actions/runs/29099940687/job/86385926800?pr=1178

I don't think there's anything we can do about this? There's no patch for go 1.26 for this. It's a matter of being able to pull in a new enough Go version in our CI, which we won't be able to do.

dashpole:
govulncheck is not a required presubmit

dashpole (approved):

PR #1181, opened 2026-07-16
Adds support for resolving OpenTelemetry authentication extensions

Angelawork:
Adds support for resolving OpenTelemetry authentication extensions in 
`opentelemetry-operations-go/exporter/collector`.

This allows `googlecloudexporter` (Metrics, Logs, and Traces) to authenticate 
against Google Cloud APIs using extensions registered in `host.GetExtensions()` 
(such as custom OAuth extensions) instead of 
relying solely on Application Default Credentials.

Key changes:
1. `Config`: Added `authenticator` string config option (`Authenticator string mapstructure:"authenticator"`).
2. Added `getAuthenticatorClientOptions` to inspect host extensions dynamically in priority order:
   - `clientOptionProvider` (`ClientOptions() []option.ClientOption`)
   - `tokenSourceProvider` (`TokenSource() oauth2.TokenSource`)
   - `oauth2.TokenSource` (`Token() (*oauth2.Token, error)`)
   - `perRPCCredentialsProvider` (`PerRPCCredentials() (credentials.PerRPCCredentials, error)`)
3. Updated `MetricsExporter.Start`, `LogsExporter.Start`, and `TraceExporter.Start` to invoke `getAuthenticatorClientOptions` before building GCP client options.

### DIFF
--- a/exporter/collector/config.go
+++ b/exporter/collector/config.go
@@ -15,28 +15,28 @@
 package collector
 
 import (
-	"cloud.google.com/go/auth/credentials/impersonate"
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
+	"runtime"
+	"strings"
+	"time"
+
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/extension/extensionauth"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/otel/metric"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"cloud.google.com/go/auth/credentials/impersonate"
 	"google.golang.org/api/option"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
 	otelgrpc "google.golang.org/grpc/stats/opentelemetry"
-	"regexp"
-	"runtime"
-	"strings"
-	"time"
-
-	"google3/third_party/golang/grpc/credentials/credentials"
-	"google3/third_party/golang/oauth2/oauth2"
 )
 
 const (
@@ -60,16 +60,20 @@ type Config struct {
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
 
-type tokenSourceProvider interface {
-	TokenSource() oauth2.TokenSource
-}
-
 type clientOptionProvider interface {
 	ClientOptions() []option.ClientOption
 }
 
-type perRPCCredentialsProvider interface {
-	PerRPCCredentials() (credentials.PerRPCCredentials, error)
+type contextTokenSourceProvider interface {
+	Token(context.Context) (*oauth2.Token, error)
+}
+
+type googleApiContextSourceWrapper struct {
+	ts contextTokenSourceProvider
+}
+
+func (w *googleApiContextSourceWrapper) Token() (*oauth2.Token, error) {
+	return w.ts.Token(context.Background())
 }
 
 // getAuthenticatorClientOptions resolves an authentication extension in host.GetExtensions() by name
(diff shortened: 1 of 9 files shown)


exporter/collector/config.go
```diff
@@ -15,26 +15,28 @@
...
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
 	otelgrpc "google.golang.org/grpc/stats/opentelemetry"
+	"regexp"
+	"runtime"
+	"strings"
+	"time"
+
+	"google3/third_party/golang/grpc/credentials/credentials"
```
braydonk: This doesn't work outside of Google. You'll have to use the public modules for this, i.e. `golang.org/x/oauth2` and `google.golang.org/grpc/credentials`.

exporter/collector/config.go
```diff
@@ -55,6 +60,56 @@ type Config struct {
...
 
+type tokenSourceProvider interface {
+	TokenSource() oauth2.TokenSource
+}
+
+type clientOptionProvider interface {
+	ClientOptions() []option.ClientOption
+}
+
+type perRPCCredentialsProvider interface {
```
braydonk: This should use the interface from the Collector's `extensionauth` package:
https://github.com/open-telemetry/opentelemetry-collector/blob/47267b188b260e1c4123de9d89bdc886f05d4d3e/extension/extensionauth/client.go#L25

exporter/collector/config.go
```diff
@@ -55,6 +60,56 @@ type Config struct {
 	DestinationProjectQuota bool              `mapstructure:"destination_project_quota"`
 }
 
+type tokenSourceProvider interface {
+	TokenSource() oauth2.TokenSource
```
braydonk: The interface we should look for is the one `oauth2clientauthextension` has, which would be these: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/90417677da7a28e0fe3910b592e1e8e1fd624288/extension/oauth2clientauthextension/extension.go#L31-L40

We can't really use these interfaces directly since they come from a component, so for now we can stick to defining private interfaces but let's make them match this extension so that we could recognize that extension if it's used.

exporter/collector/logs.go
```diff
@@ -214,7 +214,16 @@ func (l *LogsExporter) ConfigureExporter(config *logsutil.ExporterConfig) {
...
 }
 
-func (l *LogsExporter) Start(ctx context.Context, _ component.Host) error {
+func (l *LogsExporter) Start(ctx context.Context, host component.Host) error {
+	if l.cfg.Authenticator != "" {
+		authOpts, err := getAuthenticatorClientOptions(host, l.cfg.Authenticator)
+		if err != nil {
+			return err
+		}
+		l.cfg.LogConfig.ClientConfig.GetClientOptions = func() []option.ClientOption {
```
braydonk, line 224: This will likely break unit tests because they set `GetClientOptions` for mocking scenarios. In all 3 exporters you should check if `GetClientOptions` is set first and only set this if it's not.

exporter/collector/config.go
```diff
@@ -15,26 +15,28 @@
 package collector
 
 import (
+	"cloud.google.com/go/auth/credentials/impersonate"
```
braydonk: This is incorrect formatting. I think you have some other formatter here that is doing something strange. Revert the ordering back the way it was, you can use [GCI](https://github.com/daixiang0/gci) for this.

exporter/collector/logs.go
braydonk: Also I think this is missing the import for `option`.

braydonk (commented):
Please add some tests for the new behaviour.

exporter/collector/config.go
```diff
@@ -55,6 +60,60 @@ type Config struct {
...
 
+type clientOptionProvider interface {
+	ClientOptions() []option.ClientOption
+}
+
+type contextTokenSourceProvider interface {
+	Token(context.Context) (*oauth2.Token, error)
+}
+
+type oauth2ContextTokenSourceWrapper struct {
```
braydonk: Let's rename this wrapper for it's purpose, since it confused me at first glance. We should rename it to `googleApiContextSourceWrapper`

exporter/collector/logs.go
braydonk: The second comment on this line was addressed but it doesn't look like the first one was.

Angelawork: I added the check `&& l.cfg.LogConfig.ClientConfig.GetClientOptions == nil` to `if l.cfg.Authenticator != "" ` so it'll skip the authenticator block if GetClientOptions is set

braydonk (approved):

PR #1186, opened 2026-08-11
exporter/trace: update span attribute and event limits to match backend

dashpole:
Update span limits in the Cloud Trace exporter to align with the Cloud Trace backend and OTLP endpoint:
- Max display name length: 1024 bytes (was 128)
- Max attribute key length: 512 bytes (was 128)
- Max attribute string/float value length: 65532 bytes (was 256)
- Max annotation description length: 1024 bytes (was 256)
- Max annotation events per span: 256 (was 32)


exporter/trace/trace_proto.go
```diff
@@ -39,12 +39,15 @@ import (
...
 	// TODO(ymotongpoo): uncomment this after gRPC trace get supported.
-	// maxMessageEventsPerSpan    = 128.
-	maxAttributeStringValue = 256
-	maxNumLinks             = 128
-	agentLabel              = "g.co/agent"
+	// maxMessageEventsPerSpan      = 128.
+	maxDisplayNameLength           = 1024
+	maxAnnotationDescriptionLength = 1024
+	maxAttributeKeyLength          = 512
+	maxAttributeStringValue        = 65532
```
psx95: nit: This is actually 64 KB, in python this is represented as 64 * 1024 = 65536. 

Not sure if 4 bytes make a difference or if it was deliberate here.

exporter/trace/trace_proto.go
```diff
@@ -39,12 +39,15 @@ import (
...
-	maxAnnotationEventsPerSpan = 32
+	maxAnnotationEventsPerSpan     = 256
 	// TODO(ymotongpoo): uncomment this after gRPC trace get supported.
-	// maxMessageEventsPerSpan    = 128.
-	maxAttributeStringValue = 256
-	maxNumLinks             = 128
-	agentLabel              = "g.co/agent"
+	// maxMessageEventsPerSpan      = 128.
+	maxDisplayNameLength           = 1024
+	maxAnnotationDescriptionLength = 1024
```
psx95, line 46: QQ - What does `maxAnnotationDescriptionLength` map to in https://docs.cloud.google.com/trace/docs/quotas#telemetry-api-limits ?

exporter/trace/trace_proto.go
dashpole: addressed in 55d7eb5

exporter/trace/trace_proto.go
dashpole: It maps to "Maximum size of an Event object's name" (1024 bytes) in the [Telemetry API limits](https://docs.cloud.google.com/trace/docs/quotas#telemetry-api-limits) table. In OpenTelemetry, span events have a Name field, which gets converted to Span_TimeEvent_Annotation.Description in the Cloud Trace proto representation.

psx95 (approved):

PR #1190, opened 2026-08-21
Drop support for go 1.25

dashpole:
This should unblock a bunch of dependency updates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
   compatibility-test:
     strategy:
       matrix:
-        go-version: ["1.25.8", "1.26.0", "1.27.0"]
+        go-version: ["1.26.0", "1.27.0"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         arch: ["386", amd64]
         exclude:
--- a/cloudbuild-e2e-cloud-functions-gen2.yaml
+++ b/cloudbuild-e2e-cloud-functions-gen2.yaml
@@ -14,7 +14,7 @@
 
 steps:
   # Vendor dependencies, and zip the code.
-  - name: golang:1.25.0
+  - name: golang:1.26.0
     id: zip-code
     entrypoint: /bin/bash
     args:
--- a/cloudbuild-e2e-gae-standard.yaml
+++ b/cloudbuild-e2e-gae-standard.yaml
@@ -14,7 +14,7 @@
 
 steps:
   # Vendor dependencies, and zip the code.
-  - name: golang:1.25.0
+  - name: golang:1.26.0
     id: zip-code
     entrypoint: /bin/bash
     args:
@@ -33,7 +33,7 @@ steps:
     env: ["PROJECT_ID=$PROJECT_ID", "TEST_RUN_ID=$BUILD_ID"]
     args:
       - gae-standard
-      - --runtime=go125
+      - --runtime=go126
       - --appsource=/workspace/e2e-test-server/appsource.zip
       - --skip-cleanup
 logsBucket: gs://opentelemetry-ops-e2e-cloud-build-logs
--- a/exporter/collector/go.mod
+++ b/exporter/collector/go.mod
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector
 
-go 1.25.0
+go 1.26.0
 
 toolchain go1.26.3
 
--- a/Makefile
+++ b/Makefile
@@ -158,9 +158,9 @@ lint: $(GOLANGCI_LINT) $(MISSPELL)
 	done
 	$(MISSPELL) -w $(ALL_DOCS)
 	set -e; for dir in $(ALL_GO_MOD_DIRS) $(TOOLS_MOD_DIR); do \
-	  echo "go mod tidy -compat=1.24 in $${dir}"; \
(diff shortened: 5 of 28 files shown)


aabmass (approved):

cloudbuild-e2e-cloud-functions-gen2.yaml
```diff
@@ -14,7 +14,7 @@
 
 steps:
   # Vendor dependencies, and zip the code.
-  - name: golang:1.25.0
+  - name: golang:1.26.0
     id: zip-code
     entrypoint: /bin/bash
     args:
```
dashpole: `--runtime=go123` left stale in a file this PR edits — Medium — CONFIRMED (drift) / PLAUSIBLE (breakage)

The PR bumps the builder image in this file and the deploy runtime in both GAE files (`--runtime=go125` → `go126` at `cloudbuild-e2e-gae-standard.yaml:36` and `cloudbuild-e2e-gae.yaml:33`), but leaves the Cloud Functions deploy runtime at `go123`. `git log -G'runtime=go1'` shows it was last touched in #1019; #1069 and #1114 bumped the GAE files and skipped it, so it has been drifting for three release cycles.

Suggested fix: `--runtime=go126` here too, matching the two GAE files. If GCF gen2 does not yet offer `go126`, that is the actual signal — pin the runtime to the newest available and say so in a comment, rather than leaving a number nobody re-reads.

exporter/collector/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector
```
dashpole: Library modules' Go floor moves ahead of both upstream and downstream; no note anywhere

These are consumed by other projects. In the local contrib checkout, 18 `go.mod` files require this repo, spanning seven distinct modules of it:

Suggested fix: No code change needed. Put one line in the PR body and the release notes: "raises the minimum Go version for all published modules to 1.26; consumers on Go 1.25 must stay on v1.36.0/v0.60.0." Optionally check whether core collector plans to move before cutting the next release, since contrib will inherit this.

.github/workflows/ci.yml
```diff
@@ -148,7 +148,7 @@ jobs:
   compatibility-test:
```
dashpole: `DEFAULT_GO_VERSION` is labelled "minimum" but holds the maximum, and now disagrees with the new floor

`lint`, `govulncheck`, `test-race` and `test-coverage` all run at `~1.27.0`. After this PR the declared minimum is `1.26.0`. The comment and the value have contradicted each other since #1188 bumped this from `~1.26.0` to `~1.27.0`; this PR is where the two numbers are supposed to be reconciled and it leaves them apart.

Suggested fix: Either fix the comment to "Go version used for lint/test/coverage jobs", or set `DEFAULT_GO_VERSION: "~1.26.0"` to match the floor and let the matrix cover 1.27. The first is less disruptive.

dashpole: The compatibility matrix does not pin the toolchain, so the `1.25.8` leg being removed was never running Go 1.25.8

The job sets `go-version: ${{ matrix.go-version }}` and never sets `GOTOOLCHAIN`, so it runs with the default `auto`. A `toolchain go1.26.3` line newer than the installed toolchain makes the go command download and switch before running anything. I hit this directly in this container: the base install is `go1.24.7`, yet the very first `go version` executed inside the repo printed `go: downloading go1.26.3` / `go version go1.26.3 linux/amd64`.

Suggested fix: Add `GOTOOLCHAIN=local` to the `compatibility-test` job's `Setup Environment` step (`echo "GOTOOLCHAIN=local" >> $GITHUB_ENV`), which is what `Makefile:180`'s `for-all-mod` already does locally. Alternatively drop the `toolchain` lines from the module `go.mod`s and keep only the `go` directive.

dashpole: `GOOTOOLCHAIN` typo makes two jobs' toolchain pinning a no-op

The `lint` and `govulncheck` jobs intend to pin the toolchain and instead export an unread variable, so both run with `GOTOOLCHAIN=auto`.

Suggested fix: `GOTOOLCHAIN=local`.

cloudbuild-e2e-cloud-functions-gen2.yaml
```diff
@@ -14,7 +14,7 @@
 
 steps:
   # Vendor dependencies, and zip the code.
-  - name: golang:1.25.0
+  - name: golang:1.26.0
```
dashpole: Builder-image pins disagree on patch level with the `toolchain` directive

Under `GOTOOLCHAIN=auto` the `golang:1.26.0` steps will download go1.26.3 on every Cloud Build run before doing any work.

Suggested fix: Use `golang:1.26.3` in the three cloudbuild files to match the Dockerfile and the `toolchain` lines.

cloudbuild-e2e-gae-standard.yaml
```diff
@@ -33,7 +33,7 @@ steps:
     env: ["PROJECT_ID=$PROJECT_ID", "TEST_RUN_ID=$BUILD_ID"]
     args:
       - gae-standard
-      - --runtime=go125
+      - --runtime=go126
```
dashpole: `--runtime=go126` assumes App Engine already offers that runtime

GAE standard runtimes lag upstream Go releases. If `go126` is not yet GA, both e2e jobs fail at deploy.

PR #1201, opened 2026-09-04
exporter/collector/logs: Add support for the `gcp.operation` attribute

quentinmit:
This mirrors the [support from the Telemetry
API](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/otlp-log-record-to-log-entry#operation) and is needed to unblock migration off Fluent Bit.

### DIFF
--- a/exporter/collector/logs.go
+++ b/exporter/collector/logs.go
@@ -677,15 +677,8 @@ func (f *BoolOrString) UnmarshalJSON(data []byte) error {
 
 	var str string
 	if err := json.Unmarshal(data, &str); err == nil {
-		switch str {
-		case "true":
-			*f = true
-			return nil
-		case "false":
-			*f = false
-			return nil
-		}
-		return fmt.Errorf("failed to convert string to boolean: %w", err)
+		*f = str == "true"
+		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


franciscovalentecastro (approved):
LGTM!

exporter/collector/logs.go
```diff
@@ -653,6 +663,34 @@ func (f *Int64OrString) UnmarshalJSON(data []byte) error {
...
+	if err := json.Unmarshal(data, &str); err == nil {
+		switch str {
+		case "true":
+			*f = true
+			return nil
+		case "false":
+			*f = false
+			return nil
+		}
+		return fmt.Errorf("failed to convert string to boolean: %w", err)
```
dashpole: Inside this block `err` is nil from `json.Unmarshal`, so `%w` formats as `%!w(<nil>)`. Consider using `strconv.ParseBool(strings.TrimSpace(str))` to parse and capture the actual error, or formatting `str` without `%w`.

dashpole (approved):

exporter/collector/logs.go
quentinmit: Actually, I went and found the code in telemetry.googleapis.com that parses this field, and it actually treats `"true"` as true and any other string as false. So I've made that change here, which obsoletes this error return.

